PR #482, opened 2014-11-12
use more appropriate functions from Test::More in tests

reneeb:
When a test is failing, you get better debugging output when the "correct" Test::More function is used.


krimdomu:
@ferki 

currently i'm going through the open pull requests to see if we can include them for the 1.0.0 release.
Can we merge this one?

ferki:
@krimdomu: yeah, let's get this going again, so I'll squash and merge this changeset with perltidy fixes and will push my adiitional test modifications later from my branch

ferki:
@reneeb: apologies for the extreme delay, it is merged now. I squashed together your commits, applied perltidy, and also applied your changes to some of the tests that has been added meanwhile to ensure consistency.

Big thanks for this pull request! :+1: 

@krimdomu: I tested the changes in a separate branch manually and also with jenkins - everything passed, and all the changes are looking good, but since we changed lots of the tests hemselves, it might worth to be a bit more suspicious about them if something starts to fail :)

PR #502, opened 2014-12-03
Rex::FS::File accepts filenames now

reneeb:
This would help to unify code. There is no need to create filehandle objects via Rex::Interface::File and pass the filehandle...


krimdomu:
Hi reneeb,

sorry again for the delay in here. Thank you for this patch, i'm going to merge it now.

PR #506, opened 2014-12-10
Adds LinuxMint

JJ:
Adds LinuxMint, which is basically a copy of Ubuntu. Previously Mint yielded "Not supported OS" or suchlike. Thanks to @toblerone554 for hint.


ferki:
This issue looks like abandoned, but we decided to try to add Linux Mint support by recognizing it as an Ubuntu clone (which might be even easier to implement than writing hardware gathering and service/package management abstractions).

ferki:
@JJ: this has been solved via merging #745. Thanks for your help around realizing this feature!

PR #525, opened 2015-01-07
unit-tests for Rex::Commands::DB

vkroll:
* only run if Test::mysqld is installed
* only works with mysql (better then nothing)
* initialize and cleanup code for DB tests
* added some basic test for the "old" behaviour

### DIFF
--- a/t/db.t
+++ b/t/db.t
@@ -1,0 +1,102 @@
+
+use strict;
+use warnings;
+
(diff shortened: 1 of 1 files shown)


krimdomu:
Hi vkroll,

sorry for the delay. Thank you much for this patch. I'm going to merge it now.

PR #533, opened 2015-01-25
Add PkgConf command to configure packages

abeverley:
This patch adds commands ```get_pkgconf``` and ```set_pkgconf``` to configure packages. As it stands, this patch only provides support for Debian (using debconf), but it is designed to be extensible, in the same way that the Pkg commands are.

Let me know if you think this should be approached in a different way, and I'll refactor accordingly.

### DIFF
--- a/lib/Rex/PkgConf/Debian.pm
+++ b/lib/Rex/PkgConf/Debian.pm
@@ -1,0 +1,105 @@
+#
+# (c) Jan Gehring <jan.gehring@gmail.com>
+#
+# vim: set ts=2 sw=2 tw=0:
+# vim: set expandtab:
+
+package Rex::PkgConf::Debian;
+
+use strict;
+use warnings;
+
+# VERSION
+
+use Rex::Helper::Run;
+
+use Rex::PkgConf::Base;
+use base qw(Rex::PkgConf::Base);
+
+sub new {
+  my $that  = shift;
+  my $proto = ref($that) || $that;
+  my $self  = $proto->SUPER::new(@_);
+
+  bless( $self, $proto );
+
+  return $self;
+}
+
+sub get_options {
+  my ( $self, $pkg, $option ) = @_;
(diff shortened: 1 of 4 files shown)


krimdomu:
Hi Andy,

thanks much for this patch. I'm going to merge it into master branch.

abeverley:
Thanks for such quick turnarounds - really appreciated.

This is the reason I stopped using Ansible - I created a *very* simple patch, and after 5 months the issue was eventually closed with no merge :(

https://github.com/ansible/ansible/pull/7220

lib/Rex/PkgConf/Debian.pm
```diff
@@ -0,0 +1,105 @@
...
+  my $conf_cmd = "debconf-show $pkg";
+
+  Rex::Logger::debug("Running command $conf_cmd");
+  my @lines = i_run $conf_cmd;
+
+  my %config;
+  for my $line (@lines) {
+    # Expecting: * postfix/relayhost: smtp.example.com
+    Rex::Logger::debug("Parsing line $line");
+    if ( $line =~ m!^(\*?)\h+(.+):\h*(.*)! ) {
```
ferki: This line also breaks perl-5.8.9 compatibility, so I'm going to replace `\h` with normal space.

lib/Rex/PkgConf/Debian.pm
```diff
@@ -0,0 +1,105 @@
...
+      my ($already_set, $question, $value) = ($1, $2, $3, $4);
```
ferki: What is contained in `$4` at this point and/or why it gets omitted from the assignment?

abeverley: Sorry, my bad. I originally parsed the postfix/relayhost as 2 separate values, but then later combined them for various reasons. You can remove the $4.

ferki: no problem, will do! ;)

PR #534, opened 2015-01-25
Add password lock/unlock functions (Linux only)

abeverley:
Add password locking and unlocking commands (```lock_password``` and ```unlock_password```). This command currently only works on Linux.

### DIFF
--- a/lib/Rex/User/Linux.pm
+++ b/lib/Rex/User/Linux.pm
@@ -24,10 +24,13 @@ use Rex::Interface::Exec;
 use Rex::Helper::Path;
 use JSON::XS;
 
+use Rex::User::Base;
+use base qw(Rex::User::Base);
+
 sub new {
   my $that  = shift;
   my $proto = ref($that) || $that;
-  my $self  = {@_};
+  my $self  = $proto->SUPER::new(@_);
 
   bless( $self, $proto );
 
--- a/lib/Rex/User/Linux.pm
+++ b/lib/Rex/User/Linux.pm
@@ -359,6 +359,57 @@ sub get_user {
   );
 }
 
+sub lock_password {
+  my ( $self, $user ) = @_;
+
+  # Is the password already locked?
+  my $result = i_run "passwd --status $user";
+
+  die "Unexpected result from passwd: $result"
+    unless $result =~ /^$user\h+(L|NP|P)\h+/;
+
+  if ( $1 eq 'L' ) {
+    # Already locked
+    return { changed => 0 };
+  }
+  else {
+    my $ret = i_run "passwd --lock $user";
+    if ( $? != 0 ) {
+      die("Error locking account $user: $ret");
+    }
+    return {
+      changed => 1,
+      ret     => $ret,
+    };
+  }
+}
+
+sub unlock_password {
+  my ( $self, $user ) = @_;
+
+  # Is the password already unlocked?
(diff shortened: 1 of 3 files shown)


krimdomu:
Hi @abeverley,

thanks for this patch. Would you mind to add a Rex::User::Base class and change all the Rex::User::* classes to inherit from this one, so we can add the lock and unlock methods there to print error messages for operating systems where the lock/unlock isn't supported?

abeverley:
That's fine, no problem.

Stupid question, but what's the best way to update this? Shall I do a new commit in my repo, then squash both commits, and resubmit the pull request?

abeverley:
Hmmm, not sure if I'm doing this right. I've added another commit, but I don't want to squash, as I'll then have to force-push, which is surely a bad idea to a public repo?

ferki:
@abeverley: thanks for the contribution!

you're right, squashing and then force pushing to a public repo is usually not a good idea, but for pull requests handled in their own feature branches it's usually OK (I suppose it only depends on the project's preferences)

though it looks like you committed your changes directly to your master branch, which is not a big problem, but using feature branches is highly recommended :) you might be interested in checking out our [contribution guide](http://rexify.org/contribute)

also, don't worry much about it, as while it means a bit more work for us, we can squash those commits together by manually merging your changes

abeverley:
Okay, thanks. So you're happy to deal with this one as it is? Or would you rather I squash and force?

I'll make sure I branch my repo for any future PRs!

Thanks,

Andy

krimdomu:
Hi Andy,

thank you much for change. I'm going to merge it now.

lib/Rex/User/Linux.pm
```diff
@@ -359,6 +362,57 @@ sub get_user {
...
 }
 
+sub lock_password {
+  my ( $self, $user ) = @_;
+
+  # Is the password already locked?
+  my $result = i_run "passwd --status $user";
+
+  die "Unexpected result from passwd: $result"
+    unless $result =~ /^$user\h+(L|NP|P)\h+/;
```
ferki: This breaks perl-5.8.9 compatibility, I'm going to change `\h` with normal space.

lib/Rex/User/Linux.pm
```diff
@@ -359,6 +362,57 @@ sub get_user {
...
+}
+
+sub unlock_password {
+  my ( $self, $user ) = @_;
+
+  # Is the password already unlocked?
+  my $result = i_run "passwd --status $user";
+
+  die "Unexpected result from passwd: $result"
+    unless $result =~ /^$user\h+(L|NP|P)\h+/;
```
ferki: same here

lib/Rex/User/Linux.pm
abeverley: On Sun, 2015-02-01 at 17:24 -0800, Ferenc Erki wrote:
> > +  die "Unexpected result from passwd: $result"
> > +    unless $result =~ /^$user\h+(L|NP|P)\h+/;
> 
> This breaks perl-5.8.9 compatibility, I'm going to change \h with
> normal space.

Maybe use \s instead? I just thought there might be times when it
returns more than one space or a tab (I haven't seen this, just thought
it might make the code less likely to break).

ferki: Yep, I chose `\s` for now, as that should also match both space and tab - it looks good so far :+1:

PR #535, opened 2015-01-26
Optimise Rex::Commands::Augeas

abeverley:
In the case of SSH, this patch groups Augeas commands and runs them together. Starting augtool is expensive, which is now done only once for each group of commands.

In addition, the commands are passed via a file, fixing bugs if the commands have characters that would otherwise need to be escaped.


abeverley:
Hmmm, somehow I have pulled in my other commits with this. I'll close and try again.

PR #536, opened 2015-01-26
Optimisations and bug fixes for Rex::Commands::Augeas

abeverley:
In the case of SSH, this patch groups Augeas commands and runs
them together. Starting augtool is expensive, which is now done
only once for each group of commands.

In addition, the commands are passed via a file, fixing bugs
if the commands have characters that would otherwise need
to be escaped.

### DIFF
--- a/lib/Rex/Commands/Augeas.pm
+++ b/lib/Rex/Commands/Augeas.pm
@@ -53,6 +53,7 @@ use Rex::Commands;
 use Rex::Commands::Run;
 use Rex::Commands::Fs;
 use Rex::Commands::File;
+use Rex::Helper::Path;
 
 use Config::Augeas;
 use IO::String;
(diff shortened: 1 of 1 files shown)


abeverley:
<sigh>, still struggling with this. I made a second commit (48aa9c2), with the intent of pushing that as a separate pull request. However, it's now ended up in this PR. Shall I create 2 new branches with each commit, or can you handle this as it is?

Thanks,

Andy

abeverley:
Likewise, let me know if you want me to start again and split these up. Sorry!

ferki:
@abeverley: a PR is about merging changes from one of your branches to one of our branches. If you change your branch, you also change the contents of the PR, even if you need to force push. That's how github works. In other words this means if you get the groupings of commits right in your repo, it will automatically update/clean up the according PRs you may already opened from your branches.

So, if the intention here is two different changesets, then yes, please use two branhces with a PR for each of the changesets. That makes review and merging process a lot easier.

If the second set of changes are built on top of the first set of changes, then I guess it's OK to keep them in a single PR.

abeverley:
Okay, got it, thanks. Given that all these commits are related, and they (almost certainly) depend on each other, I think it might be best to leave them as-is. They all relate to one command (Augeas) and there is only one file that has changed. I'll update the title of this PR as such. Thanks.

krimdomu:
Hi Andy,

thanks also for this patch. I'm going to merge it.

PR #541, opened 2015-01-30
CLI.pm: throw error if -T used with task arg, but no matching task found; Updates to docs

cpanxaoc:
- If the user passes in the -T switch with argument(s), Rex will search the task list in the Rexfile for a matching task(s) and print only those tasks; no error is given to the user if there are no matches between the user's arguments and the task list

Note that the -Tv and -Tm switches have their own block that handles arguments that don't have matching task names; the plain -T switch does not, currently

### DIFF
--- a/bin/rex
+++ b/bin/rex
@@ -36,7 +36,9 @@ Rex is a tool to ease the execution of commands on multiple remote servers. You 
 
 =item -H              Execute task on these hosts
 
-=item -G              Execute task on these group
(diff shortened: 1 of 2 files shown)


cpanxaoc:
The second commit updates docs for command line switches in bin/rex and CLI.pm.  Not sure if some of those are supposed to be hidden, or if the docs never got updated previously

PR #542, opened 2015-01-31
Add tests for append_or_amend_line

abeverley:
Add tests for append_or_amend_line

### DIFF
--- a/t/file.t
+++ b/t/file.t
@@ -5,7 +5,7 @@ use Cwd 'getcwd';
 my $cwd = getcwd;
 
 BEGIN {
-  use Test::More tests => 47;
+  use Test::More tests => 57;
   use Data::Dumper;
 
   use_ok 'Rex';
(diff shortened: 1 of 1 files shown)


abeverley:
The tests for this will fail until #543 is also merged

ferki:
@abeverley: thanks! I think it might worth to add an explicit test or two for the newline issues you fixed on #543, but other than that it looks good :+1:

abeverley:
I did add a test - that's why it's currently failing ;-) https://github.com/RexOps/Rex/pull/542/files#diff-88a5ea5ea46b4aa50141a2fd745c9e60R220

Or do you mean something else?

ferki:
oh, my bad..that actually checks for both 'extra empty line' situations

ferki:
Hmm..it looks like we just had a case of acute _mergis prematuritis_ as the tests were not only failing becuase of not-yet-implemented behaviour, but because breaking the test-suite for 5.8.9 as well. It seems like the `\h` backslash sequence is not supported there.

We'll fix this with a followup commit.

Notes to self and tips for others for the future:
- always check the test output when there's a failure (even if it's expected to fail :)
- try to keep tests and implementation/fix within a single PR (so it is easier to test the related parts together)

abeverley:
On Sun, 2015-02-01 at 14:57 -0800, Ferenc Erki wrote:
>       * always check the test output when there's a failure (even if
>         it's expected to fail :)

Is this something I should be doing? I did try to access the report, but
I needed credentials to view it (so I just assumed it was failing
because of the other required fix).

>       * try to keep tests and implementation/fix within a single PR
>         (so it is easier to test the related parts together)

Okay, will do.

Andy

ferki:
both were suggestions for anyone who reads it, but most importantly to myself :) and to @krimdomu :P
I'll double check the jenkins permissions as I thought the test results are publicly available

krimdomu:
@abeverley 

sorry this was my fault. i removed the public "viewability" some time ago and didn't add it back. Now the tests are visible again.

krimdomu:
The local tests can be found here: http://build.rexify.org/view/Local%20Tests/

abeverley:
On Sun, 2015-02-01 at 16:08 -0800, Jan Gehring wrote:
> The local tests can be found here: http://build.rexify.org/view/Local%
> 20Tests/

Great, thanks for that.

PR #543, opened 2015-01-31
Stop append_or_amend_line inserting extra blank lines

abeverley:
This patch fixes append_or_amend_line inserting extra blank lines. It also ensures that if an amended line is at the end of a file that the newline is not removed.


krimdomu:
Hi,

thanks for this patch. I've just merged it into master branch.

Issue #516, opened 2014-12-31
multiple groups execution on cli


Azryel:
Hi,

I needed to execute a task on multiple groups without creating a group with them into.
rex -h says :   -G|-g           Execute task on these group
Sooo.. these but no s.
Anyway it did not seem to work in any way, so i added that to Rex::TaskList::Base :

```diff
--- /usr/share/perl5/vendor_perl/Rex/TaskList/Base.pm.old       2014-12-31 10:41:16.000000000 +0100
+++ /usr/share/perl5/vendor_perl/Rex/TaskList/Base.pm   2014-12-31 10:40:11.000000000 +0100
@@ -69,6 +69,18 @@
 
         push @server, sub {
           if ( !Rex::Group->is_group($group_name) ) {
+            if ($group_name =~ /,/) {
+              my @groups = ();
+              for (split /,/, $group_name) {
+                if (Rex::Group->is_group($_)) {
+                  push @groups,Rex::Group->get_group($_);
+                } else {
+                  Rex::Logger::info( "No group $_ defined.", "error" );
+                  exit 1;
+                }
+              }
+              return @groups if @groups;
+            }
             Rex::Logger::info( "No group $group_name defined.", "error" );
             exit 1;
           }
```
It seems to work for me now just with a ",".
i've never posted any patch on github before, i don't know if this is any good, but i needed this feature :P

krimdomu:
Hi Azryel,

thank you for this patch and sorry for the late response.

I can manually merge this, if you want. This is no problem. 

The normal way is to do a fork of the rex repository into your private namespace. Then you can do your changes and send a pull request. With this it is easier to merge and the history in the commit logs shows you as the author of this snippet.

So if you want to do a merge request, just let me know, then i'll wait for it.

Cheers,
Jan

ferki:
Although it can be merged manually, I would prefer using a pull request as they can be tested before merging to master :)

Also this patch seems to duplicate code for checking if a group exists or not - there might be a cleaner solution for this problem.

reneeb:
I would prefer if one could specify multiple "-g"'s

rexify -g group1 -g group2

Currently the argument parser cannot handle "array" options and the use of $::FORCE_SERVER variable is a bit obscure (if it begins with \0 then it is groups, otherwise it can be a comma separated list of servers...) ;-)

If you agree, I would rewrite that handling:

[...]

PR #544, opened 2015-01-31
added possibility to define parameters multiple times. - #516

krimdomu:
Hi,

this is a patch allowing to define command lines switches multiple times.

so it would be possible to do the following:

```
rex -g group1 -g group2 mytask
```

This patch won't filter out duplicate servers. I think implementing this would result in a rewrite of the TaskList/Base.pm module.

But i think this is an acceptable approach until we rewrite the cli arg parser.

What do you ppl think?

### DIFF
--- a/lib/Rex/Args.pm
+++ b/lib/Rex/Args.pm
@@ -84,7 +84,16 @@ sub import {
           $rex_opts{$name_param}++;
         }
         else {
-          $rex_opts{$name_param} = $c->get;
+          # multiple params defined, create an array
+          if ( exists $rex_opts{$name_param} ) {
+            if ( !ref $rex_opts{$name_param} ) {
+              $rex_opts{$name_param} = [ $rex_opts{$name_param} ];
+            }
+            push @{ $rex_opts{$name_param} }, $c->get;
+          }
+          else {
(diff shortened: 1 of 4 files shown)


ferki:
It sounds good :+1: Why is the syntax not stg like `-g group1,group2` though?

ps.: it should be discussed separately, but I felt it worth to mention here, that in the long run, it might be a good idea to switch to `Getopt::*` or `MooX::Options` based argument parsing.

krimdomu:
Definitively. In the long run (-> rex 2) we will use something like this.

ferki:
I rebased it on top of current master in order to be able to test it against the changed tests.

ferki:
@krimdomu: are there any specific reasons for the `-g group1 -g group2` syntax vs the `-g group1,group2`?

krimdomu:
@ferki i like the syntax ```-g foo -g bar``` more, because this is the normal way (for example rsync --exclude).

reneeb:
krimdomu++

ferki:
ahh, thanks guys for the explanation and feedback, now I found the more detaild explanation over on #516 :+1:

let's drop in some tests for the new functionality and then :shipit:

krimdomu:
Added tests.

PR #545, opened 2015-01-31
Add a module to manage UFW (Uncomplicated Firewall)

abeverley:
This will currently only work when ufw outputs in English, as the output of the ufw command is checked against the English output that is expected. Being native English myself, I have little experience of i18n. Is there a way to force a command to output English, regardless of the host's language?

There are currently no tests. How should one test with ufw, without affecting the firewall rules of the host system?

Happy to resubmit this PR, depending on the answers of the above. Thanks!

### DIFF
--- a/lib/Rex/Commands/UFW.pm
+++ b/lib/Rex/Commands/UFW.pm
@@ -1,0 +1,376 @@
+#
+# (c) Jan Gehring <jan.gehring@gmail.com>
+#
+# vim: set ts=2 sw=2 tw=0:
+# vim: set expandtab:
+
+=head1 NAME
+
+Rex::Commands::UFW - UFW Management Commands
+
+=head1 DESCRIPTION
+
+With this Module you can manage UFW (Uncomplicated Firewall) rules.
+
+=head1 SYNOPSIS
+
+ use Rex::Commands::UFW;
+
+ ufw_enable; # Activate firewall
+
+ ufw_disable; # Disable firewall
+
+ ufw_logging 'on';
+
+ ufw_allow app => 'www';
+
+ ufw_allow app => 'www', delete => 1;
+
+ ufw_allow
+   from => '10.10.10.1',
+   app  => 'www',
+   to   => '192.168.0.1',
+   app  => 'www',
+   log  => 'new';
+
+ # Delete a rule
+ ufw_allow
+   from   => '10.10.10.1',
+   app    => 'www',
+   to     => '192.168.0.1',
+   app    => 'www',
+   log    => 'new',
+   delete => 1;
+
+ ufw_allow
+   proto     => 'tcp',
+   from      => '10.10.10.1',
+   port      => 22,
+   to        => '192.168.0.1',
+   port      => 80,
+   direction => 'in',
+   on        => 'eth0',
+   log       => 'all';
+
+  ufw_limit app => 'OpenSSH';
+ 
+=head1 EXPORTED FUNCTIONS
+
+=over 4
+
+=cut
+
+package Rex::Commands::UFW;
+
+use strict;
+use warnings;
+
+# VERSION
+
+require Rex::Exporter;
+
+use base qw(Rex::Exporter);
+
+use vars qw(@EXPORT);
+
+use Rex::Commands::Run;
+use Rex::Helper::Run;
+use Rex::Logger;
+
+@EXPORT = qw(ufw_enable ufw_disable ufw_logging ufw_allow
+  ufw_deny ufw_reject ufw_limit);
+
+=item ufw_enable
+
+Enable UFW firewall.
+
+=cut
+
+sub ufw_enable { _ufw_disable_enable('enable') }
+
+=item ufw_disable
+
+Disable UFW firewall.
+
+=cut
+
+sub ufw_disable { _ufw_disable_enable('disable') }
+
+sub _ufw_disable_enable {
+  my $action = shift;
+  _resource_start($action);
+  my $return = _ufw_exec('status');
+
+  my $needed = $action eq 'enable' ? 'inactive' : 'active';
+  if ( $return =~ /Status: $needed/ ) {
+    my $ret = _ufw_exec("--force $action");
+    my $success = $action eq 'enable'
+                ? 'Firewall is active and enabled'
+                : 'Firewall stopped and disabled';
+    if ($ret =~ /$success/) {
+      Rex::get_current_connection()->{reporter}->report(
+        changed => 1,
+        message => "Firewall successfully changed to $action",
+      );
+    }
+    else {
+      Rex::Logger::info( "Unexpected ufw response: $ret", "warn" );
+    }
+  }
+  else {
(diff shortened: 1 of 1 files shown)


abeverley:
Also, it uses the ```Rex::get_current_connection()->{reporter}``` interface. I'm happy to update this to the new ```resource``` interface, but I wasn't sure how it works. I know you previously emailed the Apache example, but are there any examples of core modules?

ferki:
I would argue that this would fit in the rex-recipes repo better. I read the question about it on the mailing list as well, and I guess it's time for us to think about it a bit more :)

abeverley:
Okay, but I was hoping to meet the criteria of "when can a module be in core". Admittedly it hasn't got tests (yet), but I'd be happy to be its maintainer.

I looked at the recipes, but I thought they were more like sets of rex rules, rather than modules in their own right.

From my point of view, it would be useful to have some way of installing Rex, and not have to individually install lots of other modules. I appreciate that you don't necessarily want lots of modules in core though.

2 thoughts:

- Could you have an "extras" package that groups them all together? Like the ansible-modules-extras package.
- Is it worth grouping all core modules into their respective types? Sorry to use the example of Ansible again, but it has groups for files, network, system, database etc.

Thanks!

krimdomu:
Hi Andrew,

thank you for this patch but we talked about the ufw module and we think we should add a level more of abstraction before merging it to core.

The ufw module is ubuntu specific and so if someone want to write a firewall rule for ubuntu he had to use the ufw functions. and (maybe later) there is a module for redhats firewall-cmd then the people have to make a decision by themself which one to use.

So the code would look something like this: (pseudo code)

```perl
task "prepare_firewall", sub {
  if(operating_system eq "Ubuntu") {
    ufw open => 80;
  }
  elsif(operating_system eq "RedHat") {
    firewall_cmd open => 80;
  }
};
```

But it would be better to have one more abstraction layer, so you can have something like this in your Rexfile.

```perl
task "prepare_firewall", sub {
  firewall
    ensure => "open",
    port   => 80;
};
```

So the firewall module will decide for itself which firewall module it needs to use. For Ubuntu -> ufw, for Redhat -> firewall_cmd, for FreeBSD -> pf, for a unknown linux -> iptables and so on.

So we first think of an interface for the firewall module, then we can implement it. If you have suggestions for it, feel free to comment or post code :-)

Normaly we are also available in irc, so irc is often better for longer discussions or brainstorming.

abeverley:
Okay, no problem, that makes sense. I did wonder that, but then I saw there was already an iptables command, and I thought that there might be too many specific options with different firewall tools.

I'm not sure when I'll next get a chance to look at this, but I will try to do soon.

abeverley:
Oh, incidentally, ufw isn't Ubuntu specific - from what I can tell it's also packaged for Suse and Gentoo (and of course Debian), and there's no real reason it couldn't be used on any iptables-based distro. But I get your point.

krimdomu:
Yes, the iptables module is from a time where rex only supported linux. In favor of the firewall module we need to bulid a proper deprecation path for this.

So i'll put this module onto 1.1 milestone, because we can't add these functions for the next release.

ferki:
TL;DR: @abeverley: no rush, take your time, and you can squash your commits currently in this PR ;)

Long version:

@abeverley: sorry if we haven't met your expectations by quickly merging this. I can assure you that your work on rex is highly appreciated (keep them coming! :), and even more importantly it really makes us think and discuss about interesting problems and decisions.

Currently there's no official rule or guideline about what goes into core and what is better suited as an external module - though there are some tendencies about this. Most probably there's an optimal balance between "rex is an execution engine, don't bloat it with support for specific software" and "everything should be maintained in core". There are several factors to this and I think it is mainly a design decision, and I also would like to invite you to our IRC channel to participate in the discussion. We like talking and meeting with community members :)

At first, I was thinking like "we probably won't add support for specific general purpose software, e.g. nginx, mysql, etc., directly to core as that pulls in a lot of stuff that feels like stg not directly in the scope of rex and probably would be considered as bloat by some users".

Then we discussed this topic with @krimdomu, and we think that moving `iptables` support behind an abstraction layer called `Rex::Firewall`, and adding UFW support there, would feel a bit more consistent with the current model already used elsewhere in the project (like for handling services or packages).

That being said, in the end, I think it would not be really different than your intial proposal as it would mostly be the same stuff under a different namespace and some wrapping/detection code...

So it is an interesting question to decide how to proceed with these kind of stuff, and personally I really like your "separate package for `rextras`" idea as well.

Until that decisions are made (or until a new pattern emerges/evolves), I would vote for keeping consistency as much as possible and go with the `Rex::Firewall` version.

I hope the above braindump gives you a useful context :)

abeverley:
On Sun, 2015-02-01 at 15:48 -0800, Ferenc Erki wrote:
> @abeverley: sorry if we haven't met your expectations by quickly
> merging this.

That's fine, you've more than met my expectations for my other PRs!

> and I also would like to invite you to our IRC channel to participate
> in the discussion. We like talking and meeting with community
> members :)

Thank you, I'll pop along at some point, just busy with a few other
things at the moment.

> Then we discussed this topic with @krimdomu, and we think that moving
> iptables support behind an abstraction layer called Rex::Firewall, and
> adding UFW support there, would feel a bit more consistent with the
> current model already used elsewhere in the project (like for handling
> services or packages).

Makes sense.

> That being said, in the end, I think it would not be really different
> than your intial proposal as it would mostly be the same stuff under a
> different namespace and some wrapping/detection code...

Yeah, I was thinking that. I reckon the API can be very similar, but
under a firewall namespace instead. I don't have the time to look at it
just now, but I will do.

> So it is an interesting question to decide how to proceed with these
> kind of stuff, and personally I really like your "separate package for
> rextras" idea as well.

That would be good, although in this particular case, I think the
firewall stuff might be in core anyway.

Thanks,

Andy

krimdomu:
i just added a first version of the firewall module to rex-recipes repository. I think for the first few versions it is better to have it there, because the api might change when we implement more providers for it.

currently it has only an iptables provider, i'm going to add the ufw provider tomorrow.

abeverley:
Thanks for that. I hadn't forgotten, just not had the time to look at this again! Will you use code similar to this patch? It works well, supports most of the UFW functionality and is idempotent.

krimdomu:
sure, i'm going to use the code from this patch

abeverley:
Great, thanks. Sorry for not doing it myself!

krimdomu:
no problem :)

ferki:
:+1:

krimdomu:
For the 1.3 release i'm going to include the firewall module into core.

krimdomu:
@abeverley i'm sorry i hadn't had the time to include this module into 1.3 now. we need to write some acceptance tests for this in rex-build repo. And also i don't want to remove the dependency to Rex::Ext::ParamLookup but this forces me to create a package for Devel::Caller for centos5/6.

krimdomu:
packages for Devel::Caller for CentOS 5+6 are now uploaded into the nightly repository

krimdomu:
This is now merged into master see: #805

PR #546, opened 2015-01-31
Ensure update_system always updates package repositories first

abeverley:
There is a mixture of behaviours with update_system. With some operating systems it will update the package repository cache first, with others it won't. This patch attempts to make the behaviour consistent, and always update repositories first, given that this is most likely what one would want to do when updating a system.

Disclaimer: I have only experience with apt, but have attempted to cover all operating systems.

### DIFF
--- a/lib/Rex/Pkg/Debian.pm
+++ b/lib/Rex/Pkg/Debian.pm
@@ -32,7 +32,7 @@ sub new {
     install_version =>
       'APT_LISTCHANGES_FRONTEND=text DEBIAN_FRONTEND=noninteractive apt-get -o Dpkg::Options::=--force-confold --force-yes -y install %s=%s',
     update_system =>
-      'APT_LISTCHANGES_FRONTEND=text DEBIAN_FRONTEND=noninteractive apt-get -y -qq upgrade',
+      'APT_LISTCHANGES_FRONTEND=text DEBIAN_FRONTEND=noninteractive apt-get -y -qq update && APT_LISTCHANGES_FRONTEND=text DEBIAN_FRONTEND=noninteractive apt-get -y -qq upgrade',
     remove            => 'apt-get -y remove %s',
     purge             => 'dpkg --purge %s',
     update_package_db => 'apt-get -y update',
--- a/lib/Rex/Pkg/SuSE.pm
+++ b/lib/Rex/Pkg/SuSE.pm
@@ -26,7 +26,7 @@ sub new {
   $self->{commands} = {
     install           => 'zypper -n install %s',
     install_version   => 'zypper -n install $pkg-%s',
-    update_system     => 'zypper -n up',
+    update_system     => 'zypper refresh && zypper -n up',
     remove            => 'zypper -n remove %s',
     update_package_db => 'zypper --no-gpg-checks -n ref -fd',
   };
--- a/lib/Rex/Pkg/Gentoo.pm
+++ b/lib/Rex/Pkg/Gentoo.pm
@@ -27,7 +27,7 @@ sub new {
   $self->{commands} = {
     install           => 'emerge %s',
     install_version   => 'emerge =%s-%s',
-    update_system     => 'emerge --update --deep --with-bdeps=y --newuse world',
+    update_system     => 'emerge --sync && emerge --update --deep --with-bdeps=y --newuse world',
     remove            => 'emerge -C %s',
     update_package_db => 'emerge --sync',
   };


ferki:
@abeverley: thanks again for keeping us busy thinking about your propsals and reviewing your changes! ;)

Would you think it would be still useful to make this optional? E.g. something like `update_system(sync_repos_first => TRUE);` or as a general behaviour setting like `set update_package_db_before_update_system TRUE:`.

Or perhaps even by introducing an `upgrade_system` function? Which calls both `update_package_db` and `update_system` in that order.

I think it would worth to keep those functions separate as they are atomic operations, but adding direct support for "please just upgrade everything you can" is really a good idea as I think most people creates their own wrappers anyway for exactly this (including me :).

I can think of an use-case for keeping them separate with Gentoo, when there could be USE flag changes that might require running the package upgrades, but a package repo sync would be totally unnecessary. It could also be true when working with overlays.

I'll comment some questions directly on the commits too.

lib/Rex/Pkg/Debian.pm
```diff
@@ -32,7 +32,7 @@ sub new {
     install_version =>
       'APT_LISTCHANGES_FRONTEND=text DEBIAN_FRONTEND=noninteractive apt-get -o Dpkg::Options::=--force-confold --force-yes -y install %s=%s',
     update_system =>
-      'APT_LISTCHANGES_FRONTEND=text DEBIAN_FRONTEND=noninteractive apt-get -y -qq upgrade',
+      'APT_LISTCHANGES_FRONTEND=text DEBIAN_FRONTEND=noninteractive apt-get -y -qq update && APT_LISTCHANGES_FRONTEND=text DEBIAN_FRONTEND=noninteractive apt-get -y -qq upgrade',
```
ferki: What is the reason behind using a different command for repo update here than the simple `update_package_db` case?

lib/Rex/Pkg/SuSE.pm
```diff
@@ -26,7 +26,7 @@ sub new {
   $self->{commands} = {
     install           => 'zypper -n install %s',
     install_version   => 'zypper -n install $pkg-%s',
-    update_system     => 'zypper -n up',
+    update_system     => 'zypper refresh && zypper -n up',
```
ferki: Same question here as above :)

abeverley:
On Sun, 2015-02-01 at 15:20 -0800, Ferenc Erki wrote:
> @abeverley: thanks again for keeping us busy thinking about your
> propsals and reviewing your changes! ;)

You're welcome ;-)

> Would you think it would be still useful to make this optional? E.g.
> something like update_system(sync_repos_first => TRUE); or as a
> general behaviour setting like set
> update_package_db_before_update_system TRUE:.

I don't really mind. The important thing is that it's documented and
consistent across all distros, which it is not at the moment.

My assumption from the documentation, before looking at the code, was
that a system upgrade would be done regardless (with Debian), but that
will not happen unless the repos happen to have been updated first.

> Or perhaps even by introducing an upgrade_system function? Which calls
> both update_package_db and update_system in that order.

Again, I don't really mind, but the problem is more likely going to be
functionality between distros. I don't use Redhat, but from what I read,
it is not possible to update the system without also updating the
package DB.

> I think it would worth to keep those functions separate as they are
> atomic operations, but adding direct support for "please just upgrade
> everything you can" is really a good idea as I think most people
> creates their own wrappers anyway for exactly this (including me :).

Well, there is also Debian's dist-upgrade, which I was going to deal
with separately, and was thinking of some sort of flag for this as you
suggest above. But again, the problem is compatibility with other
distros, as there is no equivalent. Arguably, update_system should call
dist-upgrade rather than just upgrade.

> I can think of an use-case for keeping them separate with Gentoo, when
> there could be USE flag changes that might require running the package
> upgrades, but a package repo sync would be totally unnecessary. It
> could also be true when working with overlays.
> 
> I'll comment some questions directly on the commits too.

In short, I'm happy either way, as long as it's obvious from the
documentation.

But like I say, I think it should be consistent between distros, which
might be difficult.

Thanks,

Andy

lib/Rex/Pkg/Debian.pm
abeverley: On Sun, 2015-02-01 at 15:22 -0800, Ferenc Erki wrote:
> In lib/Rex/Pkg/Debian.pm:
> 
> > @@ -32,7 +32,7 @@ sub new {
> >      install_version =>
> >        'APT_LISTCHANGES_FRONTEND=text DEBIAN_FRONTEND=noninteractive apt-get -o Dpkg::Options::=--force-confold --force-yes -y install %s=%s',
> >      update_system =>
> > -      'APT_LISTCHANGES_FRONTEND=text DEBIAN_FRONTEND=noninteractive apt-get -y -qq upgrade',
> > +      'APT_LISTCHANGES_FRONTEND=text DEBIAN_FRONTEND=noninteractive apt-get -y -qq update && APT_LISTCHANGES_FRONTEND=text DEBIAN_FRONTEND=noninteractive apt-get -y -qq upgrade',
> 
> What is the reason behind using a different command for repo update
> here than the simple update_package_db case?

Good point, calling update_pacakge_db would be better, but is that
possible with the current code design? It looks like you're defining a
system command to run, rather than a CODEREF. Should the hash be changed
to a hash of coderefs?

Andy

ferki:
I totally agree with your thoughts about consistency and the documentation :+1: Sounds like it is time to review and finetune what those modules do.

abeverley:
On Sun, 2015-02-01 at 15:20 -0800, Ferenc Erki wrote:
> Would you think it would be still useful to make this optional? E.g.
> something like update_system(sync_repos_first => TRUE); or as a
> general behaviour setting like set
> update_package_db_before_update_system TRUE:.
> 
> Or perhaps even by introducing an upgrade_system function? Which calls
> both update_package_db and update_system in that order.

Thinking about this further, and given the difference between different
distros, I think update_system should update the repos by default.

I can't think of any reason why one would want to update a system before
updating the repos, as the chances are that it will do nothing. And also
because that's the default behaviour of other distros anyway.

Thanks,

Andy

abeverley:
I'm not going to rebase this PR just yet, until a firmer way ahead has been established.

krimdomu:
Hi,

i think Andy is right, ```update_system``` should also run ```update_package_db``` because this is what people expect.

Also we should not add more options to control the behaviour. I can't imagine a situation where you don't want to run an ```update_package_db``` before the update.

But we should not add the package db update command to the system update command. We can just use the already defined command in the function.

I'm going to add this for the 1.0 release.

krimdomu:
@ferki 
for the gentoo thing you mentioned. I don't think that we need to take care of it. If someone don't want to update the package database (due to concerns regarding stability or other things) they usually use a local mirror that can be tagged.
Also, if someone really want to stick for a special version of a package he must include the version in the ```install``` or ```pkg``` command. Because he can never be sure if someone runs an "update package db" manually on one of the servers.

Maybe i'm also wrong with this :) 

So for compatibility reasone, to not change the behaviour, i'm going to add a feature flag for this.

krimdomu:
@ferki is 75739d4 okay for you?

This will introduce a new feature flag *1.0*. If this is activated the ```update_system``` will also do a ```update_package_db```.

senorsmile:
As a rather new user, I'd like to put in my opinion: 

I am in charge of patching about 1,000 RHEL and Ubuntu servers.  We do have a multi-tiered staging for pulling in updates and releasing them to dev and production.  However, there are often times when we rely on the fact that "apt-get update" hasn't been run yet, and knowing that OUR repos are not out of date (i.e. the old packages are definitely still available) we can still install an older version.  I agree that this does not matter on RHEL, but that's just sort of annoying, to be honest.  

A further example: all of my private servers run Arch Linux (something I'd like to add some PR's for once I get my head around the code base for Rex).  If I have five different tasks that all install different packages, I do NOT want to run the equivalent of update package db (-Syy) just to install more packages.  Knowing that it has been done once is sufficient.  In Ansible (I know, I know) all of the supported package manager modules have a separate param: (update=yes, defaults to no) that you can run once per server.  If it were run on every server for every role (in Rex task), this would slow down every single run significantly.  

Just my two cents.

ferki:
TL;DR: I support adding functionality to "please just upgrade everything you can" and I find the idea very useful, but please do it so that the current behaviour doesn't change with it.

I'm fine with adding functionality which does both of updating the package information _and then_ updates all packages. This is a very commonly used stuff, it makes sense to add direct support for it, and I agree with the idea. So people (including me) can stop writing wrapper tasks around them containing both, but nothing else.

On the other hand, I want to avoid forcing this as default behaviour, or more importantly, I want to reserve the freedom of choice and the control over those operations.

Updating the package repos and updating all packages are most probably kept as separate functions for a reason by all the distros and package managers, and I'd like to keep the same according fuctionality within rex, which only provides an OS-agnostic layer over the according commands per OS.

My proposal is to add a new function called stg like `upgrade_all_packages` or `upgrade_system` or similar, which only just runs the already existing `update_package_db` and `update_system` commands.

That should be:
- relatively easy to implement
- easy to document
- doesn't change current behaviour (so no need for feature flags)
- still does what we expect from it

In the gentoo use case I was talking about USE flags: if I want to compile packages with different settings, I just change the USE flag settings in `make.conf` and do a world update (=`update_system`). Totally no need to update package repository infos, moreover it's even considered abusive by some gentoo mirrors to sync more than once a day (though it's usually not enforced in practice and/or you can just pick another mirror).

Or I can add an ebuild to the local repo overlay on the hosts I manage, and want to install them. Again, no need to update the whole package repository info.

Or if I want to parallelize host upgrades, I most probably don't want to trigger loads of connections and/or traffic against any mirror, even if I have my own repo mirror.

Probably there are similar reasons for other distributions as well.

krimdomu:
Hi all,

thank you much for all your thoughts and time you take to answer on this thread, this is really cool :+1: 

On my way back from fosdem i had some time to think about this more and i would vote for the first proposal from @ferki  (adding a parameter to the ```update_system``` function).

And indeed, it is better to give the user the ability to decide what he wants to do. Only a metadata update or also a system update.

I would vote against one more function, because i think a parameter (like proposed in the first comment) is not so confusing. Maybe we can also go one step further in the next major version of Rex to have only 1 function with 2 (or more [depending on the package manager]) parameters.

Something like this:

```perl
update_system
  update_metadata     => TRUE,
  update_packages     => TRUE,
  dist_upgrade        => TRUE,   # for debian systems for example.
  use_flag_update     => TRUE,   # for gentoo systems
  and_maybe_some_more => FALSE;
```

ferki:
+1 for the options of update_system :+1:

+1000 for @abeverley for coining this

A variation for the syntax:
```perl
update_system
  update_package_db => TRUE;
```

Also, probably the distro-specific options should just fail silently when they are being passed for a system for which they do not apply, e.g. `dist_upgrade` on a Red Hat system. That way the authors of the rex modules/code can reuse their code on different systems without modification.

Bonus reason against doing both update and upgrade in a single step: that way, in the general case, the administrator cannot be sure about what packages will be changed on their system, which can be potentially dangerous.

Also generally "the admin knows better [what needs to be run]" - let's give them the power to decide what's best for their use-case.

\<fun\>Maybe we can call the option stg like `I_know_what_I_am_doing_and_if_something_breaks_I_get_to_keep_all_the_pieces => TRUE` :)\</fun\>

ferki:
I also think that I should just shut up, get some sleep and read all the docs before I write stuff, instead of doing it off the top of my head :) `yum update` does seem to upgrade the package metadata and then install available upgrades (though it might be needed to `yum makecache` before and/or `yum clean metadata` to make sure that it does not use cached information?).

The main principles remain the same though:
1. don't force a specific distro's way of doing things onto other distros ~~ "do what makes sense on the given distro"
2. the same rex function call should do equivalent operations on different distros

We can definitely improve the current situation in various ways, but I'm not 100% sure if we can resolve all conflicts in this case - so what we can't resolve, we should carefully document, along with the description of the problem, the chosen and/or the alternative solutions, and possibly the arguments pro and contra for those solutions.

lib/Rex/Pkg/Debian.pm
ferki: When we will bump into a limitation because of the currrent implementation, that conversion totally sounds like a good idea :+1:

ferki:
@senorsmile: thanks for your input on this topic! We always like to hear/read more use cases :+1: Also it looks like we're thinking very similarly about these matters :)

krimdomu:
Hi @abeverley ,

after about 47105840 seconds the `update_system` function is now extended with the things i mentioned in my comment from Feb. 3, 2015. 
With the next release this can be used when the feature flag `1.5` is loaded.

PR #547, opened 2015-01-31
Die if trying to run augeas without augtool installed

abeverley:
If augtool is not installed the command will fail silently and appear to function correctly


ferki:
It looks simple enough, but let's wait a bit until we get passing tests again.

ferki:
@abeverley: could you rebase this branch on top of current master, please? So it can be tested against the new fixed tests. You need to force push after this.

ps.: probably worth doing the same for your other PRs as well

abeverley:
Okay, done that for this, let me know if there is anything else you need.

PR #549, opened 2015-02-02
no hard dependency for DBI anymore

vkroll:
* if DBI is not installed, the test will all be skipped


ferki:
@vkroll: thanks for the hotfix, I just merged this manually (after amending the commit message)

PR #561, opened 2015-02-13
stop 'profiling' before returning from a subroutine.

andrejzverev:


krimdomu:
Hi,

thanks for this patch. I've just merged it into master.

PR #562, opened 2015-02-13
Warn if Augeas command fails

abeverley:
Currently if Augeas does not exit correctly, then it will silently fail. This patch ensures that a warning is thrown.


krimdomu:
Hi Andy,

thanks for this patch. I've just merged it into master.

PR #567, opened 2015-02-15
dist.ini: Add Twitter and IRC links to metaresources block

cpanxaoc:
The IRC metaresource will get you a pretty red banner on your MetaCPAN page for Rex that is clickable; when clicked in a browser, you are taken to mibbit.org for a web IRC session.  You may not actually want this "feature", but I thought I'd throw it out there in any case.


krimdomu:
Hi,

thanks for the addition. I'm going to merge it. In the past we already had links to a irc webchat and had some troubles with spam. But i'll give it a try :)

PR #568, opened 2015-02-15
Rex::Helper:Run->i_run: check no_path_cleanup before calling get_path

cpanxaoc:
- Rex::Commands::Run->can_run() calls into Rex::Helper::Run->i_run(),
  which doesn't check for feature 'no_path_cleanup', causing commands
  outside of the Rex "safe paths" to not be found when can_run() was
  called, but the same commands could be found when
  Rex::Commands::Run->run() is called with
  "-feature => no_path_cleanup" set

### DIFF
--- a/lib/Rex/Helper/Run.pm
+++ b/lib/Rex/Helper/Run.pm
@@ -69,7 +69,11 @@ sub i_run {
     $is_no_hup = 1;
   }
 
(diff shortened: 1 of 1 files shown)


krimdomu:
Hi,

thanks four this fix. I've merged it into master branch.

PR #569, opened 2015-02-15
Rex::Commands::MD5: Make Rex use the /sbin/md5 binary on OS X

cpanxaoc:
- Mac OS X (uname -s = 'Darwin') also has an MD5 checksum command located at '/sbin/md5', same as the existing check for "uname -s" = "BSD" in Rex::Commands::MD5


krimdomu:
Hi,

thanks four this patch. I've merged it into master branch.

PR #575, opened 2015-02-21
Add preferred type option for mount command (fix #469)

### LINKED ISSUE #469: Type option for mount()
What we can specify as `fs` option in Rex for `mount()`, is actually called `type` in `mount` terminology and this can be confusing, especially when one tries to use bit more fancier mounts, like `mount -t proc ...`.

I propose
- adding a `type` option for `mount()` (to be in sync with the terminology)
- which should have precedence over the already existing `fs` option (if someone uses `type` he/she probably _means_ it)
- with printing a warning about this precedence when both are specified.

@krimdomu: what's your opinion?

ferki:
@jfried: please review and merge/discuss


krimdomu:
yes, lets merge this now.

PR #585, opened 2015-03-08
Speed up connecting to Boxes

ferki:
@krimdomu: please review and merge/discuss

There was a fixed amount of delay before SSH was considered as up and running for Boxes. During development this could be very annoying by forcing this wait on every `rex Test:run` call, while the Box might already be up and running.

I think the cases when it might have been required to give some time for SSH to settle, e.g. just after startup, should be handled just fine by normal SSH retry attempts (fixme :)


ferki:
Oops, opened by mistake with the previous hub pull-request message still in cache :)

PR #586, opened 2015-03-08
Ensure proper return values for is_{dir,file} functions (fix #584)

### LINKED ISSUE #584: Rex::Test::Base:has_file passes on non-existing file
`$t->has_file('/not/existing/file');` passes with ok, but with no test description if the file does not exist.

`$t->ok( is_dir('/not/existing/dir') );` behaves similarly, so I suspect something around the return values of `is_file()` and `is_dir()`.

I used Net::OpenSSH when I experienced this.

ferki:
- explicit return values
- accept sockets as files


krimdomu:
this is a good patch :+1:

PR #587, opened 2015-03-08
Allow has_stat test to handle directories (fix #582)

### LINKED ISSUE #582: Rex::Test::Base::has_stat does not work on directories
It explicitly checks if the passed argument is a file, while it could easily work on directories as well.

ferki:


krimdomu:
i think this is okay, but maybe we also want to allow making a stat() call to block devices or character devices, ...

maybe it is better to do something like this:

```perl
my %stat = stat $path;
if( ! %stat ) {
  $self->ok(0, "File $path not found.");
}
```

ferki:
I think it's already handled by `Rex::Interface::Fs` objects as `is_file()` methods actually returning true for the following types:
- regular file
- links
- block devices
- character devices
- pipes
- unix sockets

 Which is basically all stuff that should be treated as file, except directories, which we handle separately with `is_dir()`. I think that is pretty much for which we can get a stat.

Though it might be more efficient to just try to run a single `stat` than running potentially both `is_file` and `is_dir`, so I think I'll update it anyway before merging.

ferki:
@krimdomu: I updated te PR with the single call to `stat` you recommended. This lead me to drop in some extras (like the `diag` method and the explicit PASS or FAIL output at the end), and to also fix other buggy cases with has_stat (like when the checks UID was not existing on the remote machine).

Upon successfull tests, I'm gonna merge it all at once.

PR #588, opened 2015-03-08
Drop potentially dangerous --force-yes option (fix #559)

### LINKED ISSUE #559: Prevent silently accepting and installing unsigned packages
Hi guys,

Just dropping this in here, as much as a reminder for me to look at sometime (I don't have the time now).

From the code, it looks like when installing a package, a force install is always done (?). Similar to the discussion on #546, I don't think this should be the default. For example, if there was a problem authenticating the package, then you wouldn't normally want to force install. As it stands, this will happen silently, and you would know no better that there was a problem.

ferki:
It needs testing as I'm not 100% sure if it would cause `apt-get` to wait for input or not.

### DIFF
--- a/lib/Rex/Pkg/Debian.pm
+++ b/lib/Rex/Pkg/Debian.pm
@@ -28,9 +28,9 @@ sub new {
 
   $self->{commands} = {
     install =>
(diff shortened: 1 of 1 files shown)


krimdomu:
i think this is a good patch. we should merge it.

Issue #602, opened 2015-03-21
Rex::Interface::Shell should be extensible


krimdomu:
To allow different shells we need to extend Rex::Interface::Shell to be extensible with modules/plugins.

PR #603, opened 2015-03-21
added code to make shells pluggable. - RexOps/Rex#602

krimdomu:
Code for #602


ferki:
What happens if it doesn't find any of the defined shells?

krimdomu:
good question... i've forgot the default

krimdomu:
now with a default :)

Issue #600, opened 2015-03-21
File::extract command should create target directory if it's not present


bechtoldt:
Related LoC: https://github.com/RexOps/Rex/blob/master/lib/Rex/Commands/File.pm#L1153

I could provide a PR that adds a check whether the target directory exists. If it's not, extract() is going to create it.

PR #604, opened 2015-03-21
mkdir $to in Commands::File::extract() if $to doesn't exist yet, refs #600

bechtoldt:


ferki:
@bechtoldt: thanks for your patch! Unfortunately it didn't pass tests yet. I'm going to leave notes directly on the commits about how to fix them.

lib/Rex/Commands/File.pm
```diff
@@ -21,7 +21,7 @@ With this module you can manipulate files.
    }
    $fh->close;
  };
- 
+
```
ferki: Looks like you changed all the POD sections fro some reason. I beleive it was just changed by your editor automatically. Can you send a commit that does not remove those spaces?

lib/Rex/Commands/File.pm
```diff
@@ -314,7 +314,7 @@ Furthermore, if a path prefix matches multiple prefix entries in 'path_map',
 e.g. "files/etc/ntpd.conf" matching both "files/" and "files/etc/", the
 longer matching prefix(es) have precedence over shorter ones. Note that
 keys without a trailing slash (i.e. "files/etc") will be treated as having
-a trailing slash when matching the prefix ("files/etc/"). 
+a trailing slash when matching the prefix ("files/etc/").
```
ferki: OK, this trailing space removal is needed, but out of the scope of the topic at hadn (creating target directory before exatracting stuff), so if possible I'd put it into a separate commit.

lib/Rex/Commands/File.pm
```diff
@@ -1112,14 +1112,17 @@ sub _append_or_update {
 
 =item extract($file [, %options])
 
-This function extracts a file. Supported formats are .box, .tar, .tar.gz, .tgz, .tar.Z, .tar.bz2, .tbz2, .zip, .gz, .bz2, .war, .jar.
+This function extracts a file. The target directory $to (if specified) will be created automatically
+if doesn't exist yet.
```
ferki: I love the fact that you didn't forget about updating documentation! :+1: This line break seems a bit odd though.

t/commands/file/extract.t
```diff
@@ -0,0 +1,41 @@
+use strict;
+use warnings;
+
+use Cwd 'getcwd';
+my $cwd = getcwd;
+
+BEGIN {
+  use Test::More tests => 57;
```
ferki: Looks like you planned 57 tests, but there are much less implemented.

t/commands/file/extract.t
```diff
@@ -0,0 +1,41 @@
+use strict;
```
ferki: Bonus points for tests! ^^

t/commands/file/extract.t
```diff
@@ -0,0 +1,41 @@
...
+  Rex::Commands::File->import;
+  Rex::Interface::Exec;->import;
+}
+
+my $tmp_dir = "/tmp";
+if ( $^O =~ m/^MSWin/ ) {
+  $tmp_dir = $ENV{TMP};
+}
+
+my $filename = "${tmp_dir}/test-$$.tgz";
```
ferki: What happens to this file on Windows (where the path separator is `\`)?

t/commands/file/extract.t
```diff
@@ -0,0 +1,41 @@
...
+my $to = "myfile_extracted/";
+
+my $exec = Rex::Interface::Exec->create;
+my $cmd = "tar cfvz ${filename} /usr/local/bin/";
```
ferki: What happens in `tar` is not available on the system on which the tests are being run? Or when there's no `/usr/local/bin` directory?

t/commands/file/extract.t
```diff
@@ -0,0 +1,41 @@
...
+$exec->exec($cmd);
```
ferki: Can it be that a simple `run` command would be adequate here instead of the dance with `Rex::Interface::Exec`?

t/commands/file/extract.t
```diff
@@ -0,0 +1,41 @@
...
+
+ok( is_file($filename), "file '${filename}' created" );
+
+extract(
+  $filename,
+  owner => "root",
```
ferki: What happens if the tests are being run with a non-root user?

lib/Rex/Commands/File.pm
krimdomu: No, if this is removed, pod will think that the example code doesn't belong together. those spaces are needed.

ferki: Yes, the spaces in the POD code sections are needed (=where it was the only content on the line). But I believe on this single line here, this single trailing space is safe to remove.

t/commands/file/extract.t
bechtoldt: just copied it from another file, I'll fix this

t/commands/file/extract.t
bechtoldt: Windows wasn't in the scope of this PR, don't have a windows box here, sorry.

t/commands/file/extract.t
bechtoldt: sure, I'll try to correct it

t/commands/file/extract.t
ferki: No problem, but Rex can be installed and run from Windows, just like almost every other CPAN distribution (or on Mac OS X, for that matter). And upon installing from CPAN it will run the tests and we'd like to avoid failing them. We can help modifying those parts and provide testing (and get lots of automated test outputs from CPAN testers.

The only important thing about this topic is that we can't assume that it will only run on Unix-like systems, so we try to use OS-agnostic, built-in file path manipulation tools like `File::Spec`.

t/commands/file/extract.t
ferki: Maybe the easiest way to resolve it is to supply a tiny example archive file in the `t/` directory and just try to work with it.

ferki:
@bechtoldt: I squashed your commits together upon merge, slightly modified the documentation part and reworded the commit message. Your changes will be included in the next release. Thanks for your contribution!

Issue #554, opened 2015-02-04
hide passwords in debug output


ferki:
We often ask for debug output of a rex run during support cases, and we always have to give the warning to remove sensitive data from it first. We could help this step by providing a `--hide-passwords` or similar CLI option.

Of course, `sensitive data` might mean different things for different users, but still, hiding some of the common cases might help streamlining the support cases. That means we should document what is hidden, so people can decide if they want to remove more stuff before sending the debug output.

Alternatively we could create an `anonymizer` for debug output, which can be applied to debug outputs just before sending them.

petersonchen:
+1 

I need this feature as well , I develop some task or libs and share with my MIS team , when run task failed ,  I want to clarify the issue  but  I don't want to ask him to run with -d  , because It's difficult to explain and to ask him to remove passwords in log file everytime.

krimdomu:
There is now a new function inside the Rex::Logger module. 

```perl
Rex::Logger::masq("This is data: %s -> %s", "sensitiv-data", "more-sensitiv-data");
```

This will replace the sensitiv data with stars (*). 

I have added this function to the places where i know that we will print out sensitiv data. If there are some places i didn't found yet, please open an issue for them.

PR #606, opened 2015-03-21
added a Rex::Logger::masq() function that can masq sensitiv data for logging output. (fix for RexOps/Rex#554)

krimdomu:
If someone want to print out also the sensitiv data, he can add an environment variable. I don't want to add an cli option for this, because the environment variable rises more attention than a cli option.

```
REX_DEBUG_INSECURE=1 rex ...
```

references: #554

### DIFF
--- a/lib/Rex/Logger.pm
+++ b/lib/Rex/Logger.pm
@@ -264,6 +264,19 @@ sub format_string {
   return $line;
 }
 
+sub masq {
+  my ( $format, @params ) = @_;
+
+  return $format if scalar @params == 0;
+  return $format if scalar( grep { defined } @params ) == 0;
+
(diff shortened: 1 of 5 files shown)


lib/Rex/Logger.pm
```diff
@@ -264,6 +264,23 @@ sub format_string {
...
+
+  return $format if scalar @params == 0;
+  return $format if scalar( grep { defined } @params ) == 0;
+
+  if ( exists $ENV{REX_DEBUG_INSECURE} && $ENV{REX_DEBUG_INSECURE} eq "1" ) {
+    return sprintf( $format, @params );
+  }
+
+  for my $p (@params) {
+    $p =~ s/./*/g;
```
ferki: We might also just replace it with a fixed length string to avoid giving hints about the length of the masked sensitive data.

Issue #608, opened 2015-03-22
return values of is_file() and is_dir() are wrong


krimdomu:
return values if file or directory is not found must be ```undef```. (see tests for #495)

we also have to write some tests for #584

PR #609, opened 2015-03-22
fix for #608 - wrong return value for is_file and is_dir

krimdomu:
See #608


krimdomu:
A test for #584 was also added to RexOps/rex-build

PR #612, opened 2015-03-24
Describe platform-specific dependencies

ferki:
Let's depend on Net::SSH2 only on Windows systems, and on Net::OpenSSH + Net::SFTP::Foreign on anything else. Rex defaults to this latter pair of modules already if they are installed.


krimdomu:
Oh, this is cool!

Issue #499, opened 2014-12-02
Configurable behaviour for CMDB


ferki:
Currently the CMDB module iterates through a list of candidate files for a queried key, and returns with the first match for it. It would be nice if the CMDB could support a merging behaviour as well, e.g. iterating through the same files and merging the same keys from different files. Of course this behaviour should be made optional, probably along with choosing a merge strategy and/or file processing order.

PR #619, opened 2015-04-02
added YAML DeepMerge CMDB - fixes RexOps/Rex#499

krimdomu:
This adds a DeepMerge YAML CMDB.

To load it:

```perl
set cmdb => {
  type => 'Rex::CMDB::YAML::DeepMerge',
};
``` 

To set merge behavior:
```perl
set cmdb => {
  type => 'Rex::CMDB::YAML::DeepMerge',
  merge_behavior => 'SOME-TYPE',
};
```

The merge is done with Hash::Merge (https://metacpan.org/pod/Hash::Merge) so the behavior can also be customized.

### DIFF
--- a/lib/Rex.pm
+++ b/lib/Rex.pm
@@ -597,15 +597,7 @@ sub import {
 
         Rex::Commands::set(
           cmdb => {
-            type => "YAML",
-            path => [
-              "cmdb/{operatingsystem}/{hostname}.yml",
-              "cmdb/{operatingsystem}/default.yml",
-              "cmdb/{environment}/{hostname}.yml",
-              "cmdb/{environment}/default.yml",
-              "cmdb/{hostname}.yml",
-              "cmdb/default.yml",
-            ],
+            type => "Rex::CMDB::YAML::Default",
           }
         );
 
--- a/lib/Rex/CMDB.pm
+++ b/lib/Rex/CMDB.pm
@@ -94,6 +94,10 @@ sub cmdb {
     $klass = "Rex::CMDB::$klass";
   }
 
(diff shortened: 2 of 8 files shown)


ferki:
@krimdomu: would you consider specifying the current behaviour as a custom Hash::Merge behaviour and use that by default (to provide backwards compatibility), instead of splitting the YAML module into a Default and a DeepMerge variant? I think it can be done with `specify_behaviour` of Hash::Merge.

Or am I missing some reason to keep them separate? :)

krimdomu:
well the only reason is, that with the separation there is no possibility to break the old code :)

ferki:
Well, in my opinion that's why we have tests for the current behaviour, so we can be sure that we don't break compatibility either way ;)

If I'm not mistaken the current behaviour is to "always keep the left value" (=the first value found), right? That sounds simple enough to describe and use as default merge behaviour, and also easy to override if `merge_behaviour` is passed.

If you're willing to merge it this way, I'm willing to code it :) I'd like to do this because I believe this would simplify long-term maintenance.

ps.: we also can probably easily support passing a hashref to `merge_behaviour`, and calling `specify_behaviour` in this case, allowing users to describe arbitrary merging rules on top of the builtin ones.

krimdomu:
@ferki sure, we can merge both modules into one. but then we should default to "no merge". So that the current behaviour doesn't change.

ferki:
@krimdomu: of course, maybe above I left out the bit that the default should/would be the current behaviour (first value found wins) ^^

ferki:
see also #632

ferki:
fixed with #632

PR #620, opened 2015-04-02
stat() should return a hash or undef.

mbroadhead:
This previously would return 0 for non-existing files. When called like 'my %stat = $fs->stat("foo")', you would get 'Odd number of elements in hash assignment' perl warnings.

### DIFF
--- a/lib/Rex/Commands/Fs.pm
+++ b/lib/Rex/Commands/Fs.pm
@@ -522,7 +522,11 @@ sub stat {
   Rex::Logger::debug("Getting fs stat from $file");
 
   my $fs = Rex::Interface::Fs->create;
-  %ret = $fs->stat($file) or die("Can't stat $file");
+
+  # may return undef, so capture into a list first.
+  my @stat = $fs->stat($file);
+  die("Can't stat $file") if !defined $stat[0];
+  %ret = @stat;
 
   return %ret;
 }
--- a/lib/Rex/Interface/Fs/HTTP.pm
+++ b/lib/Rex/Interface/Fs/HTTP.pm
@@ -77,6 +77,8 @@ sub stat {
   if ( $resp->{ok} ) {
     return %{ $resp->{stat} };
   }
+
+  return undef;
 }
 
 sub is_readable {
--- a/lib/Rex/Interface/Fs/Local.pm
+++ b/lib/Rex/Interface/Fs/Local.pm
@@ -131,6 +131,7 @@ sub stat {
     return %ret;
   }
 
+  return undef;
 }
 
 sub is_readable {
--- a/lib/Rex/Interface/Fs/OpenSSH.pm
+++ b/lib/Rex/Interface/Fs/OpenSSH.pm
@@ -117,7 +117,7 @@ sub stat {
   my $sftp = Rex::get_sftp();
   my $ret  = $sftp->stat($file);
 
-  if ( !$ret ) { return; }
+  if ( !$ret ) { return undef; }
 
   my %ret = (
     mode  => sprintf( "%04o", $ret->perm & 07777 ),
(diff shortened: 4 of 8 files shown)


ferki:
@mbroadhead: thanks for your patch! I think returning undef would be good in this case, though I believe  using an explicit `return undef;` would be even better. Could you please update the pull request and make sure all the `Rex::Interface::Fs` modules are behaving the same way?

I also have two questions before merging:
* Do you think there would be other cases than "file does not exist", when `stat` would fail? It seems we're not explicitly checking for that exact situation, but the implementation seems more like "if nothing else worked, just return". I'm just not sure if that's intentional.
* Would returning a single `undef` value solve the original problem (e.g. returning a different, but still single `0` value causes "Odd number of elements in hash assignment")? Can it be that returning an empty hash would be more appropriate? Or throwing an error? Or perhaps returning a hashref instead of hash (that would probably break backwards compatibility though)?

@krimdomu: I think the above questions are interesting to think about/discuss before changing `stat` behaviour

mbroadhead:
@ferki:
> Do you think there would be other cases than "file does not exist", when stat would fail?

I haven't researched this, but I imagine other failures could arise -  file permissions, OS stat system call could be missing, etc.

> It seems we're not explicitly checking for that exact situation, but the implementation seems more like "if nothing else worked, just return". I'm just not sure if that's intentional.

You are correct, all of these Fs modules follow an "if nothing else worked, just return" pattern.

> Would returning a single undef value solve the original problem (e.g. returning a different, but still single 0 value causes "Odd number of elements in hash assignment")?

No, returing a single value of undef doesn't solve the original problem. My test case did not handle it. I think returning an empty hash is more appropriate. Rex::Interface::Fs::Sudo->stat() returns an empty hash. I think all other Rex::Interface::Fs modules should also return an empty hash if they can't obtain stat information.

I have fixed all the Rex::Interface::Fs modules to return an empty hash explicitly if no stat information could be found.

What are your thoughts @ferki and @krimdomu?

ferki:
@mbroadhead: thanks for the updates and for thinking about the matter together! Although I would probably have used a formal `return {};`, it looks good to me.

Now I'm leaning more and more towards the idea that when something unexpected happens during stat call:
- returning an explicit undef from the `Rex::Interface::Fs` modules would be OK as a sign of "we don't know" (perhaps adding a debug level message about this would be nice though)
- the upper level `Rex::Commands::Fs::stat()` call would/could still detect this case explicitly and throw an error/issue a warning
- at the same time we should ensure that `Rex::Commands::Fs::stat()` always returns a hash

Of course, using empty hashes to signal the "we don't know" case is just a slight variation of this. Looks like the current stats is a mix of the two.

@krimdomu: how it would be best to unify the behaviour in your opinion?

krimdomu:
@ferki i think your approach is good.
The `Rex::Interface::Fs` methods should return undef if they fail and `Rex::Commands::Fs::stat()` should detect this and throw an exception.
`Rex::Commands::Fs::stat()` should always (except if an exception is throws) return an hash. (more precisely a list of zero items or a list of even items)

ferki:
@mbroadhead: sorry for the extra loops, but looks like your proposal raised more questions than I originally thought ^^ Can you please modify your pull request based on the above discussion?

Alternatively we can amend your commit before merging it (still giving credits to you for the fix, of course :)

mbroadhead:
@ferki: I've modified my pull request to what I think @krimdomu explained. Feel free to modify as needed. Let me know if you have any feedback.

ferki:
@mbroadhead: thanks for the update! This is now merged. In a separate commit I added explicit checks for number of returned elements just to be sure Rex can catch weird unexpected outputs.

PR #624, opened 2015-04-07
Allow single-character tasknames (fix #621)

### LINKED ISSUE #621: Rex warns unnecessarily on tasks which are only 1 character long
This is the warning:
```
[2015-04-06 20:56:23] WARN - Please use only the following characters for task names:
[2015-04-06 20:56:23] WARN -   A-Z, a-z, 0-9 and _
```

ferki:


kablamo:
Nice!  Thanks

PR #641, opened 2015-04-23
Fail if requested environment is not found

ferki:
@krimdomu: it should work™, but most probably can be done more optimally

### DIFF
--- a/lib/Rex/Commands/Box.pm
+++ b/lib/Rex/Commands/Box.pm
@@ -413,13 +413,9 @@ sub boxes {
 
 }
 
-Rex::TaskList->create()->create_task(
-  "get_sys_info",
-  sub {
-    return { get_system_information() };
-  },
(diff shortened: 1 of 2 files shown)


ferki:
I think I'll give it a go to find a more sophisticated solution for this. The first attempt feels more like a hack.

ferki:
Okay, after taking a second look at the possibilities, I think we only can move around the second commit's detection logic to different parts, but I haven't found any reasonably better way to do it.

krimdomu:
It would be nice if we could have some automated Rex::Box tests. I have to test if we can install virtualbox on build.rexify.org without affecting kvm.

ferki:
@krimdomu: `Rex::Box` works with KVM ;)

krimdomu:
@ferki  yes but we should also test virtualbox  and amazon and possibly other providers for Rex::Box :)

PR #642, opened 2015-04-23
Allow cron operations without specifying the user

ferki:
If we don't require the user to be specified for cronjob operations, it would make it possible for users to manage their _own_ crontabs.


ferki:
On a second thought, it might not be _that_ simple as in the first comments.

krimdomu:
I think this looks good. I'm going to squash and merge it.

PR #650, opened 2015-04-29
perl -w turns on warnings for every dependency

kablamo:
perl -w turns on warnings for every dependency.  probably not what we want?

### DIFF
--- a/bin/rex
+++ b/bin/rex
@@ -1,4 +1,4 @@
-#!perl -w
+#!perl
 
 #
 # (c) Jan Gehring <jan.gehring@gmail.com>
--- a/bin/rexify
+++ b/bin/rexify
@@ -1,4 +1,4 @@
-#!perl -w
+#!perl
 
 #
 # (c) Jan Gehring <jan.gehring@gmail.com>


ferki:
@kablamo: nicely spotted, thanks! I agree with replacing `perl -w` in the shebang. Though I'm not sure that trying to execute `env` or `/usr/bin/env` on every supported platform wouldn't cause problems (like on Windows). So I would go for simply removing the ` -w` part (e.g. having `#!perl` in the first line). Can you replace your commit with that, if you agree?

Can I also kindly recommend using a commit message that describes the change, instead of the problem it tries to solve? I'm thinking of something like "Update shebang" in this case. For example I really like [this summary on the topic](http://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html).

kablamo:
I'm pretty sure /usr/bin/env is the right way to do this.   cpanm uses /usr/bin/env and that targets as wide of an audience as possible.  Windows ignores the shebang line.  It uses file extensions to determine how to run a file.

See also:
http://stackoverflow.com/questions/2791954/what-should-i-use-for-a-perl-scripts-shebang-line/2792076#2792076

I'll update my commit message

kablamo:
ready for review

krimdomu:
@kablamo the problem with #!/usr/bin/env is, that during the installation ExtUtils::MakeMaker doesn't rewrite this. (In some newer versions it do it, but i don't know how much these newer versions is distributed.) You can see the discussion about this here: https://github.com/Perl-Toolchain-Gang/ExtUtils-MakeMaker/issues/58

kablamo:
@krimdomu wow.  interesting.  unknown unknowns.  reverted to #!perl

krimdomu:
I'll also merge this. Thanks for this patch.

PR #651, opened 2015-04-29
Fixed yaml output for server groups

kablamo:
`rex -Ty` was dumping a Perl object Rex::Group::Entry::Server as yaml.  This patch makes it so we are dumping data and its a lot more readable.

### DIFF
--- a/lib/Rex/CLI.pm
+++ b/lib/Rex/CLI.pm
@@ -452,9 +452,9 @@ CHECK_OVERWRITE: {
 
     for my $group ( keys %groups ) {
       my @servers = 
-        map { $_->name }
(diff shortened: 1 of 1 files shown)


krimdomu:
Hi @kablamo ,

thanks for the patch. But Rex-JobControl depends on this output. So currently i can't merge this.
In one of the next releases i will rewrite the cli interface. I think then we can add this.

kablamo:
@krimdomu ok fair enough.  Now using -Tyy instead of messing with -Ty.

kablamo:
@krimdomu closing this one.  the yaml fix is now part of my tab completion pr.  i think thats simpler.

PR #652, opened 2015-04-29
Tab completion

kablamo:
ready for review

### DIFF
--- a/misc/rex-tab-completion.bash
+++ b/misc/rex-tab-completion.bash
@@ -1,0 +1,66 @@
+#!/bin/bash
+# bash completion for rex
+
+_rex()
+{
+    local cur prev
+
+    COMPREPLY=()
+    _get_comp_words_by_ref -n : cur prev
+
+    if [[ -z $_rex_yaml ]]; then
+        _rex_yaml=$(rex -Ty 2>/dev/null)
+    fi
+
+    case "$prev" in
+        -f)
+            _filedir
+            ;;
+        -H)
+            if [ -f Rexfile ]; then
+                hosts=( $(echo "$_rex_yaml" | perl -MYAML -MList::MoreUtils=uniq -E 'my $groups = Load(join "", <>)->{groups}; say $_->{name} for uniq sort map { @{ $groups->{$_} } } keys %$groups') )
+                COMPREPLY=( $( compgen -W '${hosts[@]}' -- "$cur" ) )
+            fi
+            ;;
+        -E)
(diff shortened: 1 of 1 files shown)


ferki:
@kablamo: many thanks for this contribution! As far as I can see, the first two commit messages can be squashed together, and can be described like `Add -Tyy CLI option`, right? You can do that and just (force) push again to your same branch, or alternatively we can do it for you manually before merging.

lib/Rex/CLI.pm
```diff
@@ -451,8 +451,10 @@ CHECK_OVERWRITE: {
     my %real_groups;
 
     for my $group ( keys %groups ) {
-      my @servers = map { $_->get_servers }
+      my @servers = 
+        map { $_->get_servers }
         Rex::Group->get_group_object($group)->get_servers;
+      @servers = map { $_->name } @servers if $opts{'y'} >= 2;
```
ferki: Can you add this new option (`-Tyy`) to the help output with a short description, please?

kablamo: I guess the reason I didn't do this is because I don't know how to describe it.  `-Ty` is `List all known tasks, in YAML format`.  But that description applies perfectly to `-Tyy` as well.  

Btw thats the main reason I'm not thrilled with this solution.  I would have preferred to add a key to the output in `-Ty`.

But this is fine too.  Do you have a suggestion for a description?

ferki: Perhaps it can be called `-Tc` instead and described as `List all objects for tab-completion`.

ferki:
@kablamo: great job, it works pretty well for me (with 93 tasks, 72 hosts in 12 groups and 1 environment) :+1: 

I believe it's very close to be merged and it would be nice to start shipping it as soon as possible ^^

Thinking a bit more longterm, what do you think about using modules like [`Complete`](https://metacpan.org/pod/Complete), and its submodules for Bash, Zsh, etc.?

kablamo:
I updated the tab completion thing to use the existing `-Ty` output.  That way I don't need to add an extra cmdline option.

@ferki Ready for review.

Complete looks interesting!  Haven't seen that before so I don't have much to say.  Complicated enough I think I would have to use it to have an opinion.

ferki:
LGTM :+1:

krimdomu:
@kablamo 
thank you for this patch. I'm going to merge it. This will be included in the next release.

kablamo:
sweet! :cake:

Issue #631, opened 2015-04-13
rex -T namespace


kablamo:
Lets say `rex -T` outputs the following
```
Tasks
 a:apple
 deploy:a
 deploy:b
 deploy:c
 git:pull
 grep:logs
 r:river
 z:zebra
```
We have a lot of tasks.  So it would be nice if `rex -T deploy` acted to filter everything but the deploy tasks so the output looks like this:
```
 deploy:a
 deploy:b
 deploy:c
```

kablamo:
Sorry I keep making tickets and not submitting pull requests.  Am hoping to circle around and solve at least some of these.

ferki:
@kablamo: no problem at all, we really appreciate feedback and especially opening issues ;) This is really a good idea and I thought it's already there as it works when a taskname is passed. But looks like it does not work in the general case since it breaks when only submodule names are passed.

Alternatively/additionally a `-Tf` switch can be introduced to perform a search for the passed argument in the list of all tasknames (e.g. to be able to find substrings in the middle of them).

krimdomu:
Yes, this is a really good idea. :+1:

kablamo:
feature added

PR #653, opened 2015-04-29
Prettier output

kablamo:
This is mostly a pr for https://github.com/RexOps/Rex/issues/631 -- filtering Task output for `rex -T` by namespace

Also there are a few other fixes for `rex -T` output in here.  See git commit msgs.

Ready for review.

### DIFF
--- a/lib/Rex/CLI.pm
+++ b/lib/Rex/CLI.pm
@@ -467,68 +467,7 @@ CHECK_OVERWRITE: {
     );
   }
   elsif ( $opts{'T'} ) {
-    Rex::Logger::debug("Listing Tasks and Batches");
-    _print_color( "Tasks\n", "yellow" );
-    my @tasks = Rex::TaskList->create()->get_tasks;
-    unless (@tasks) {
-      print "  no tasks defined.\n";
-      exit;
-    }
-    if ( defined $ARGV[0] ) {
-      @tasks = grep { $_ =~ /^$ARGV[0]/ } @tasks;
-    }
-
-    # Warn the user if they pass in arguments to '-T' and no task names
-    # were found that match those arguments
-    if ( scalar(@tasks) == 0 ) {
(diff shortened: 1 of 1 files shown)


lib/Rex/CLI.pm
```diff
@@ -517,8 +519,15 @@ CHECK_OVERWRITE: {
 
     my %groups = Rex::Group->get_groups;
     _print_color( "Server Groups\n", "yellow" ) if ( keys %groups );
+    my ($cols) = Term::ReadKey::GetTerminalSize(*STDOUT);
+    $Text::Wrap::columns = $cols // 80;
```
krimdomu: Hi @kablamo ,

the pr fails because the *//* operator is not available in perl 5.8. But for Rex 1.x we want to keep the perl 5.8 compatibility. So i think this line can also be written like this:

```perl
$Text::Wrap::columns = $cols || 80;
```

ferki: @kablamo: more specifically we need to maintain perl-5.8.9 compatibility until CentOS/RHEL 5 EOL. You can run test for that locally before committing/pushing with `dzil test --no-author --release`.

ferki:
@kablamo: it is very nice! If I'm not mistaken it is actually two logical changesets (taskname filtering and output formatting), but I guess we can merge them at once (or just cherry-pick the commits).

Do you think the same _prettification_ can be applied to the `-Tv` output as well?

kablamo:
@ferki Ready for review.  I got carried away and refactored all of `rex -T`

kablamo:
Ok added one last thing to make task output prettier. see commit msg. done messing around for real now.

krimdomu:
I think this looks also good. I'm going to squash the commits and merge it.

krimdomu:
Is merged and will be included in the next release. Thanks much!

PR #655, opened 2015-04-29
Rex/Commands.pm: update modules listed in POD

cpanxaoc:
Not sure if some of these modules have been deprecated, for example, Upload may be deprecated for the File module.


krimdomu:
Hi @cpanxaoc 

thanks for the documentation. I'm going to merge this :+1:

PR #662, opened 2015-05-03
Testsuite cleanup

ferki:
This is an ongoing work made available early for automated testing and code review.


ferki:
So, finally they pass all tests ^^ The main advantages of the above changes:
- make the local test suite ~35% faster
- automatically tries to `use` any new modules we add (so we don't forget to add them)

@krimdomu: please review and discuss, though before merge I would like to squash some of the commits together.

PR #672, opened 2015-05-07
Set parallelism automatically (fix #491)

### LINKED ISSUE #491: Automatic setting of parellelism
It would be nice if there would be an option with which Rex would automatically determine the number of hosts a task need to run on, and it would set parallelism automatically to that number. This would also possibly need a default, but overridable upper limit to prevent opening too much connections at once.

ferki:
@krimdomu: please review and discuss/merge.

### DIFF
--- a/lib/Rex/Fork/Task.pm
+++ b/lib/Rex/Fork/Task.pm
@@ -41,7 +41,7 @@ sub start {
 
     # only allow this if no parallelism is given.
     # with parallelism active it doesn't make sense.
-    if ( $Rex::WITH_EXIT_STATUS && Rex::Config->get_parallelism == 1 ) {
+    if ($Rex::WITH_EXIT_STATUS) {
       eval {
         &$func($self);
         1;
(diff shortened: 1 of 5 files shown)


lib/Rex/Fork/Task.pm
```diff
@@ -41,7 +41,7 @@ sub start {
 
     # only allow this if no parallelism is given.
     # with parallelism active it doesn't make sense.
-    if ( $Rex::WITH_EXIT_STATUS && Rex::Config->get_parallelism == 1 ) {
+    if ( $Rex::WITH_EXIT_STATUS && Rex::Config->get_parallelism eq '1' ) {
```
krimdomu: @ferki Why this change from numeric to string comparison?

krimdomu:
looks good :+1:

lib/Rex/Fork/Task.pm
ferki: @krimdomu: for me it was needed because the `-t` CLI argument was converted to string from the previous integer. As I was not 100% sure about that part, I'm totally up for better suggestions to handle it. If you could try to revert this single line and test locally, and can come up with a more suitable solution, I would greatly appreciate it!

PR #673, opened 2015-05-07
Add bios_boot option to GPT initialization

ferki:


krimdomu:
@ferki looks good :+1:

PR #688, opened 2015-05-15
Log STDERR on errors when auto_die is enabled

kopytov:
When executed command fails it usualy writes error message to STDERR. But rex's "run" doesn't put it to the log and you often do not undestand what goes wrong. To analyze the problem you have to run rex again with -d option and it is not useful if your task is not idempotent or executing for a long time.

This patch simply adds STDERR to a log message.


ferki:
@kopytov: thanks for this patch! And sorry for the late response.

PR #689, opened 2015-05-17
Iptables.pm: show error from iptables on non-zero exit status

cpanxaoc:
If there's something wrong with an IPTables rule you try to add to a host via the `iptables` DSL function, Rex should capture the output from `iptables` and display it, so that you can try and fix the offending `iptables` command(s).


krimdomu:
Hi  @cpanxaoc ,
thanks for this patch. I've just merged it into master branch and it will be part of the next release :)

PR #690, opened 2015-05-17
Iptables.pm: add long-form iptables examples to POD

cpanxaoc:
- Also added an explanation of how the Rex DSL 'iptables -F' or  'iptables "flush"' is handled by the _Iptables.pm_ module


krimdomu:
Hi @cpanxaoc 
thanks also for this patch. I've also merged it.

Issue #691, opened 2015-05-17
Display a summary of success/failure at the end of a run (like prove)


kablamo:
I propose the following:

- If a task dies or returns 0 the task has failed.
- Otherwise a task has succeeded.

The summary might look like this if everything was successful:
```
SUMMARY
deploy:product succeeded on all hosts.
Time: 25 seconds
```

The summary might look like this if there were failures:
```
SUMMARY
deploy:product failed on server1, server9, server23
Time: 25 seconds
```

ferki:
@kablamo: it's a good idea! I like that the summary is kept really short.

ferki:
One question though about the return values of the tasks: what is the reason behind treating 0 as failure? Is that a reliable way of detecting failure?

ferki:
It's already fixed on #697

PR #697, opened 2015-05-20
Summarize success/failure at the end of a run

kablamo:
# What this pr does

This pr logs a summary of successes/failures at the end of a run so the user doesn't have to scroll up through the logs to see if anything failed which can be a chore when deploying to many servers.

It also answers a fundamental question about rex: what defines success or failure of a task.  This pr defines a failed task as one that has died.  This is a simple solution for people who use the `exec_autodie` feature.  See [this comment](https://github.com/RexOps/Rex/issues/706#issuecomment-127365886) for more discussion on this design decision.

The summary looks like this if everything was successful:
```
SUMMARY
2015-08-04 08:16:19 <local> INFO : SUMMARY
2015-08-04 08:16:19 <local> INFO : All tasks successful on all hosts
```

The summary looks like this if there were failures:
```
SUMMARY
2015-07-31 17:04:37 <local> INFO : SUMMARY
2015-07-31 17:04:37 <local> INFO :  - deploy:test failed on eric
2015-07-31 17:04:37 <local> INFO :  - deploy:test failed on caine
2015-07-31 17:04:37 <local> INFO :  - deploy:test failed on jaryd
2015-07-31 17:04:37 <local> INFO : 3/3 task execution failures
```

See also: https://github.com/RexOps/Rex/issues/691

# Testing
 - [x] `t/summary.t` - Without the `exec_autodie` feature 
 - [x] `t/summary.t` - With the `exec_autodie` feature
 - [x] `t/summary.t` - With `Rex::TaskList::Base` and `Rex::Task::Manager`
 - [x] `t/summary.t` - With `Rex::TaskList::Parallel_ForkManager` and `Parallel::ForkManager`
 - [x] `t/summary.t` - With transactions
 - [x] `t/shared.t` - Parallel execution (handled via new tests added in https://github.com/RexOps/Rex/pull/774)

### DIFF
--- a/lib/Rex/Fork/Manager.pm
+++ b/lib/Rex/Fork/Manager.pm
@@ -27,18 +27,20 @@ sub new {
 }
 
 sub add {
-  my ( $self, $task, $start ) = @_;
-  my $f = Rex::Fork::Task->new( task => $task );
+  my ( $self, $coderef, $task, $server ) = @_;
+  my $f = Rex::Fork::Task->new(
+    coderef => $coderef,
+    task    => $task,
+    server  => $server,
+  );
 
   push( @{ $self->{'forks'} }, $f );
 
-  if ($start) {
-    $f->start;
-    ++$self->{'running'};
+  $f->start;
+  ++$self->{'running'};
 
-    if ( $self->{'running'} >= $self->{'max'} ) {
-      $self->wait_for_one;
-    }
+  if ( $self->{'running'} >= $self->{'max'} ) {
+    $self->wait_for_one;
   }
 }
 
--- a/lib/Rex/Fork/Task.pm
+++ b/lib/Rex/Fork/Task.pm
@@ -37,7 +37,7 @@ sub start {
     $self->{chld} = 1;
 
     eval { $self->{coderef}->($self) };
-    my $exit_code = $@ ? ($? || 1) : 0;
+    my $exit_code = $@ ? (($? >> 8) || 1) : 0;
     push @SUMMARY, {
       task      => $self->{task}->name,
       server    => $self->{server},
--- a/lib/Rex/Fork/Task.pm
+++ b/lib/Rex/Fork/Task.pm
@@ -13,10 +13,8 @@ use POSIX ":sys_wait_h";
 # VERSION
 
 BEGIN {
-
   use Rex::Shared::Var;
-  share qw(@PROCESS_LIST);
-
+  share qw(@SUMMARY);
 }
 
 sub new {
@@ -37,29 +35,17 @@ sub start {
   if ( $self->{pid} = fork ) { return $self->{pid}; }
   else {
     $self->{chld} = 1;
-    my $func = $self->{task};
 
-    # only allow this if no parallelism is given.
-    # with parallelism active it doesn't make sense.
-    if ($Rex::WITH_EXIT_STATUS) {
-      eval {
-        &$func($self);
-        1;
-      } or do {
-        push( @PROCESS_LIST, $? || 1 );
-        $self->{'running'} = 0;
-        die($@);
-      };
+    eval { $self->{coderef}->($self) };
+    my $exit_code = $@ ? ($? || 1) : 0;
+    push @SUMMARY, {
+      task      => $self->{task}->name,
+      server    => $self->{server},
+      exit_code => $exit_code,
+    };
 
-      $self->{'running'} = 0;
-      push( @PROCESS_LIST, 0 );
-      exit();
-    }
-    else {
-      &$func($self);
-      $self->{'running'} = 0;
-      exit();
-    }
+    $self->{'running'} = 0;
+    exit();
   }
 }
 
--- a/lib/Rex/Fork/Task.pm
+++ b/lib/Rex/Fork/Task.pm
@@ -37,12 +37,13 @@ sub start {
     $self->{chld} = 1;
 
     eval { $self->{coderef}->($self) };
-    my $exit_code = $@ ? (($? >> 8) || 1) : 0;
-    push @SUMMARY, {
+    my $exit_code = $@ ? ( ( $? >> 8 ) || 1 ) : 0;
+    push @SUMMARY,
+      {
       task      => $self->{task}->name,
       server    => $self->{server},
       exit_code => $exit_code,
-    };
+      };
 
     $self->{'running'} = 0;
     exit();
--- a/t/summary.t
+++ b/t/summary.t
@@ -1,0 +1,133 @@
+use strict;
+use warnings;
+
+use Test::More tests => 2;
+use Test::Deep;
+
+use Rex::Config;
+use Rex::Commands;
+use Rex::Commands::Run;
+use Rex::Transaction;
+
+$::QUIET = 1;
+
+subtest "distributor => 'Base'" => sub {
+
+  subtest 'exec_autodie => 0' => sub {
+    Rex::Config->set_exec_autodie(0);
+    Rex::Config->set_distributor('Base');
+    test_summary(
+      task0 => { server => '<local>', task => 'task0', exit_code => 1 },
+      task1 => { server => '<local>', task => 'task1', exit_code => 0 },
+      task2 => { server => '<local>', task => 'task2', exit_code => 0 },
+      task3 => { server => '<local>', task => 'task3', exit_code => 1 },
+    );
+  };
+
+  subtest 'exec_autodie => 1' => sub {
+    Rex::Config->set_exec_autodie(1);
+    Rex::Config->set_distributor('Base');
+    test_summary(
+      task0 => { server => '<local>', task => 'task0', exit_code => 1 },
+      task1 => { server => '<local>', task => 'task1', exit_code => 2 },
+      task2 => { server => '<local>', task => 'task2', exit_code => 0 },
+      task3 => { server => '<local>', task => 'task3', exit_code => 1 },
+    );
+  };
+};
+
+SKIP: {
+  skip "Parallel::ForkManager is not installed", 1
+    if parallel_forkmanager_not_installed();
+
+  subtest "distributor => 'Parallel_ForkManager'" => sub {
+    subtest 'exec_autodie => 0' => sub {
+      Rex::Config->set_exec_autodie(0);
+      Rex::Config->set_distributor('Parallel_ForkManager');
+      test_summary(
+        task0 => { server => '<local>', task => 'task0', exit_code => 1 },
+        task1 => { server => '<local>', task => 'task1', exit_code => 0 },
+        task2 => { server => '<local>', task => 'task2', exit_code => 0 },
+        task3 => { server => '<local>', task => 'task3', exit_code => 1 },
+      );
+    };
+
+    subtest 'exec_autodie => 1' => sub {
+      Rex::Config->set_exec_autodie(1);
+      Rex::Config->set_distributor('Parallel_ForkManager');
+      test_summary(
+        task0 => { server => '<local>', task => 'task0', exit_code => 1 },
+        task1 => { server => '<local>', task => 'task1', exit_code => 2 },
+        task2 => { server => '<local>', task => 'task2', exit_code => 0 },
+        task3 => { server => '<local>', task => 'task3', exit_code => 1 },
+      );
+    };
+  };
+}
+
+sub create_tasks {
+  desc "desc 0";
+  task "task0" => sub {
+    die "bork0";
+  };
+
+  desc "desc 1";
+  task "task1" => sub {
+    run "ls asdfxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx";
+  };
+
+  desc "desc 2";
+  task "task2" => sub {
+    my $cmd = "ls";
+    $cmd = "dir" if $^O =~ /MSWin32/;
+    run $cmd;
+  };
+
+  desc "desc 3";
+  task "task3" => sub {
+    transaction {
+      do_task qw/task0/;
+    };
+  };
+}
+
+sub test_summary {
+  my (%expected) = @_;
+
+  $Rex::TaskList::task_list = undef;
+
+  create_tasks();
+
+  my @summary;
+  my @expected_summary;
+  my $test_description;
+
+  for my $task_name ( Rex::TaskList->create->get_tasks ) {
+    Rex::TaskList->run($task_name);
+    @summary = Rex::TaskList->create->get_summary;
+
+    push @expected_summary, $expected{$task_name};
+
+    $test_description =
+      $expected{$task_name}->{exit_code} == 0
+      ? "$task_name succeeded"
+      : "$task_name failed";
+
+    cmp_deeply \@summary, \@expected_summary, $test_description;
+  }
+
+  my $distributor = Rex::Config->get_distributor;
+  no warnings;
+
+  @Rex::Fork::Task::SUMMARY = ()
+    if $distributor eq 'Base';
+
+  @Rex::TaskList::Parallel_ForkManager::SUMMARY = ()
+    if $distributor eq 'Parallel_ForkManager';
+}
+
+sub parallel_forkmanager_not_installed {
+  eval { require Parallel::ForkManager };
+  return 1 if $@;
+  return 0;
+}
--- a/lib/Rex/CLI.pm
+++ b/lib/Rex/CLI.pm
@@ -538,6 +538,7 @@ CHECK_OVERWRITE: {
   }
 
   Rex::Logger::info($_) for summarize();
+
   #print ">> $$\n";
   #print Dumper(\@exit_codes);
   # lock loeschen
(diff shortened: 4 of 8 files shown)


lib/Rex/Fork/Task.pm
```diff
@@ -56,7 +56,18 @@ sub start {
       exit();
     }
     else {
-      &$func($self);
+      eval { &$func($self) };
+      my $exit_code = $@ ? ($? || 1) : 0;
+      push @PROCESS_LIST, $exit_code;
```
kablamo: If tasks fail, the exit value of rex should reflect that.  That doesn't currently happen.  This fixes the problem.

lib/Rex/Fork/Manager.pm
```diff
@@ -27,18 +27,20 @@ sub new {
 }
 
 sub add {
-  my ( $self, $task, $start ) = @_;
```
kablamo: I acked but don't see that $start arg being used anywhere.  So I removed it.

lib/Rex/Fork/Task.pm
ferki: As a general rule I agree with exiting with a non-OK error code unless everything is OK. Do you think we can somehow add a test to check for this behaviour?

ferki:
@kablamo: looks like some of the code has changed since the commit and cannot be merged cleanly. Can you please rebase it on top of current master and force push again to this branch?

Also, do you plan to add the total execution time as proposed on #691?

Do you think it would make sense to put this into an `END` block so the summary gets printed even e.g. if `exec_autodie` is in effect?

lib/Rex/Fork/Task.pm
kablamo: Sorry, not sure I understand.  What do you want to the test to prove?

ferki: Like creating two tasks in the tests, one of them always succeeeding, and the other always failing. Then running each of them and checking if the exit code would be what we expect.

I haven't looked into this deeply yet, it's just an idea. But I guess if we're going to guarantee (and document) this behaviour, it would be very nice to also detect it somehow if we accidentally break this promise.

kablamo:
`exec_autodie` only causes the task to die.  It doesn't cause rex to die.  So I don't think its necessary to do the summary stuff in an `END` block.  

The only exception is if you have the `exit_status` feature enabled AND `parallelism` set to 1.  In that case, Rex dies immediately after the task fails.  Which I think is wrong.  I was tempted to fix it by removing the `if` block here: https://github.com/RexOps/Rex/pull/697/files#diff-5a0abb75a1efa761a64c39dad9a91626R44  But I didn't because I was worried about backwards compatibility.  

However I've changed my mind now --I think I should remove it.  Personally I think its a bug not a feature.

kablamo:
I don't plan to add total execution time.  There is already a profiling feature isn't there?  Anyway I'm just lazy to do that.   Its not important to me atm :) -- maybe in a future iteration.

ferki:
Execution time: a'ight, I just asked because you referenced #691, and there you proposed including total execution time as well, so I just couldn't tell whether you decided to leave it out completely, or if it is a work in progress and you will add it later before completing the work.

If execution time is important to someone, they can just use a `sayformat` including the timestamp and/or execute the task with `time rex mytask`. Plus if they are interested in the internal execution times for the different function calls, then profiling is there.

`END` block: maybe `exec_autodie` wasn't the best example I could come up with, but my point was something else anyway. I was thinking that if for _any reason_ the execution is aborted, it would still be nice to know what was successful and what was not. I think having thins within an `END` block we don't need to worry about _why_ the execution is stopped, yet still having the informative output (maybe except in some weird cases, when rex is aborted aggressively by some signal or something like that - but in that case the main concern is not about the summary anymore).

`exit_status`: well, the only info I could dig up about this featre flag is that it

>tells Rex to return a non zero value on exit if a task fails

so in that spirit I would keep the whole exit status handling completely inside an `if` block that checks if this feature flag is enabled or not (for backward compatibility). Otherwise I think everyone should enable this feature flag, so we might even make it default (and then provide a `no_exit_status` flag to disable it).

I think we can remove the parallelism check though. I mean as a user I'm interested in the proper exit status even if I run tasks parallelly even with the `exit_status` feature flag.

ferki:
Hmm..maybe we are trying to do too much in a single PR. If I undertand correctly one part of the change is about the exit status and the other part is about printing a summary. I'm not sure though if they can be splitted easily into two separate commits/PRs.

kablamo:
Rebased and did a forced push

kablamo:
@ferki I see what my confusion is.  

You updated that bit about the parallelism == 1 in another pr.  Now Rex dies immediately whenever a task fails.  Thats not good because if rex is doing stuff in parallel all the other tasks are going to exit in the middle and you could leave a server in some crazy middle state.  

Also I don't believe thats how `exit_status` was designed originally.  Note how the exit status is handled here: https://github.com/RexOps/Rex/blob/master/lib/Rex/CLI.pm#L624

Rather than me writing a bunch of words, I'll submit a fix and you can tell me if its crazy.

kablamo:
@ferki using an `END` block doesn't work because it gets called every time a process ends which is a lot when parallelism > 1.  I tried using a `DESTROY` block but that also gets called every time a process ends.  I guess because the process is forked and so everything is copied.  I was thinking of doing something tricky in the `DESTROY` block so it only prints the summary when the parent calls it, but the `%SUMMARY` shared variable appears to be empty when the parent calls it.  

I think that happens because all the other objects get destroyed first and Rex::CLI is the last to get destroyed so when its called there is nothing left in `%SUMMARY`.  I'm kind of guessing.  I'm sure there is a way to do it, but a solution is not coming to mind.  I kinda want to just leave it as is for now unless you have a clever solution.

krimdomu:
Hi @kablamo 

thank you much for this pr. I'm going to test this a some environments and try to merge this for 1.4 release then.

kablamo:
Arg.  So I've come to see the light with regards to not dying in tasks.  Unfortunately this summary thing doesn't work without `exec_autodie`.  I think I want to change it to use task return values instead of task exit status `$?`.

lib/Rex/Fork/Manager.pm
ferki: I think `$start` is being used in (Rex::TaskList::Base#L317)[https://github.com/RexOps/Rex/blob/master/lib/Rex/TaskList/Base.pm#L317]. Can you please add it back?

ferki:
Erm..I'm not sure if we mean the same thing by "task return value" and "task exit status", and what's the difference between them. Can you describe your thoughts about that?

Also I'm sorry, but due to my limited brain capacity I sometimes have a hard time understanding what happens actually in the various PRs and why :) I still think we're trying to deal with too much different things here in a single changeset. In order to eleminate the uncertainty about whether we talk about the same thing, do you think can we add tests first for this behaviour, please?

I'm thinking of one or more example situations (like all pass, all fail, partial success, `exec_autodie` on/off, `parallelism` 1/2, etc.), together with expected results. I think the best place for that would be in [RexOps/rex-build](https://github.com/RexOps/rex-build), but maybe it can be done in this repo's `t/` directory as well.

I really would love to see a functionality like this in one of the next releases. Making this behaviour dependent on a feature flag first can also be an option.

ferki:
Maybe it would also be a good idea to first work on ensuring that all tasks, no matter what, are returning some information about their failure or success. I feel that is the main underlying question here (and on severeal other related issues), and if we could sort that out first, this would actually be quite easy to implement afterwards.

[This comment](https://github.com/RexOps/Rex/issues/706#issuecomment-105981767) by @krimdomu is also slightly related. Maybe that data structure can be extended with a `success` key containing a `TRUE` or `FALSE` value depending on what the task thinks about its own success?

kablamo:
You are right.  There should be tests for this.

The **return value of the task** is pretty clear I think.  You get to the end of a task and `return 1` or `return 0`.

The **exit status of the task** is probably not as clear.  Each task is evaluated inside an `eval()`.  By exit status I meant, check whether the task died or not. 

Here is some psuedo code.  This is the algorithm I want to use to determine if a task was successful or not. 
```
my $success = eval { $task->() };
$success = 0 if $@;
say "task successful" if $success;
say "task failed" if !$success;
```

ferki:
Many thanks for the clarification! I agree that distinction makes sense.

Logging IRC stuff here: we're trying to focus on solving what seems to be an underlying problem for various issues, namely making all tasks to return at least a single success information as either `TRUE` or `FALSE` (+ controlling this behaviour with a feature flag for now). Most probably it will be implemented on a separate PR.

kablamo:
Ok sorry to do a reversal.  But I think returning success/failure from the task is not a great way to do this.  

I think a task should die on failure.  That way people can return data from their task using https://metacpan.org/pod/Parallel::ForkManager#RETRIEVING-DATASTRUCTURES-from-child-processes

kablamo:
@ferki We talked about doing task success/failure in a separate pr from the summary output.  But all the code in this pr is for dealing with task success/failure except the changes to CLI.pm.  Happy to talk on irc some more.

kablamo:
I updated the description to describe the latest changes and discussions.

kablamo:
FYI I did a forced push.  This now work with both Base and Parallel_ForkManager as the distributor.  Tests are included.

kablamo:
Looking at parallel execution.  I found a race condition exists in Rex::Shared::Var.  ...because it has a shared lock.  I don't get why its a shared lock.  I believe it should be an exclusive lock.  Hmm I think having the shared lock allows multiple instances of rex to run and use the same vars.db file.  Not sure why you would ever want that behavior.  I'm thinking `vars.db` should be renamed to `vars.$pid.db`.

kablamo:
@ferki This pr is ready for review again.  See the description for the testing I've done.  https://github.com/RexOps/Rex/pull/774 should get merged first though.

krimdomu:
Hi @kablamo ,

currently i'm testing this pull request and i have one problem when i'm using `do_task`. 
This happends only when `Rex::Transaction` is used. So i think this should be not that large problem.
I'm trying to track this down in the next days.

An example Rexfile:

```perl
use Rex -feature => ['1.2'];
use Data::Dumper;
use Rex::Transaction;

user "zzzz";
password 'yyyy';
pass_auth;

group servers => 'localhost', '127.0.0.1', '127.0.0.2';

task rollout => sub {
  transaction {
    do_task "prepare";
    do_task "deploy_app";
    do_task "test_app";
    do_task "switch_app";
  };
  
  say "done";
};

task prepare => group => servers => sub {
  say "prepare: " . connection->server;
};

task deploy_app => group => servers => sub {
  say "deploy_app: " . connection->server;
};

task test_app => group => servers => sub {
  say "test_app: " . connection->server;
  die if(connection->server eq "127.0.0.1");
};

task switch_app => group => servers => sub {
  say "switch_app: " . connection->server;
};

1;
```

The output is:
```
[2015-10-03 22:48:23] INFO - SUMMARY
[2015-10-03 22:48:23] INFO -  - rollout failed on <local>
[2015-10-03 22:48:23] INFO - 1/1 task execution failures
```

The expected output would be something like this:
```
[2015-10-03 22:48:23] INFO - SUMMARY
[2015-10-03 22:48:23] INFO -  - test_app failed on 127.0.0.1
[2015-10-03 22:48:23] INFO - 1/8 task execution failures
```

ferki:
@krimdomu: can it be somehow related to #774 (which is close to be merged IMHO)?

krimdomu:
@ferki no this is not related. just tested it with both merged, same result.

kablamo:
Hmm I did that intentionally.  It was the output I expected.  :smile_cat:   Isn't it weird to get 8 task execution failures when I only specified 1 task and 1 host on the command line?

It might be nice to have something like a stack trace though:

```
[2015-10-03 22:48:23] INFO - SUMMARY
[2015-10-03 22:48:23] INFO -  - rollout > test_app failed on 127.0.0.1
[2015-10-03 22:48:23] INFO - 1/1 task execution failures
```

But I'd rather save it for a later pr if you don't mind.  Doing it right adds complexity and I'm eager to get this pr merged.

krimdomu:
The stacktrace thing is a cool idea. And yes, we can do this with a later pr.

ferki:
Nice, prerequisite PR #774 is also almost ready to be merged :+1: I think here we only need rebasing, and some squashing before merge, right?

krimdomu:
for me, it can be merged

t/summary.t
```diff
@@ -0,0 +1,119 @@
+use Test::Most;
```
ferki: What is the reason to use `Test::Most`? I could only identify `cmp_deeply`, but for that `Test::Deep` would have been enough (or maybe even `is_deeply` without additional test dependencies). And I would keep exact test plans if possible :)

I'm not against using `Test::Most`, but for the sake of consistency, I'd prefer applying this to all (most?) of the other tests too then in a separate PR. Or just importing, what the tests actually needs.

ferki:
@krimdomu: ok, I'll handle merging these, thanks for the review!
@kablamo: let's merge #774 first, then this one, with some rebasing and squashing! :)

ferki:
@kablamo: hi, I believe this is ready to be merged, especially since #774 has already been merged as a prerequisite (hmm..I wonder if it would make sense to rebase on top of a PR branch in this case in the future).

It seems like there are some merge conflicts though :/ Can you help us out with resolving the conflict, then rebasing on top of current master + squashing commits into logical changesets, please?

kablamo:
Rebased

kablamo:
wut.  more conflicts?  fixing...

kablamo:
Rebased again.  Ready for merge.

ferki:
Thanks!

I'll try to squash the perltidy commit into the other ones. I see no point having cosmetic-only commits, when it's easily avoidable in the first place (e.g. setting up the editor do it by pressing a button, or automatically upon saving the file and/or with a local git commit hook).

ferki:
OK, it's quite close! Two test issues:
- `t/summary.t` fails if Parallel::ForkManager is not present - please make sure those tests only run if the module is present
- please use fixed test plans instead of simply calling `done_testing` - you may need to adjust the plan according to the above point, but see examples in `t/{cron,db,task}.t`

ferki:
`t/summary.t` also ssems to take up to 67% of total testing time (it went up to 18 seconds from 6 for me, summary.t taking 12 secs to finish). Are there any way we can make it faster? I know 18 seconds still sounds fast, but if possible I'd avoid tripling the time the test suite needs to run.

ferki:
Also, `die 'bork0'` and `die 'boop'`  tests seem redundant. `run ls` will fail on windows.

ferki:
Okay, I'm gonna give it a go with my usual tasks to give it some extendede testing, but unfortunately I don't think it can be merged before the tests are fixed.

t/summary.t
```diff
@@ -0,0 +1,131 @@
...
+      task0 => { server => '<local>', task => 'task0', exit_code => 1 },
+      task1 => { server => '<local>', task => 'task1', exit_code => 2 },
+      task2 => { server => '<local>', task => 'task2', exit_code => 0 },
+      task3 => { server => '<local>', task => 'task3', exit_code => 1 },
+      task4 => { server => '<local>', task => 'task4', exit_code => 1 },
+    );
+  };
+};
+
+done_testing();
```
ferki: Please use test plans if possible.

t/summary.t
```diff
@@ -0,0 +1,131 @@
...
+      task0 => { server => '<local>', task => 'task0', exit_code => 1 },
+      task1 => { server => '<local>', task => 'task1', exit_code => 2 },
+      task2 => { server => '<local>', task => 'task2', exit_code => 0 },
+      task3 => { server => '<local>', task => 'task3', exit_code => 1 },
+      task4 => { server => '<local>', task => 'task4', exit_code => 1 },
+    );
+  };
+};
+
+subtest "distributor => 'Parallel_ForkManager'" => sub {
```
ferki: These tests will die if `Parallel::ForkManager` is not present. Please run them only if the module is installed (and adjust the test plan accordingly.

t/summary.t
```diff
@@ -0,0 +1,131 @@
...
+  };
+
+  desc "desc 2";
+  task "task2" => sub {
+    run "ls";
+  };
+
+  desc "desc 3";
+  task "task3" => sub {
+    die "boop";
```
ferki: There are two `die` test tasks. Why?

t/summary.t
```diff
@@ -0,0 +1,131 @@
...
+  };
+
+  desc "desc 1";
+  task "task1" => sub {
+    run "ls asdfxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx";
+  };
+
+  desc "desc 2";
+  task "task2" => sub {
+    run "ls";
```
ferki: This is going to fail on windows.

ferki:
Is it supposed to work with `rex -G mygroup -e 'say run "uptime"'` too?

kablamo:
Fixed tests and it now works with rex -e.  Ready for merge.  (I hope).

ferki:
Thanks @kablamo, great job! And thanks for your patience too!

kablamo:
:balloon: \o/ :balloon:

ferki:
Sorry for commenting long after closing, but this PR is interrelated with many issues which are still open. I'd like to call for help triaging all related issues again. Thanks.

kablamo:
@ferki I've been wanting to get back into this.  I'm on vacation finally and flying out to singapore on monday for a couple weeks.  One of my goals for vacation is to spend lots of time on Rex.  But its likely you won't see me until this coming Wednesday.

PR #698, opened 2015-05-21
Usage updates

kablamo:
1. I tweaked the descriptions slightly
2. I reordered the descriptions a bit
3. I added some empty lines to make the different groups of options stand out a bit.  It was a bit hard to parse because there are so many options
4. I added some usage statements at the top

I know this is super subjective and readability is in the eye of the beholder.  But I think this output is a little better.  What do you think?

```
ubuntu@eric:~/code/rex ⚡ rex -h
usage: 
  rex [<options>] [-H <host>] [-G <group>] <task> [<task-options>]
  rex -T[m|y|v] [<string>]

  -b     Run batch
  -e     Run the given code fragment
  -E     Execute a task on the given environment
  -G|-g  Execute a task on the given server groups
  -H     Execute a task on the given hosts (space delimited)
  -z     Execute a task on hosts from this command's output

  -K     Public key file for the ssh connection
  -P     Private key file for the ssh connection
  -p     Password for the ssh connection
  -u     Username for the ssh connection

  -d     Show debug output
  -dd    Show more debug output (includes profiling output)
  -m     Monochrome output: no colors
  -o     Output format
  -q     Quiet mode: no log output
  -qw    Quiet mode: only output warnings and errors
  -Q     Really quiet: output nothing

  -T     List tasks
  -Tm    List tasks in machine-readable format
  -Tv    List tasks verbosely
  -Ty    List tasks in YAML format

  -c     Turn cache ON
  -C     Turn cache OFF
  -f     Use this file instead of Rexfile
  -F     Force: disregard lock file
  -h     Display this help message
  -M     Load this module instead of Rexfile
  -s     Use sudo for every command
  -S     Password for sudo
  -t     Number of threads to use (aka 'parallelism' param)
  -v     Display (R)?ex version


```


ferki:
hi @kablamo: looks good, thanks! :+1: I'm going to squash the two commits together before merging.

krimdomu:
really nice! :+1:

ferki:
@kablamo: this is now merged manually. I squashed your commits into one, applied perltidy (and hopefully fixed jenkins tests to detect this case) and modified `-dd` to `-ddd` as that's the one that actually enabled profiling output. I also went through the recognized list of options to make sure everything is in the help message (and I found `-O` is recognized but not used, so I'll remove that to avoid future confusion).

kablamo:
@ferki awesome!

PR #699, opened 2015-05-21
Add support for CloudLinux (RedHat clone)

kopytov:
This adds support for CloudLinux (RedHat clone distro for shared hosting providers: cloudlinux.com).

    $ lsb_release -s -i
    CloudLinuxServer


ferki:
@kopytov: thanks for this patch as well! I merged it manually after rewording the commit message a bit.

PR #700, opened 2015-05-21
Low priority for REX_USER environment variable

kopytov:
When REX_USER is set globally I get a lot of errors on make test:

    $ export REX_USER=blabla
    $ make test
    PERL_DL_NONLAZY=1 "/opt/perl/5.20/bin/perl" "-MExtUtils::Command::MM" "-MTest::Harness" "-e" "undef *Test::Harness::Switches; test_harness(0, 'blib/lib', 'blib/arch')" t/*.t t/commands/*.t
    
    #   Failed test 'merge_auth - user'
    #   at t/0.31.t line 53.
    #          got: 'blabla'
    #     expected: 'root3'

    #   Failed test 'merge_auth - user'
    #   at t/0.31.t line 64.
    #          got: 'blabla'
    #     expected: 'root3'
    
    ...

This patch reduce priority for REX_USER environment variable. If you set user in you Rexfile it will be user instead of REX_USER and tests will works properly.

Please review.

### DIFF
--- a/lib/Rex/Config.pm
+++ b/lib/Rex/Config.pm
@@ -435,12 +435,12 @@ sub has_user {
 sub get_user {
   my $class = shift;
 
-  if ( exists $ENV{REX_USER} ) {
-    return $ENV{REX_USER};
-  }
-
   if ($user) {
     return $user;
+  }
+
+  if ( exists $ENV{REX_USER} ) {
+    return $ENV{REX_USER};
   }
 
   return getlogin || getpwuid($<) || "Kilroy";


krimdomu:
Hi,
thank you for this patch. The thing with the REX_USER environment variable is to overwrite the user that is set in the Rexfile. So i think it would be better to fix the tests, so that they set the environment to an expected state before they run.

ferki:
Or even better, we can test for the correctness of this behaviour :)

E.g. would it be used correctly if there's a `REX_USER` environment variable present? If not present, let's emulate as if it were present and still run the same test. Plus run the remaining auth tests as if there were no `REX_USER` env variable set.

krimdomu:
@ferki :+1:

Issue #685, opened 2015-05-12
Use =head2 POD directive for Rex class methods and exported DSL functions


cpanxaoc:
The POD style for Rex class methods and subclass exported DSL functions is to currently use `=over/=back` blocks with `=item` directives for individual methods or exported DSL functions.  This works, however, if you use `=head2` directives instead of `=item` directives for these, they'll be listed in Contents section for search.cpan.org/MetaCPAN, as well as being more strongly highlighted when viewing POD in a browser.  The `=over/=back/=item` blocks should still be used for method/function arguments, but it would be nice if, when you viewed a Rex class online, you immediately see what DSL functions are available in that class in the "Contents" box of SCO/MetaCPAN.

ferki:
@cpanxaoc: that sounds great! We actually wanted to achieve the same with the API docs on rexify.org. I guess we can use the same `=head2` directives to drive content generation in `misc/create_pod.sh` as well (which is a bit cryptic script and only generates sensible output with perl-5.14.4, so it could definitely use a revamp).

ferki:
@cpanxaoc: thanks again for this contribution! I merged it manually after squashing your commits together and slightly rewording the commit message.

PR #701, opened 2015-05-22
Issue #685; convert POD =item to =head2 for methods/DSL functions

cpanxaoc:
- For all instances of (R)?ex class methods and exported DSL functions, convert the POD of those items to use the `=head2` directive, not the `=item` directive, so the methods/functions are show in the POD's index (at the top of the page)

For reference, the script I used is here:

https://github.com/cpanxaoc/rex-misc/blob/master/convert_item_to_head2.pl

And output from the script is here:

https://gist.github.com/cpanxaoc/24d94a0bb1df2e525555

### DIFF
--- a/lib/Rex.pm
+++ b/lib/Rex.pm
@@ -49,8 +49,6 @@ See L<Rex::Commands> for a list of all commands you can use.
 
 =head1 CLASS METHODS
 
-=over 4
-
 =cut
 
 package Rex;
@@ -228,7 +226,7 @@ sub modified_caller {
   }
 }
 
-=item get_current_connection
+=head2 get_current_connection
 
 This function is deprecated since 0.28! See Rex::Commands::connection.
 
@@ -272,7 +270,7 @@ sub get_current_connection_object {
   return Rex::get_current_connection()->{conn};
 }
 
-=item is_ssh
+=head2 is_ssh
 
 Returns 1 if the current connection is a ssh connection. 0 if not.
 
@@ -289,7 +287,7 @@ sub is_ssh {
   return 0;
 }
 
-=item is_local
+=head2 is_local
 
 Returns 1 if the current connection is local. Otherwise 0.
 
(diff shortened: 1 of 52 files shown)


ferki:
@cpanxaoc: thanks, looks good at first glance! Before merging, we need to find out why the tests failed at first try and how it affects the current HTML generation for the website.

I found these issues so far:
- after removing an `=over 4` or `=back` line, there's a double empty line remaining: it should either be cleaned up by a subsequent perltidy run (using the `.perltidyrc` profile in the project root), or the script can remove the superfluous empty lines it would leave after the conversion
- I'm not sure why, but tests seem to fail with messages like this:
```
$ prove t/0.31.t
t/0.31.t .. 1/147 
#   Failed test 'use Rex;'
#   at t/0.31.t line 7.
#     Tried to use 'Rex'.
#     Error:  Rex/Config.pm did not return a true value at /home/ferki/projects/rex/lib/Rex.pm line 73.
# BEGIN failed--compilation aborted at /home/ferki/projects/rex/lib/Rex.pm line 73.
# Compilation failed in require at t/0.31.t line 7.
# BEGIN failed--compilation aborted at t/0.31.t line 7.
```

cpanxaoc:
I see the problem, I'm working on fixing it.

I'll make sure to run the tests next time ;)

cpanxaoc:
Tests successfully run via 'dzil test --no-author', after I added Test::UseAllModules

The problem was removing =back directives without removing associated =cut directives, it basically left a bunch of =cut directives by themselves, which I don't think Perl liked.

cpanxaoc:
Closing this pull request.  I backed out my original commit, rebased my tree from origin/master, then made a new commit with fixes for the issues that Ferki found, but GitHub hasn't updated the original commit, so the tests still fail, as it's using the old commit.  GitHub is telling me I should have just made new commits in my working branch and pushed them to update this pull request, which is not what I did.

ferki:
Even forced push didn't work? Normally that's how we update a feautre branch when the history changed.

cpanxaoc:
I force pushed, but the pull request didn't update for whatever reason, it sent the original PR commit to Jenkins.

ferki:
Continued on #705.

PR #704, opened 2015-05-25
Fix auth test when REX_USER env is present

kopytov:
Guys, please review.

This patch adds new tests which checks if REX_USER environment variable correctly overwrites user from Rexfile. And fixes other auth test by deleteing REX_USER environment variable before user being set.

### DIFF
--- a/t/0.31.t
+++ b/t/0.31.t
@@ -5,6 +5,8 @@ use Test::More tests => 139;
 
 use Rex -feature => '0.31';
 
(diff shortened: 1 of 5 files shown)


ferki:
@kopytov: thanks you for your follow-up patch!  Looks good at first glance!

ferki:
@kopytov: this is now merged, thank you! I also added you to the list of contributors.

PR #713, opened 2015-06-06
feature for treating hostnames starting with sequential numbers

okaoka:
prefix "^" enables you to treat hostnames starting with sequential numbers like the following .
```
[cassandra]
^[01..02]-cassandra
```


ferki:
@okaoka: nice, thanks for this patch!

Do you think we can use a slightly different method to recognize the leading numbers (and avoid using `^` as a special marker)? I'm thinking about something like:
- if the entry matches `m{^\[\d+(?:\.\.|,)\d+(?:/\d+)\]}` treat it as host

okaoka:
@ferki
I tried not to use special marker .

ferki:
@okaoka: If I understand correctly it would recognize entries like `[01..02]-cassandra` ("one or more digits, followed by optional `..`, followed by one or more digits, all within square brackets").

I think rex should also handle the following cases (those are supported by e.g. the `-H` CLI option, and recognized as host specification):
- `[01,02]-cassandra` (comma separated)
- `[01..05/2]-cassandra` (range operator with a stepping)

ps.: the full [list of supported expressions](http://www.rexify.org/api/Rex/Commands.pm.html#group-name-servers-) can be found in the docs, but the above should be all of them currently

okaoka:
@ferki
Sorry for my oversight.
I fixed regular expression for treating missing expressions.

ferki:
@okaoka: no problem, thanks for the update! And sorry for the bit late feedback too.. I expect to be able to review/merge it soon for the next release.

ferki:
@okaoka: this is now merged, thank you! I squashed your commits together before merging, added some documentation updates about the chenges, and I also added you to the list of contributors.

PR #717, opened 2015-06-11
Add function for checkout of Git tags

eskaaren:
This patch makes it possible to use Git tags with Rex.
In the Rexfile specify tag => "tagname" instead of branch.
This is useful in environments where tags are used a lot like
CI/build systems.


eskaaren:
I will re-upload this pull request from a branch instead of master.

PR #718, opened 2015-06-11
Add function for checkout of Git tags

eskaaren:
This patch makes it possible to use Git tags with Rex.
In the Rexfile specify tag => "tagname" instead of branch.
This is useful in environments where tags are used a lot like
CI/build systems.


krimdomu:
Hi @eskaaren 
thanks for this patch. I'm going to merge this now into master branch.
I also ran perltidy on the code, so don't be suprised that the syntax changed a bit :)

PR #720, opened 2015-06-13
Fix LEFT_PRECEDENCE typo

kopytov:
Just fix typo in pod.


ferki:
@kopytov: good catch, thanks!

Issue #694, opened 2015-05-19
When run_task is called on a remote server, adding a before_task_start("ALL") hook makes it run locally


mbroadhead:
This code works as expected:
```perl
use Rex -base;
use feature qw(say);

user "random";

task "local_task_that_calls_another_task_remotely" => sub {
  run_task "foo", on => "example.net";
};

task "foo" => sub {
  say "foo runs on " . connection->server;
};

1;
```
The above code outputs:
```
    $ rex local_task_that_calls_another_task_remotely
    [2015-05-19 13:53:44] INFO - Running task local_task_that_calls_another_task_remotely on <local>
    [2015-05-19 13:53:44] INFO - Connecting to example.net:22 (random)
    foo runs on example.net
```

Now, if you add a before_task_start("ALL"), the "foo" task will run locally. This is not what I expect to happen. I expect the "foo" task to still run on example.net:
```perl
use Rex -base;
use feature qw(say);

user "random";

task "local_task_that_calls_another_task_remotely" => sub {
  run_task "foo", on => "example.net";
};

task "foo" => sub {
  say "foo runs on " . connection->server;
};

before_task_start "ALL" => sub {};

1;
```
The above code outputs:
```
    $ rex local_task_that_calls_another_task_remotely
    [2015-05-19 13:59:01] INFO - Running task local_task_that_calls_another_task_remotely on <local>
    foo runs on <local>
```

I also tried `before_task_start "ALL" => sub { return @_; };` but that didn't work either.

Am I using the before_task_start("ALL") hook incorrectly or is this a bug?

mbroadhead:
This is also the case when using `before_task_start("foo") => sub {};` as well.

If, however, I only add  `before_task_start("local_task_that_calls_another_task_remotely") => sub {};`, the foo tasks runs on the correct host.

krimdomu:
Hi @mbroadhead 
thanks for the report. We're going to look at it.

petersonchen:
Hi @krimdomu 

I got the same problem after using before_task_start call 
is there any workaround / patch to fix it ? 
My Rex version is 1.2.1

mbroadhead:
On a related note, using run_task()  within a task that has a remote host connection causes it to attempt to connect via ssh, even if it is suppose to be local. Example:
```perl
use Rex -base;
use feature qw(say);

task "call_run_task_foo_locally" => sub {
    run_task "foo";
};

task "foo" => sub {
    say "foo runs on " . connection->server;
};

[...]

PR #727, opened 2015-06-19
Don't recreate connection when executing rethink_connection.

mbroadhead:
We need to create a task object's connection as late as possible, so
that everything is setup that it relies on. When modifying one of a tasks
hooks, it would run rethink_connection which in turn would re-create a
local connection before it knew it was supposed to be an ssh connection.

This should fix issue #694

### DIFF
--- a/lib/Rex/Task.pm
+++ b/lib/Rex/Task.pm
@@ -385,7 +385,6 @@ sub modify {
 sub rethink_connection {
   my ($self) = @_;
   delete $self->{connection};
-  $self->connection;
 }
 
 =head2 user


mbroadhead:
@krimdomu @ferki Have you had a chance to review issue #694 and related pull request #727 ? Let me know if you need anymore information.

krimdomu:
@mbroadhead sorry i hadn't hat the possibility to look at it yet because i can't really type at the moment (handwrist problems)
I'll take a look at it as soon as possible.

mbroadhead:
@krimdomu That's okay. I hope you get better soon.

ferki:
@mbroadhead: thanks for this patch! So far it looks good to me, also after running a VM test as well (on Debian 7, chosen randomly from our pool).

On #694, you mentioned some nice examples about how to reproduce it. Can you maybe add a testcase for [tests/hooks.rex](https://github.com/RexOps/rex-build/blob/master/tests/hooks.rex) in the rex-build repo so we can make sure we're covered?

ferki:
@mbroadhead: I slightly reworded the commit message, but this is merged now. Thanks for the fix!

krimdomu:
@ferki 
We need to write some tests for this before releasing 1.4.0
because of: #789

PR #735, opened 2015-06-25
Skip can_run test on win32 systems that don't implement where

jluis:


ferki:
@jluis: many thanks for this patch!

I was thinking about something like this for quite some time now, but my fear is that if we don't fail the testsuite on those systems which are known to be not supported (like WinXP, or basically every Windows release before Vista), then it will become a runtime failure, which is a bigger problem.

I think it would be cleaner to fail the testsuite with the proper message (like `Unsupported OS detected: WinXP`). People can still decide to install Rex from CPAN without running the tests despite the failure, and doing so implies agreeing to the associated risks because they chose to ignore the failures.

The other option seems to be implementing proper `which`/`where` logic for all Windows systems before Vista. Which might be totally possible, but I wonder if it would worth the effort to create and then maintain that piece of code, which is needed solely to be compatible with OSes that are unsupported by their own upstream.

Those are just my thoughts about the problem. Despite I'm not a Windows user anymore, I'm absolutely pro-Windows in terms of Rex support, and I would totally like to see a nice solution for this dilemma. I just couldn't resolve it on my own so far :)

jluis:
@ferki I did the patch earlier this June for CPAN pull request chalenge and also toyed with the idea of using a  where.cmd/where.bat for the distro in fact I did it first using where.cmd that pases the tests:
```cmd
@echo off
@SET FULLNAME=
@for %%e in (%PATHEXT%) do @for %%i in (%1%%e) do @if NOT "%%~$PATH:i"=="" set FULLNAME=%%~$PATH:i
@if "(%FULLNAME%)" =="()" exit /B -3
@echo %FULLNAME%
```
But I never have used Rex and didn't found time this month to look at the source and try it.

If you dont mind it I will redo the patch so that it emits the warning  and skips the rest of can_run test if it cannot find a valid where ( it will fail the first test and skip the rest)

ferki:
@jluis: I was thinking about it a lot, but since `can_run` is not just part of the public API, but also used internally, I still think the `can_run` tests should _fail_ if there's a problem (vs. _skipping, but installing_, even with a warning message). I believe accepting this risk, and overriding the testsuite should be a conscious choice of the user, and I think printing a warning would not be enough.

Using your code-snippet might be a workaround for per-Vista Windows systems, but again, *IMHO* there's not much point putting lots of effort into supporting abandoned OSes. Personally I would go for detecting this case during build time (vs. test time), with `Dist::Zilla`, and fail as early as possible (fixing e.g. false FAIL reports from CPAN Testers as well).

@krimdomu: what are your thoughts on this topic?

ps.: @jluis: if you're looking for other candidates for the CPAN PR challenge, there are some issues marked with the label `easy-to-fix` :) Your efforts in the challenge are very well appreciated and I'd be happy to help your PR merged. Keep up the good work ;)

jluis:
@ferki 
I modified the pull request with a new test for suported windows versions to fail earlier  on those  systems.
When the checker test fails the rest of the test are skipped.

ferki:
@jluis: based on the information posted on #751, I'm going to close this PR unmerged for now. Let me know if you'd like to send a PR with the logic proposed on the referenced issue. Otherwise I'm going to implement it latest before the next release.

I'm also making an explicit reference here to your [previous comment containing a code snippet that might be used for emulating `which`/`where` functionality on Windows platforms that do not have those tools](https://github.com/RexOps/Rex/pull/735#issuecomment-115607636).

PR #736, opened 2015-06-28
Added key_file method to pkg/Debian and added to pod for Pkg about keys

brainbuz:
The Pkg :: repository documentation made no mention of how to import the signing keys, I had to read the Debian submodule to figure it out. I also had to add a key from a file and wget doesn't accept the file:// resource type, so I added a method for file (the other option I considered was to switch from wget to curl).


ferki:
@brainbuz: thanks for this patch (and sorry for the late response). It looks good, and I'm going to merge it for the next release. Extra kudos for the documentation bits ;)

ps.: I just noticed you used your clone's master branch to implement and send this PR - depending on your workflow you'd like follow, it might be easier to keep those changes in a separate branch next time.

ferki:
@brainbuz: I merged it after cleaning up a trailing space, and slightly rewording the commit message. I also added you to the list of contributors :) Thanks again!

Issue #712, opened 2015-06-04
pass filename of `file` command through `glob` function


ferki:
Contructs like this:
```perl
file '/tmp/test1', ensure => 'directory';
file '/tmp/test1', ensure => 'directory';
```
can be simplifed like this:
```perl
file [ qw( /tmp/test1 /tmp/test2 ) ], ensure => 'directory';
```
by relying on the fact that `file` command accepts an array reference as its first argument. Following up on that idea, it can be further simplified to:
```perl
file [ glob('/tmp/test{1,2}') ], ensure => 'directory';
```

I think it would be nice to run the filenames passed via `file` command's first argument through `glob` (regardless if they are given as simple scalars or multiple ones in an array reference), making it possible (and maybe more natural) to write constructs like this:
```perl
file '/tmp/test{1,2}', ensure => 'directory';
```

jluis:
@ferki on https://github.com/jluis/Rex/commit/9b5f94db605581c9bb6d8cb1cc42d59dfaf3a645 I have put test and code to resolve this isue but I am not surte that the filename is santized enohg for glob to proccess.

I don't check for the existence of chars cound make glob stat for the filename ony if there are braces to pas the braces and the text inside to glob.

ferki:
After carefully considering different use-cases and possibilities with @jluis, we decided not to implement this as a feature in Rex. Instead, we're going to document the multiple file creation options, including examples for glob usage, and also warnings about possible pitfalls.

PR #738, opened 2015-07-03
Allow 'file' to process glob {a,,c,--} expressions on filenames  #712

jluis:
Only allows glob expressions that are not affected by the file system.
{marcros} are preserved escaping the braces before the glob call


lib/Rex/Commands/File.pm
```diff
@@ -290,6 +290,21 @@ This function is the successor of I<install file>. Please use this function to u
...
+
+  file '/tmp/test1', ensure => 'directory';
+  file '/tmp/test2', ensure => 'directory';
+
+  file [ qw( /tmp/test1 /tmp/test2 ) ], ensure => 'directory'; #use array ref
+
+  file [ glob('/tmp/test{1,2}') ], ensure => 'directory'; #explicit glob call for array contents
+
+  file '/tmp/test{1,2}', ensure => 'directory';
+
```
ferki: Love the documentation bits :+1: Even though some typos made it through :)

lib/Rex/Commands/File.pm
```diff
@@ -357,6 +372,19 @@ sub file {
...
+      my @ret;
+      # quote braces to keep macros
+      $exp[1] =~ s/({[^,}]*)}/\\$1\\}/g;
+  
+      for my $f ( glob $exp[1] ) {
+          push ( @ret, file( "$exp[0]$f$exp[2]", @options) );
+      }
+
+      return \@ret;
+    }
```
ferki: I have a hard time understanding what this code branch does, especially the regex in the condition. Can you explain it, please? I probably miss a corner case you found, because I expected the solution would be simpler, something like:
```perl
my @files_after_glob = glob($file);

if ( scalar @files_after_glob > 1 ) {
    Rex::Logger::debug( "Files after glob: " . join ', ', @files_after_glob);
    my $ret = file( \@files_after_glob, @options );
    return $ret;
}
```

t/file.t
```diff
@@ -23,20 +23,21 @@ if ( $^O =~ m/^MSWin/ ) {
   $tmp_dir = $ENV{TMP};
 }
 
-my $filename = "$cwd/test-$$.txt";
+my $filename = "$cwd/test-{a,b}-$$.txt";
```
ferki: Kudos for the testing! :+1: Would it be enough to test only the newly added "glob" case?

lib/Rex/Commands/File.pm
jluis: El Dimarts 18 Agost 2015, a les 13:19:03, Ferenc Erki va escriure:
> > @@ -357,6 +372,19 @@ sub file {
> >      }
> >  
> >      return \@ret;
> > +  } 
> > +  elsif ( my @exp = $file =~ m/^((?:\\.|[^\\\{])*)({(?:\\.|[^\\\*?\[])*})((?:\\.|[^\\\}])*)$/ ) {
> > +    if ( $exp[1] =~ /{[^},]*,[^}]*}/ ) {
> > +      my @ret;
> > +      # quote braces to keep macros
> > +      $exp[1] =~ s/({[^,}]*)}/\\$1\\}/g;
> > +  
> > +      for my $f ( glob $exp[1] ) {
> > +          push ( @ret, file( "$exp[0]$f$exp[2]", @options) );
> > +      }
> > +
> > +      return \@ret;
> > +    }
> 
> I have a hard time understanding what this code branch does, especially the regex in the condition. Can you explain it, please? I probably miss a corner case you found, because I expected the solution would be simpler, something like:
> ```perl
> my @files_after_glob = glob($file);
> 
> if ( scalar @files_after_glob > 1 ) {
>     Rex::Logger::debug( "Files after glob: " . join ', ', @files_after_glob);
>     my $ret = file( \@files_after_glob, @options );
>     return $ret;
> }
> ```

Sorry for the late answer. I was on holidays and my time on the computer was for the upcoming YAPC::Europe on Granada and its view on Lanyrd <G>

My assumption is that the machine that runs the glob is not the one with the path to be tested. The first regex checks that the expression that we pass to glob don't have wildcards the second is to be sure that glob will return more than one path

t/file.t
jluis: Yes when testing for the glob case we indirectly also test the array reference and the single string  but using it for the rest of the test would double the time <G>

lib/Rex/Commands/File.pm
ferki: No problem at all :) I'm looking forward to YAPC::EU as well ^^

I think I get it now what this branch tries to do: if the passed filename looks like a glob expression, it only wans to process the ones that has curly braces in it (because it doesn't really a good idea to process `/tmp/*` and run on a list of filenames that is there at the moment).

Maybe we can rephrase the feature's idea: support filenames expressions with curly braces in it, like if they were processed by blob.

In that case, it might be enough to detect the case "if the filename has curly braces that looks like a glob expression", and then "generate a list of files like glob would do". In this case all the filenames that is supposed to have a literal curly brace in them, needs an escaped curly brace.

I need to think about that a bit more.

t/file.t
ferki: That's true, but what if the simple filename handling and/or the array reference method breaks in the future, and the test suite only complains about a globbed form? I'd prefer making sure the test suite shows exactly what is broken to avoid extra debugging. Creating explicit test cases for the 3 different forms also has a second purpose as a form of documentation of intended/supported use cases.

jluis:
@ferki I think that the pull request is now complete we can comment is anything more is necessary on this evening Barcelona.pm meeting

ferki:
@jluis: I squashed together the commits here and slightly modified the wording, but this is now merged. Thank you!

PR #739, opened 2015-07-06
A way for adding several ssh keys, also from file.

eskaaren:
This patch makes it possible to add several keys to the users
authorized_keys and also makes it possible to either append or
overwrite.

This is very useful when several people are using Rex for handling
the same systems.

An example of how it could be used:

task "my_user", sub {

  my $user = getlogin || getpwuid($<);
  my @key_files = (
    "/root/.ssh/id_rsa.pub",
    "/home/$user/.ssh/id_rsa.pub",
  );

  my @ssh_keys;
  for my $file (@key_files) {
    open my $fh, "<", "$file";
    while (my $line = <$fh>) {
      chomp $line;
      push(@ssh_keys, "$line");
    }
  }

  create_user "server_user",
              home            => '/home/server_user',
              comment         => 'Demo user',
              expire          => '2100-05-30',
              password        => 'password',
              system          => 1,
              create_home     => TRUE,
              ssh_keys_append => TRUE,
              ssh_keys        => \@ssh_keys;
};

### DIFF
--- a/lib/Rex/Commands/User.pm
+++ b/lib/Rex/Commands/User.pm
@@ -15,18 +15,20 @@ With this module you can manage user and groups.
 =head1 SYNOPSIS
 
  use Rex::Commands::User;
- 
+
  task "create-user", "remoteserver", sub {
    create_user "root",
-     uid         => 0,
-     home        => '/root',
-     comment     => 'Root Account',
-     expire      => '2011-05-30',
-     groups      => [ 'root', '...' ],
-     password    => 'blahblah',
-     system      => 1,
-     create_home => TRUE,
-     ssh_key     => "ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAABAQChUw...";
+     uid              => 0,
+     home             => '/root',
+     comment          => 'Root Account',
+     expire           => '2011-05-30',
+     groups           => [ 'root', '...' ],
+     password         => 'blahblah',
+     system           => 1,
+     create_home      => TRUE,
+     ssh_keys_append  => TRUE, # overwrite is default
+     ssh_key          => "ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAABAQChUw...",
+     ssh_keys         => [ "ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAABAQChUw...", "..." ];
  };
 
 =head1 EXPORTED FUNCTIONS
@@ -52,8 +54,8 @@ use vars qw(@EXPORT);
 use base qw(Rex::Exporter);
 
 @EXPORT = qw(create_user delete_user get_uid get_user user_list
-  user_groups create_group delete_group get_group get_gid
-  account lock_password unlock_password
+user_groups create_group delete_group get_group get_gid
+account lock_password unlock_password
 );
 
 =head2 account($name, %option)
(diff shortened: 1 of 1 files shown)


ferki:
@eskaaren: thanks for the patch! Though I think an `ssh_keys` option would be redundant, as `ssh_key` already writes it's value to the `authorized_keys` file, e.g. if you need to deploy multiple keys, in a general case it can be written as:
```perl
create_user 'server_user',
  ...,
  ssh_key => join "\n", @ssh_keys,
  ...;
```

Append would also be just `cat`ting the file contents into an array, and then adding the new keys with `push` or similar.

I would write something like this for your example:
```perl
task 'my_user', sub {
  my @key_files = qw( /root/.ssh/id_rsa.pub /home/$user/.ssh/id_rsa.pub );
  my @remote_keys = split "\n", cat '~/.ssh/authorized_keys';

  LOCAL{
    my @local_keys;
    push ( @local_keys, split "\n", cat $_ ) foreach @key_files;
  }

  create_user "server_user",
    home        => '/home/server_user',
    comment     => 'Demo user',
    expire      => '2100-05-30',
    password    => 'password',
    system      => 1,
    create_home => TRUE,
    ssh_key     => join "\n", @remote_keys, @local_keys;
};
```

Also, it seems you changed the indentation of several lines - I'm not sure if that was intentional or not.

I'm thinking how ssh key management can be improved. Maybe an `ssh_key_file` option would fit better. Which would just read the file and use it as `authorized_keys` on the remote. And if an arrayref is passed, where each element is a keyfile, it could automatically concatenate their content and deploy those.

Even then I'm not fully convinced yet that is should go directly into core - maybe an ssh key management module for the rex-recipes repository would be a better idea for now, until we see how it works out.

eskaaren:
@ferki Thanks for the reply. This solution obviously was just a way to solve this without having deep knowledge of how things work in Rex. I left the current ssh_key part as it was to make it backwards compatible for anyone already using the current implementation but now I see that I could probably have achieved what I needed by doing something like you have suggested.

Anyway it feels a little kludgy (also my implementation) for what one tries to do which should be dead simple.. 

At least I hope you got some ideas for improvement and please let me know if you do a recipe or similar and I can test it.

The indentation happened because of some automatic rule that is configured at work. I will try to fix that for the next patch submitted.

ferki:
@eskaaren: of course, there's always room for improvement ^^ I agree that handling of ssh keys is one of them :) You definitely got me thinking ;)

In my experience it is indeed dead simple in most of the specific use cases we come in contact with, but I'm afraid the general solution is a complex one as seemingly everybody would like to do it in a bit different way, like:
- managing single or multiple keys per user
- the keys can come from:
 - local file(s)
 - remote file(s)
 - or from some external store, like
  - CMDB
  - SQL database
  - LDAP
  - webservice (e.g. Google Apps)
- appending vs "absolute deploy"
- revoking key(s)
- etc..

So let's not underestimate its complexity just yet :)

I's say that kind of manipulation should be left out of Rex _core_ (at least for now), and the `ssh_key` option can be used to just write the content that is compiled "anyhow", depending on the use case. On the other hand, handling the above totally sounds like a rex extension module to me (like `Rex::SSH::KeyManager` or something similar). Which would be configurable, could support multiple sources, etc., and would just return the compiled ssh key(s) as a string.

Like:
```perl
create_user 'myuser',
  ...,
  ssh_key => sub { compile_ssh_keys( user => 'myuser' ); }
```

And from there, who knows, if its features get reasonably stable over time and its design robust, then it can find its way to core quite easily.

But given the above example pseudo-code, it may just become a recommendation/best practice to just encourage everyone to write such a subroutine for their given use case (vs. taking on the maintenance burden of trying to solve everyone's use case at once). Especially given that Rex code is just Perl after all.

I guess there's a sweet middle path somewhere between those "extremes" :)

eskaaren:
You are right that it can be complex to create a general solution for this that works for all scenarios, and when I said "dead simple" I referred to the use of the solution not the creation of it :)

Anyway, keep up the good work. I have used Puppet in previous jobs but this time I will push hard for Rex and will continue sending pull requests now and then.

eskaaren:
Just for future reference, this is a working example:

```

task "my_user", sub {

  my $local_keys = LOCAL{
    my @key_files = qw(
      /root/.ssh/id_rsa.pub
      /home/user/.ssh/id_rsa.pub
    );
    my @tmp;
    push ( @tmp, cat $_ ) foreach @key_files;
    return \@tmp;
  };

  my @remote_keys = split "\n", cat '~/.ssh/authorized_keys';

  create_user "myuser",
              home        => '/home/myuser',
              comment     => 'My user',
              expire      => '2100-05-30',
              password    => 'password',
              system      => 1,
              create_home => TRUE,
              ssh_key     => join "\n", @$local_keys, @remote_keys;
};

```

eskaaren:
The original problem was solved so I am closing this.

PR #747, opened 2015-07-28
Detect services that are not already on runlevels

theappleman:
Services that are not enabled in a runlevel are terminated with a
newline in `rc-config list`. This newline does not match the grep
pattern "\s" and Rex reports that the service does not exist.

Make service_exists look for word breaks "\b". This includes spaces and
newlines.

	$ rc-config list | grep "\sepmd\s"
	$ rc-config list | grep "\bepmd\b"
	  epmd
	$ rc-config list | grep "\bepmd\b" | xxd
	0000000: 2020 6570 6d64 0a                          epmd.
	$ rc-config list | grep "\budev\b" | xxd
	0000000: 2020 7564 6576 2020 2020 2020 2020 2020    udev
	0000010: 2020 2020 2020 2020 2020 2020 7379 7369    sysi
	0000020: 6e69 740a                                  nit.

### DIFF
--- a/lib/Rex/Service/Gentoo.pm
+++ b/lib/Rex/Service/Gentoo.pm
@@ -33,7 +33,7 @@ sub new {
     ensure_stop    => 'rc-update del %s',
     ensure_start   => 'rc-update add %s',
     action         => '/etc/init.d/%s %s',
-    service_exists => 'rc-config list | grep "\s%s\s"',
+    service_exists => 'rc-config list | awk \'!/^(Available i|I)nit scripts/ && $1 ~ /^%s$/{print$1}\'',
   };
 
   return $self;


ferki:
@theappleman: thanks for the patch! :+1: It's a good catch, standard `grep` does not match the end of line with `\s` or `[:space]:`.

ferki:
@theappleman: I'm about to merge this change to an upcoming relase. During testing I found out that `\b` of `grep` matches the word boundary on any character that is not "alphanumeric or underscore".

This for example means `rc-config list | grep "\biptables\b"` matches both `iptables` and `arno-iptables-firewall`. Or `rc-config list | grep "\bnet\b"` matches `net.lo` (and all other services for network interfaces on gentoo).

I believe `rc-config list | grep -E "\s+iptables(\s|$)"` might work correctly. What do you think?

theappleman:
Great catch! I agree, it is surprisingly easy to make `\b` catch other services than it is `\s`. My only concern is compatibility with the `-E` flag (or `egrep`), but I can't reason changing out gnu grep for an alternative.

ferki:
@theappleman: I'm gonna merge it with the updated pattern. I only added `-E` to spare some backslashes in the pattern for easier readability. We also opened a [gentoo bug](https://bugs.gentoo.org/show_bug.cgi?id=558082) about adding an option to make the output more machine-friendy and/or providing an `rc-config exists $service` function.

Until then, I guess the new pattern will do :) Could you update your PR with that, please? I can change it if you want, but then I would change basically the whole PR and I don't want to steal credits :)

theappleman:
I've rebased the branch with master and updated the pattern.

ferki:
@theappleman: oh, man, I found even more corner cases:
- `rc-config list` starts with a header `Available init scripts`, so our pattern matches service name `Available` which obviously doesn't exist, so we need to add `--brief`
- our pattern might match the runlevel part of the `rc-config` output (if someone looks for a service named `default`, or if there's an `iptables` runlevel specified for whatever reason under `/etc/runlevels`, so we need to use at least the start-of-string anchor (`^`)

Maybe it's even easier to run the `rc-config` output through `awk "{print $1}"` to get only the first column, and then match with `grep "^%s$"`. That might be the best option with the less possibilities for errors.

Feel free to send an update with those, but with your permission, I'm just gonna modify the command, and commit/merge it for you like that.

theappleman:
```
awk '!/^(Available i|I)nit scripts/ && $1 ~ /^%s$/{print$1}'
```
This would get the $1 service name and skip the 'Available' line regardless of if it does or doesn't exist (with `--brief`) without the extra pipe to `grep`.

The alt version is for using `rc-config list <runlevel>`, but it's not strictly necessary here:
```
$ rc-config list boot
Init scripts to be started by runlevel boot
...
```

I'll fix up the PR tomorrow.

ferki:
@theappleman: sorry for the complication, but during testing I found out why `grep` was used in the first place. Rex evaluates the `service_exists` command's exit code, and with grep it is success if pattern is found, and failure if pattern not found.

So I believe, the command should be `rc-config --brief list | awk "{print $1}" | grep "^%s$"`, maybe putting the filtering of the header line into awk and dropping `--brief`. But personally I find `--brief` more readable and to the point.

ferki:
Also the `!` looks like to cause problems and/or the parentheses might need escaping. Running `rc-config list | awk \'!/^(Available i|I)nit scripts/ && $1 ~ /^%s$/{print$1}\'` manually with a replaced service name doesn't seem to work for me on Gentoo.

ferki:
So this version is tested and works for me:
```perl
service_exists => q(rc-config --brief list | awk '{print $1}' | grep '^%s$'),
```

ferki:
@theappleman: as I'd like to get things moving forward, I just went ahead and fixed the original issue with the detection command mentioned above. Thanks for your report and help working on this!

PR #749, opened 2015-07-29
Task chaining

kablamo:
# The present
If you issue a command like `rex -G group task1 task2 task3`, Rex will:

 1. Run task1 on all servers, 
 2. Run task2 on all servers,
 2. Run task3 on all servers.  

# The desired future
At $work, we would prefer to see:
 1. Connect to server1 and run task1 && task2 && task3,
 2. Connect to server2 and run task1 && task2 && task3,
 2. Connect to server3 and run task1 && task2 && task3,
 2. etc

The '&&' signifies that task2 won't happen unless task1 is successful.  Success is is determined first by exit status and second by the return value of the task.

# Why this future is desireable

1.  Its quite a bit faster.  The current solution can be slow if you are deploying to lots of servers because it needs to re-connect and re-authenticate and it adds up.  This pr reuses the connection for each task.  

2. To minimize time between steps.  If I do something like change how data is stored in memcache, I want to restart the web server right away after that.  I could build a custom task every time, but there are a large number of possible permutations and task chaining is really a flexible solution.

3. It is often unsafe to run subsequent tasks if the first one fails.

This pr implements these ideas.  I'm happy to improve this pr.  Its admittedly a little rough around the edges, but I wanted to get some review/discussion of this idea, before going farther.

### DIFF
--- a/lib/Rex/CLI.pm
+++ b/lib/Rex/CLI.pm
@@ -562,10 +562,11 @@ CHECK_OVERWRITE: {
     }
 
     if ( defined $ARGV[0] ) {
+      my @tasks;
+
       for my $task (@ARGV) {
         if ( Rex::TaskList->create()->is_task($task) ) {
-          Rex::Logger::debug("Running task: $task");
-          Rex::TaskList->run($task);
+          push @tasks, $task;
         }
         elsif ( $task =~ m/^\-\-/ || $task =~ m/=/ ) {
 
@@ -578,6 +579,9 @@ CHECK_OVERWRITE: {
           );
         }
       }
+
+     #Rex::Logger::debug("Running task: $task");
+      Rex::TaskList->run(\@tasks);
     }
   };
 
--- a/lib/Rex/Task.pm
+++ b/lib/Rex/Task.pm
@@ -748,10 +748,11 @@ sub run {
     # this is a method call
     # so run the task
 
-    my $in_transaction = $options{in_transaction};
+    my $in_transaction    = $options{in_transaction};
+    my $cached_connection = $options{cached_connection};
 
     $self->run_hook( \$server, "before" );
-    $self->connect($server);
+    $self->connect($server) unless $cached_connection;
 
     my $start_time = time;
 
@@ -813,7 +814,7 @@ sub run {
 
     Rex::get_current_connection()->{reporter}->write_report();
 
-    $self->disconnect($server) unless ($in_transaction);
+    $self->disconnect($server) unless $in_transaction || $cached_connection;
     $self->run_hook( \$server, "after" );
 
     return $ret;
--- a/lib/Rex/TaskList.pm
+++ b/lib/Rex/TaskList.pm
@@ -50,16 +50,18 @@ sub create {
 }
 
 sub run {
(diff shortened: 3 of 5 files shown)


ferki:
For sequential execution of multiple tasks, there's already `batch`. I wonder if it would be easier to refactor that to support and recognize "task result signals", and solve the connection caching/reusing separately.

Or `needs` also can be used for handling task dependencies, and it does reuse the existing connection already. This PR seems to break that feature :( For example for a very similar purpose we use something like this at $work:

```perl
use Rex -feature => [qw(exec_autodie)];

task 'compile', sub {
  ...
};

task 'prepare', sub {
  ...
};
...
task 'deploy', group => 'mygroup', sub {
  needs 'compile';
  needs 'prepare';
  needs 'upload';
  needs 'configure';
};
```

Then `rex deploy` runs all tasks sequentially on a single host, reusing the connection. `exec_autodie` stops execution if some commands are failing unexpectedly, and the "sub-tasks" can also just do their own sanity checking to stop the process if needed. Error/sanity checking can also bee done within the `deploy` task, between the `needs` calls.

kablamo:
`batch` doesn't solve my problem as described in the description -- they have the same problem as chained tasks:  The first task runs on every server.  Then the second task runs on every server.  I want to run all the tasks on one server, and then run all the tasks on the next server.

`needs` is interesting.  I didn't know about that.  I'll fix it.  It doesn't solve my problem either because it is not as flexible as task chaining.  Its not always the same 5 tasks so I don't want to hard code it ahead of time.  Task chaining allows me to ask for what I need at the moment I need it instead of hard coding each permutation ahead of time.

ferki:
@kablamo: agreed, having an option for ad-hoc task chaining sounds nice, I commented the above as workaround till it gets implemented. Two more thoughts:
 1. adding the task return value feature we just talked about recently woudl most probably help here massively as well
 2. if we're going to make ad-hoc task chaining possible from the command line, then I think we need to choose a different syntax, as `rex task1 task2 task3` is already used and changing its behaviour would break compatibility (so I guess we need either a feature flag to let people opt-in or a different CLI option, or something)

eskaaren:
+1 for that something like this is needed.

I am using Rex to replace Ant and similar tools that have dependency resolution built in.

Sometimes you just want the script resolving its run order without having to hard-code it in a separate task. 

pseudo example:

```perl
task first => sub {
  depends => [third, second];
  say "first"
};

task second => sub {
  depends => [fourth, third];
  ...
};

task third => sub {
  ...
};

task fourth => sub {
  ...
};
``` 

When running ```rex first``` this would make the following happen:

```
third
fourth
second
first
```

Notice that 'third' is not run again when resolving for 'second' since the dependency already was fulfilled when it ran the first time.

I have tried 'needs', 'before', 'after' etc. but it is not as slick as the more traditional build script style in my opinion. 

Perhaps there is another way of doing this?
Highly sought after feature!

kablamo:
This branch was a nice experiment, but its way more complicated than I thought so I'm going to close this pr.  Continuing work is being coordinated from https://github.com/RexOps/Rex/issues/844 if you are interested in following along.

PR #750, opened 2015-07-30
Make "ensure" work properly on FreeBSD by deleting "\n" when calling …

timp87:
A small change for FreeBSD. It makes "ensure" work. Before that it couldn't recognize that service is already added to /etc/rc.conf.

Looks like the same fix should be applied on OpenBSD and NetBSD.

### DIFF
--- a/lib/Rex/Service/FreeBSD.pm
+++ b/lib/Rex/Service/FreeBSD.pm
@@ -49,7 +49,8 @@ sub ensure {
   }
   elsif ( $what =~ /^start/ || $what =~ m/^run/ ) {
     $self->start( $service, $options );
-    append_if_no_such_line "/etc/rc.conf", "${service}_enable=\"YES\"\n";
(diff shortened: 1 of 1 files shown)


timp87:
I'll be back later with more changes related to BSD.

ferki:
@timp87: thanks for the patch! I was going to merge it, but if you plan to send more changes, I'll wait for them. (Hint: as a general rule, multiple small PRs are preferred over one big "all inclusive" PR, if possible ;)

ps.: as of this moment we only have FreeBSD VM images for automated testing - if you happen to be able to send OpenBSD or NetBSD ones, just ping me on IRC or in a separate issue.

timp87:
Thank you for the fast reply!
I decided to create a my own branch for FreeBSD (later for Open- and NetBSD) to fix bugs that I'll find.
I'll create separate PRs for different parts of the Rex.
Thank you for the Rex!

PR #752, opened 2015-07-31
Make "ensure" work properly on FreeBSD

timp87:
1. A small change for FreeBSD, that makes "ensure" work properly.
Try to make it smart and keep simple as possible.

2. Make (R)?ex work with FreeBSD base system services too.

### DIFF
--- a/lib/Rex/Service/FreeBSD.pm
+++ b/lib/Rex/Service/FreeBSD.pm
@@ -49,9 +49,9 @@ sub ensure {
   }
   elsif ( $what =~ /^start/ || $what =~ m/^run/ ) {
     $self->start( $service, $options );
-    append_if_no_such_line "/etc/rc.conf",
+    append_or_amend_line "/etc/rc.conf",
       line => "${service}_enable=\"YES\"",
-      regexp => qr/^\s*${service}_enable="?[Yy][Ee][Ss]"?/;
+      regexp => qr/^\s*${service}_enable="?([Yy][Ee][Ss]|[Nn][Oo])"?/;
   }
 
   return 1;
--- a/lib/Rex/Service/FreeBSD.pm
+++ b/lib/Rex/Service/FreeBSD.pm
@@ -26,12 +26,12 @@ sub new {
   bless( $self, $proto );
 
   $self->{commands} = {
-    start   => '/usr/local/etc/rc.d/%s onestart',
-    restart => '/usr/local/etc/rc.d/%s onerestart',
-    stop    => '/usr/local/etc/rc.d/%s onestop',
-    reload  => '/usr/local/etc/rc.d/%s onereload',
-    status  => '/usr/local/etc/rc.d/%s onestatus',
-    action  => '/usr/local/etc/rc.d/%s %s',
+    start   => '/usr/sbin/service %s onestart',
+    restart => '/usr/sbin/service %s onerestart',
+    stop    => '/usr/sbin/service %s onestop',
+    reload  => '/usr/sbin/service %s onereload',
+    status  => '/usr/sbin/service %s onestatus',
+    action  => '/usr/sbin/service %s %s',
   };
 
   return $self;
--- a/lib/Rex/Service/FreeBSD.pm
+++ b/lib/Rex/Service/FreeBSD.pm
@@ -44,11 +44,19 @@ sub ensure {
 
   if ( $what =~ /^stop/ ) {
     $self->stop( $service, $options );
+    file "/etc/rc.conf.d/${service}",
+      ensure => "absent";
+    delete_lines_matching "/etc/rc.conf.local",
+      matching => qr/^\s*${service}_enable="?((?i)YES)"?/;
     delete_lines_matching "/etc/rc.conf",
       matching => qr/^\s*${service}_enable="?((?i)YES)"?/;
   }
   elsif ( $what =~ /^start/ || $what =~ m/^run/ ) {
     $self->start( $service, $options );
+    file "/etc/rc.conf.d/${service}",
+      ensure => "absent";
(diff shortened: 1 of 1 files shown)


timp87:
There are still cases when service is disabled in /etc/rc.conf, but recognized enabled by (R)?ex and vise versa.

In general I don't want to complicate "ensure" and I plan to convert it to use /usr/sbin/sysrc in future instead of hell with manual rc.conf editing.
Bad news: a sysrc utility first appeared in FreeBSD 9.2.
Good news: FreeBSD 8.4 EOL is coming in a month or so.

lib/Rex/Service/FreeBSD.pm
```diff
@@ -45,11 +45,13 @@ sub ensure {
   if ( $what =~ /^stop/ ) {
     $self->stop( $service, $options );
     delete_lines_matching "/etc/rc.conf",
-      matching => qr/${service}_enable="YES"/;
+      matching => qr/${service}_enable="?[Yy][Ee][Ss]"?/;
```
ferki: Does the `qr/${service}_enable="?[Yy][Ee][Ss]"?/` pattern has any benefits over normal case-insensitive matching?
Like:
- `qr/${service}_enable="?YES"?/i` (matches the whole expression case-insensitively)
- `qr/${service}_enable="?((?i)YES)"?/` (matches only the `YES` part case-insensitively)

If there's no specific reason to use the character list version, I would say the case-insensitive form is a bit clearer on the intent here.

Same question applies to line 54 below.

lib/Rex/Service/FreeBSD.pm
```diff
@@ -45,11 +45,13 @@ sub ensure {
...
   }
   elsif ( $what =~ /^start/ || $what =~ m/^run/ ) {
     $self->start( $service, $options );
-    append_if_no_such_line "/etc/rc.conf", "${service}_enable=\"YES\"\n";
+    append_or_amend_line "/etc/rc.conf",
+      line => "${service}_enable=\"YES\"",
+      regexp => qr/^\s*${service}_enable="?([Yy][Ee][Ss]|[Nn][Oo])"?/;
```
ferki: Would the leading `^\s*` part of the pattern be needed in the pattern on line 48 above too? Or there's some reason I miss behind the slightly different pattern?

lib/Rex/Service/FreeBSD.pm
```diff
@@ -26,12 +26,12 @@ sub new {
...
-    stop    => '/usr/local/etc/rc.d/%s onestop',
-    reload  => '/usr/local/etc/rc.d/%s onereload',
-    status  => '/usr/local/etc/rc.d/%s onestatus',
-    action  => '/usr/local/etc/rc.d/%s %s',
+    start   => '/usr/sbin/service %s onestart',
+    restart => '/usr/sbin/service %s onerestart',
+    stop    => '/usr/sbin/service %s onestop',
+    reload  => '/usr/sbin/service %s onereload',
+    status  => '/usr/sbin/service %s onestatus',
+    action  => '/usr/sbin/service %s %s',
```
ferki: I agree it looks more _modern_ :+1:

I trust you on this, but can you please point me to some documentation on where it got introduced and/or if it's still in use? I'm a beginner with FreeBSD and I could make great use of such documentation when considering FreeBSD changes ^^

ferki:
@timp87: thanks much for this patch! And sorry for the bit late feedback :/

I consider merging this for the upcoming release, though I left some questions on the changes.

FreeBSD contributions are warmly welcome, and if you think the `/usr/sbin/sysrc` is the way to go, feel free to send another PR for that too, provided it is supported in all current versions of FreeBSD that is supported by upstream (which is 9.3 and 10.2 as of now, so should be OK to change this).

lib/Rex/Service/FreeBSD.pm
timp87: Well, no. It doesn't. Just an old way (a bit shell-ish) to do it, sorry =).
> qr/${service}_enable="?YES"?/i (matches the whole expression case-insensitively)
We shouldn't make the whole expression case-insensitive. It doesn't make sense.
> qr/${service}_enable="?((?i)YES)"?/ (matches only the YES part case-insensitively)
is definitely looks better. I vote for this one! =)
Same story with line 54 below ;)

lib/Rex/Service/FreeBSD.pm
timp87: Hmm, yes. You are right.

lib/Rex/Service/FreeBSD.pm
timp87: You can look into [FreeBSD handbook](https://www.freebsd.org/doc/handbook/), for example, [this page](https://www.freebsd.org/doc/handbook/configtuning-rcd.html).
Or there is a [man page](https://www.freebsd.org/cgi/man.cgi?query=service(8)).

ferki:
@timp87: thanks for the feedback! Can you send an update commit, please?

timp87:
@ferki: thanks! I'm pretty new to git and github. Have I done it right?

ferki:
@timp87: yeah, looks good :+1: I'm gonna rebase it on top if current master and/or squash the multiple commits together upon merge for you.

timp87:
99.99% of FreeBSD folks uses /etc/rc.conf to control services. Let's not break the habits.

ferki:
@timp87: thanks, this is now merged!

I applied some changes though before merging:
- squashed your commits into two: one about using `/usr/sbin/service` for services, and one about making sure ensure works on FreeBSD
- slightly reworded the commit messages
- applied perltidy to keep code formatting

timp87:
@ferki thanks a lot!

PR #754, opened 2015-08-03
Add 'print => boolean' option to 'run' command for printing continuously

eskaaren:
This patch makes it possible for printing the output of an external command continuously while it is running. It only works for local connections since the implementation of that and SSH etc are separate and has to be solved separately.

Using the functionality in this patch it is possible to call external scripts from Rex when using it in Jenkins jobs and similar situations where you want the output to be printed to see the progress of long running tasks.

### DIFF
--- a/lib/Rex/Commands/Run.pm
+++ b/lib/Rex/Commands/Run.pm
@@ -99,6 +99,11 @@ Supported options are:
   creates       => $file_to_create
     tries to create $file_to_create upon execution
     skips execution if the file already exists
+  print           => TRUE
+    Automatically prints using Rex::Logger::info. Used for printing lines from
+    long running processes that has output during the run. The external script
+    must not buffer output if continuous printing is wanted. In a Perl script
+    this can be achieved by setting $|=1.
 
 Examples:
 
@@ -136,7 +141,12 @@ If you want to set custom environment variables you can do it like this:
 If you want to end the command upon receiving a certain output:
  run "my_command",
    end_if_matched => qr/PATTERN/;
-   
+
+If you want to run an external script and print output continuously:
+ run "description",
+   command  => "perl myscript.pl",
+   print    => TRUE;
+
 =cut
 
 our $LAST_OUTPUT; # this variable stores the last output of a run.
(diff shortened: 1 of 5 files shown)


eskaaren:
I have redone this patch so that it now just does the printing and also works with SSH tasks.

eskaaren:
Hmm something happens when I push to GH and whitespaces are changed, that was not intentional.

kablamo:
What do you mean by printing continuously?  I think you mean streaming output like `tail -f`? So you don't have to wait for a long running job to complete before seeing the output, right?

I don't think this patch works...   Have you tried testing it with a command like
```
echo super && sleep 10 && echo duper
```

I recommend you look at this code: https://github.com/RexOps/rex-recipes/blob/1.3/Rex/Commands/StreamRun/__module__.pm

Thats what I used to do streaming output.

eskaaren:
Yes and as you can see the patch prints the lines in the while loops that are used for building the return value if you were to do something like

```perl
my $out = run "...";
```

This works for me, I am using it with some pretty large programs that takes up to 20 minutes to complete with mixed python, php, shell etc.

Also the command you suggested produced the expected output when I tested it here. How did that behave when you tested it?

kablamo:
Ah, I get it now.  Sorry about that.  I tried it and it works.  However, I remember now that I went down this rabbit hole too.  The problem is all the stdout prints before any of the stderr.  I wasn't satisfied with that.  I wanted the output to look exactly like the it would on the terminal with stdout and stderr mixed together.  Thats why I had to use the StreamRun module I linked to above.  In the end, I had to write my own version of `run()` and rewrite parts of StreamRun.  This seems like a high priority problem.

ferki:
I _think_ the situation around this problem domain (confusion about getting one or both of STDOUT and STDERR, optionally like `tail -f` would do, etc.) can also be improved by refactoring the SSH interface modules of Rex to have a (more) consistent behaviour/interface. In fact, I'm looking forward to do some hacking around [`Net::SSH::Any`](https://metacpan.org/pod/Net::SSH::Any).

eskaaren:
Should we close this old PR? There are probably better ways to do the tail -f style print in newer versions of Rex? Or is this still something yet to be implemented?

Tweaki:
Can one of the admins verify this patch?

krimdomu:
Yes i think we can close this. This should work with the run callback already.
Thanks for the heads up.

PR #757, opened 2015-08-14
Added option for rebase during git pull.

eskaaren:
This can be useful if you do one offs for some server that checks out the rest via a repo or want to make sure your live hacking is preserved even if you do another checkout via Rex.

### DIFF
--- a/lib/Rex/Commands/SCM.pm
+++ b/lib/Rex/Commands/SCM.pm
@@ -24,8 +24,8 @@ All these functions are not idempotent.
     url => 'git@foo.bar:myrepo.git';
  
  set repository => "myrepo2",
-    url => "https://foo.bar/myrepo",
-    type => "subversion",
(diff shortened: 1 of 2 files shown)


lib/Rex/Commands/SCM.pm
```diff
@@ -19,26 +19,31 @@ All these functions are not idempotent.
 =head1 SYNOPSIS
 
  use Rex::Commands::SCM;
- 
+
```
ferki: Please add those leading spaces back. They are marking the code block and needed for correct POD formatting.

eskaaren: OK will do. I have an automatic rule that removes that kind of whitespace :/

ferki:
Good idea, thanks!

krimdomu:
hi @eskaaren 
thank you for this patch. I've just merged it into master branch.

PR #758, opened 2015-08-17
Do not silently fail on update_system errors

abeverley:
The ```update_system``` command silently fails in the event of an error. Most of the other Pkg commands correctly report errors; this patch ensures that ```update_system``` does the same.

### DIFF
--- a/lib/Rex/Pkg/Base.pm
+++ b/lib/Rex/Pkg/Base.pm
@@ -94,7 +94,17 @@ sub update_system {
   }
 
   my $cmd = $self->{commands}->{update_system};
-  i_run $cmd;
+  my $f = i_run $cmd;
+
+  unless ( $? == 0 ) {
+    Rex::Logger::info( "Error updating system.", "warn" );
+    Rex::Logger::debug($f);
+    die("Error updating system");
+  }
+
+  Rex::Logger::debug("System successfully updated.");
+
+  return 1;
 }
 
 sub remove {


ferki:
@abeverley: nice, thanks! I think I'm going to merge it for the upcoming 1.3.3 release.

I guess we can find a more descriptive variable name than `$f` and maybe use an `error` level message if rex is going to die, but that is a topic of a different issue.

abeverley:
On Mon, 2015-08-17 at 05:05 -0700, Ferenc Erki wrote:
> @abeverley: nice, thanks! I think I'm going to merge it for the upcoming 1.3.3 
> release.

Great, thanks.

> I guess we can find a more descriptive variable name than `$f` and maybe use an 
> `error` level message if rex is going to die, but that is a topic of a different 
> issue.

Sorry, that was me being lazy - I just copied the code from the other functions in
Pkg.pm.

I originally tried to capture STDERR, but I ran into strange problems whereby the
$err variable was always blank, even though the error text was sent to STDERR on
the host system. I spent a couple of hours debugging, but couldn't get it to work
(strangely it worked okay with other commands, it was just "apt-get upgrade" that
I had trouble with).

Andy

ferki:
>Sorry, that was me being lazy - I just copied the code from the other functions in
Pkg.pm.

No problem, I noticed it's like this throughout that module. I was referring to that as "different issue". It seems like a simple refactor though.

>I originally tried to capture STDERR

Yeah, sometimes it can be tricky, it seems to depend on multiple factors (SSH module being used and/or using tty/no_tty feature flag at least). I'd say lets merge it like this so we have some consistency and some extra noise when something goes wrong. We can go forward from there.

abeverley:
On Mon, 2015-08-17 at 05:14 -0700, Ferenc Erki wrote:
> I'd say lets merge it like this so we have some consistency and some extra noise 
> when something goes wrong.

Great, thanks. Yes, that's what I was thinking - some noise is better than no
noise!

ferki:
@abeverley: thanks again for this patch! I applied perltidy to it before merging.

PR #759, opened 2015-08-17
Improve NetBSD support

timp87:
1. Use the same way to enable service in NetBSD like recently proposed for FreeBSD.

2. Enable/disable services in NetBSD carefully.
In NetBSD we also should keep in mind /etc/rc.conf.d/${service} file. It's a per service config file and its content has higher precedence than /etc/rc.conf settings.

### DIFF
--- a/lib/Rex/Service/NetBSD.pm
+++ b/lib/Rex/Service/NetBSD.pm
@@ -44,18 +44,18 @@ sub ensure {
 
   if ( $what =~ /^stop/ ) {
     $self->stop( $service, $options );
-    delete_lines_matching "/etc/rc.conf",
-      matching => qr/^\s*${service}="?((?i)YES)"?/;
     file "/etc/rc.conf.d/${service}",
       ensure => "absent";
+    delete_lines_matching "/etc/rc.conf",
+      matching => qr/^\s*${service}="?((?i)YES)"?/;
   }
   elsif ( $what =~ /^start/ || $what =~ m/^run/ ) {
     $self->start( $service, $options );
+    file "/etc/rc.conf.d/${service}",
+      ensure => "absent";
     append_or_amend_line "/etc/rc.conf",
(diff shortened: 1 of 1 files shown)


timp87:
Before that, "ensure" for service didn't work. Again, because of trailing "\n" in line parameter.

timp87:
@ferki  How do you think, should we check if service is already enabled before enabling it anyway?

ferki:
@timp87: we can do that, and it would be nice, but I believe if `ensure => "started"` is specified, we really should not care what is the current setting, and we should just start it right away, and register it to be started on boot.

I guess both "check if it's already enabled in the file" and "search and replace to enable it in the file" both needs to read and search the file anyway, so rex won't really spare any operations, while adding complexity.

Fix me if I missed something :)

timp87:
@ferki I agree

timp87:
It should be better to delete a file with higher precedence first, then edit another. Just in case.

timp87:
In case you ask: NetBSD doesn't look into /etc/rc.conf.local file. OpenBSD and FreeBSD do.

timp87:
@ferki, turns out NetBSD packages don't install rc script to /etc/rc.d dir. They assume that user should do it manually. Like
```shell
pkg_add nginx
cp /usr/pkg/share/example/rc.d/nginx /etc/rc.d/
```
Then you can manipulate nginx.
It doesn't even have a flag to do it automatically, because they have common software in base system and in packages. For example, bind(named) is included into the base system and you can install it from packages. Of course, rc script are different. How should we handle this?

timp87:
Sorry, I'm wrong. rc script auto installation is controlled through PKG_RCD_SCRIPTS env var. So users can set it manually in Rexfile

ferki:
@timp87: thanks for this too! I applied perltidy before merging + slightly reworded the commit message.

PR #762, opened 2015-08-19
Fixes #760 timeout for OpenSSH

### LINKED ISSUE #760: Timeout does not work with OpenSSH
From irc discussion, it seems like $timeout is not passed to the constructor when using OpenSSH:

[15:00] <FErki> but judging by a quick glance, the timeout is handled by the Rex::Interface::Connection::OpenSSH::connect() method (like getting the set value or defaulting to 3), but looks like it doesn't actually get passed to Net::OpenSSH constructor
[15:01] <FErki> I wonder why that didn't cause a problem earlier (or why an actual timeout takes _that_ long, like 10+ minutes)
[15:01] <FErki> eskaaren: I guess it's a system-wide or user setting there
[15:02] <FErki> eskaaren: I'd say it's a bug that at least the default timeout is not passwd to the Net::OpenSSH constructor, but that might have been a conscious decision earlier
[15:02] <FErki> I need to dig up the history and see the details
[15:02] <eskaaren> Doing 'ssh -o ConnectTimeout=1 server..' works on the commandline
[15:02] <eskaaren> But I think you have found the issue in Rex
[15:03] <FErki> eskaaren: thanks for the cross-check! Net::OpenSSH->new() accepts a timeout param, this should be quite easy to fix
[15:03] <FErki> eskaaren: can you please open an issue about that? or, if you feel like, a PR :)
[15:07] <FErki> eskaaren: until it's fixed, you can just set any OpenSSH options with e.g. `Rex::Config->set_openssh_opt( ConnectTimeout => 3 );`
[15:09] <FErki> also, `set connection => 'SSH'` forced rex to use Net::SSH2 as SSH implementation in the background, that's why timeouts worked (because Rex::Interface::Connection::SSH respects the timeout setting specified within rex vs. OpenSSH just uses whatever is set on the system or for the user)

eskaaren:
I have verified that this works when setting 
```perl
timeout $int;
``` 
in Rexfile.

### DIFF
--- a/lib/Rex/Interface/Connection/OpenSSH.pm
+++ b/lib/Rex/Interface/Connection/OpenSSH.pm
@@ -79,6 +79,7 @@ sub connect {
   Rex::Logger::debug( Dumper( \%ssh_opts ) );
 
   $ssh_opts{LogLevel} ||= "QUIET";
+  $ssh_opts{ConnectTimeout} = $timeout;
 
   my %net_openssh_constructor_options = (
     exists $ssh_opts{initialize_options} ? $ssh_opts{initialize_options} : () );


ferki:
Thanks! I was thinking to add `timeout => $timeout` to the constructor call of `Net::OpenSSH` on line 147, but this solution is just as fine (or even better).

PR #765, opened 2015-08-21
force apt-listchanges to not run

aowi:
depending on user configuration, it might ask for confirmation and thus
require interactivity. Since it can't tell the difference between apt
invoked by a user, and apt invoked by rex, and since we can't rely on
there being a speciel profile just for rex that we can use, just disable
apt-listchanges completely.


ferki:
@aowi: thanks for the patch!

PR #770, opened 2015-08-24
More fixes for ensure option for FreeBSD services

timp87:
1. Service's rc_script name can differ from rcvar which is used to make service start from boot.
Extract rcvar from rc script.
2. As additional profit we can even complain about incorrect service name.
3. As another profit we can get current rcvar value ($rcvalue) and decide if we actually need to make our manipulations with rc.conf*.
4. Wrap /etc/rc.conf.local editing inside is_file to calm down warnings.
5. Get current service status (already running?) and do our stuff if it's really needed.
6. Edit regexp when we need to enable service in /etc/rc.conf* files. I. e. catch incorrect assignments like 'mysql_enable="bla"'.

perltidy is happy if you care =)

### DIFF
--- a/lib/Rex/Service/FreeBSD.pm
+++ b/lib/Rex/Service/FreeBSD.pm
@@ -43,35 +43,34 @@ sub ensure {
 
   my $what = $options->{ensure};
 
-  my $rcout = i_run "/usr/sbin/service $service rcvar";
-
-  my ( $rcvar, $rcvalue );
-  unless ( ( $rcvar, $rcvalue ) = $rcout =~ m/^(\S+)=(\S+)$/m ) {
-    Rex::Logger::info(
-      "Service $service can't be started: not installed or no such rc scipt.",
-      "error" );
+  my $rccom = "/usr/sbin/service $service rcvar";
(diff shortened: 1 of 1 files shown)


lib/Rex/Service/FreeBSD.pm
```diff
@@ -42,22 +43,43 @@ sub ensure {
 
   my $what = $options->{ensure};
 
+  my $rccom = "/usr/sbin/service $service rcvar";
+  my $rcout = i_run $rccom;
+  if ( $? != 0 ) {
+    Rex::Logger::info( "Running `$rccom` failed", "error" );
+    return 0;
```
timp87: we should decide here and there (line 56) if we really have to do here.

ferki: Looking around in the code, I believe `return 0` is correct here.

ferki:
@timp87: thanks for the updates, this is now merged after rebasing, and squashing!

ferki:
@timp87: I had to run perltidy afterwards, but it is fine. I activated automated tests for it, so we should be able catch untidy code earlier :)

PR #774, opened 2015-08-27
Race condition in Rex::Shared::Var

kablamo:
`Rex::Shared::Var` is a tool for doing IPC (inter process communication).  It uses Storable to write data to a file in one process and read the data from the file in another.  This useful when you want to run tasks in parallel and fork children but need to communicate data between processes.  There are a couple problems with the current implementation:

## Shared lock instead of exclusive lock.

The main problem is it uses a shared lock for both reading and writing.  (See this [stackoverflow question](http://stackoverflow.com/questions/11837428/whats-the-difference-between-an-exclusive-lock-and-a-shared-lock) for more details).  This allows a race condition to occur between reading from the file and writing to the file which could cause a loss of data.  Currently `Rex::Shared::Var` is only used in `Rex::Fork::Task`.  I suspect the problem is likely become more visible if [this pr](https://github.com/RexOps/Rex/pull/697) is merged.

My first commit includes a a test which demonstrates this problem.  If you run `t/shared.t` several times in a row you will usually see the test fail.

The simplest solution is to use an exclusive lock to both read and write.  Processes will block until an exclusive lock becomes available.  

## If users run more than one instance of rex from the same directory data loss occurs

`Rex::Shared::Var` writes data to a file named `vars.db`.  All processes share this file.  For the most part thats a good thing -- the point is for children and parent to communicate after all.  But it becomes bad if you are running more than one instance of `rex -F` from the same directory.

### DIFF
--- a/lib/Rex/Shared/Var/Common.pm
+++ b/lib/Rex/Shared/Var/Common.pm
@@ -1,0 +1,62 @@
+#
+# (c) Jan Gehring <jan.gehring@gmail.com>
+#
+# vim: set ts=2 sw=2 tw=0:
+# vim: set expandtab:
+
+package Rex::Shared::Var::Common;
+
+use strict;
+use warnings;
+
+require Exporter;
+use base qw/Exporter/;
+our @EXPORT_OK = qw/__lock __store __retrieve/;
+
+# VERSION
+
+use Fcntl qw(:DEFAULT :flock);
+use Storable;
+
+our $FILE      = "vars.db.$$";
+our $LOCK_FILE = "vars.db.lock.$$";
+our $PID       = $$;
+
+sub __lock {
+  sysopen( my $dblock, $LOCK_FILE, O_RDONLY | O_CREAT ) or die($!);
+  flock( $dblock, LOCK_EX ) or die($!);
+
+  my $ret = $_[0]->();
+
+  close($dblock);
+
+  return $ret;
+}
+
+sub __store {
+  my $ref = shift;
+  store( $ref, $FILE );
+}
+
+sub __retrieve {
+  return {} unless -f $FILE;
+  return retrieve($FILE);
+
+}
+
+sub END {
+    # $PID gets set when Rex starts.  This value remains the same after the
+    # process forks.  So $PID is always the pid of the master process.  $$
+    # however is always the pid of the current process.  This checks if we are
+    # in the master process or not and only removes the $FILE and $LOCK_FILE if
+    # we are in the master process.
+
+    # return if we exiting a child process
+    return unless $$ eq $Rex::Shared::Var::Common::PID;
+
+    # we are exiting the master process
+    unlink $FILE      if -f $FILE;
+    unlink $LOCK_FILE if -f $LOCK_FILE;
+}
+
+1;
--- a/lib/Rex/Shared/Var/Common.pm
+++ b/lib/Rex/Shared/Var/Common.pm
@@ -18,9 +18,12 @@ our @EXPORT_OK = qw/__lock __store __retrieve/;
 use Fcntl qw(:DEFAULT :flock);
 use Storable;
 
-our $FILE      = "vars.db.$$";
-our $LOCK_FILE = "vars.db.lock.$$";
-our $PID       = $$;
+# $PARENT_PID gets set when Rex starts.  This value remains the same after the
+# process forks.  So $PARENT_PID is always the pid of the parent process.  $$
+# however is always the pid of the current process.  
+our $PARENT_PID = $$;
+our $FILE       = "vars.db.$PARENT_PID";
+our $LOCK_FILE  = "vars.db.lock.$PARENT_PID";
 
 sub __lock {
   sysopen( my $dblock, $LOCK_FILE, O_RDONLY | O_CREAT ) or die($!);
@@ -45,14 +48,8 @@ sub __retrieve {
 }
 
 sub END {
-    # $PID gets set when Rex starts.  This value remains the same after the
-    # process forks.  So $PID is always the pid of the master process.  $$
-    # however is always the pid of the current process.  This checks if we are
-    # in the master process or not and only removes the $FILE and $LOCK_FILE if
-    # we are in the master process.
-
     # return if we exiting a child process
-    return unless $$ eq $Rex::Shared::Var::Common::PID;
+    return unless $$ eq $PARENT_PID;
 
     # we are exiting the master process
     unlink $FILE      if -f $FILE;
--- a/t/shared.t
+++ b/t/shared.t
@@ -2,10 +2,10 @@ use strict;
 use warnings;
 
 BEGIN {
-  use Test::More tests => 8;
-
+  use Test::More tests => 9;
+  use Test::Deep;
+  use Time::HiRes;
   use Rex::Shared::Var;
-
   share(qw($scalar @array %hash));
 }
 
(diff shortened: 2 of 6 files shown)


kablamo:
@ferki This pr is ready for review.

t/shared.t
```diff
@@ -33,5 +32,23 @@ is( $hash{multi}->{key1},      "foo", "multidimension, key1" );
...
-unlink("vars.db");
-unlink("vars.db.lock");
+{
+    @array = (0);
+    my @pids;
+    for my $i (0..5) {
+        my $pid = fork();
+        if ($pid == 0) {
+            # child
+            sleep 1;
```
ferki: Are there any (deterministic) way to test this without a `sleep` (especially a full second long)? Could `pause` of `POSIX` work here?

lib/Rex/Shared/Var/Common.pm
```diff
@@ -0,0 +1,59 @@
+#
```
ferki: Nice refactoring! :+1:

t/shared.t
ferki: Something like this may work, even with fewer forks:
```
{
  @array = (0);
  my @pids;
  for my $i ( 0 .. 5 ) {
    my $pid = fork();
    if ( $pid == 0 ) {
      # child
      $SIG{CONT} = &push_and_exit;
      POSIX::pause;

      sub push_and_exit {
        push @array, 1;
        exit 0;
      }
    }
    push @pids, $pid;
  }

  print kill 'CONT' => @pids;
  waitpid $_, 0 for @pids;

  cmp_deeply \@array, [qw/0 1 1 1 1 1 1/], 'race condition avoided';
}
```
What do you think?

kablamo: Thats an awesome idea.  I did not know about `pause()`.  However your example code has a bug.  Should be
```
$SIG{CONT} = \&push_and_exit;
```
instead of
```
$SIG{CONT = &push_and_exit;
```

The problem is when I run this code it hangs.  Can't figure out why.  :rage1:

kablamo: Ohhh.  the signal is being sent before the signal handlers are setup.

kablamo: Actually the signal is being sent before the children have even forked.  In other words, I'd have to `sleep 1` before `kill CONT => @pids` to make this solution work which would be less reliable and might not work if the CPU is busy.  

I vote we keep it the way it is.

ferki: Yup, yours shuold be the correct syntax, but it also hangs for me. Let's figure it out, as I think `pause` should work, and would be much more elegant.

t/shared.t
```diff
@@ -2,10 +2,9 @@ use strict;
 use warnings;
 
 BEGIN {
-  use Test::More tests => 8;
-
+  use Test::More;
```
ferki: If possible, I would prefer keeping test plans here.

t/shared.t
ferki: Ok, this seems to work for me:
```perl
{
  @array = (0);
  my @pids;
  for my $i ( 0 .. 5 ) {
    my $pid = fork();
    sub push_and_exit { push @array, 1; exit 0; }
    $SIG{CONT} = \&push_and_exit;
    if ( $pid == 0 ) {

      # child
      POSIX::pause;
    }
    push @pids, $pid;
  }

  kill 'CONT' => @pids;
  waitpid $_, 0 for @pids;

  cmp_deeply \@array, [qw/0 1 1 1 1 1 1/], 'race condition avoided';
}
```
Can you also test is please?

kablamo: That still hangs on my machine.  

I don't think its a reliable test with `pause()`.  There is a race condition.  The parent will send the 'CONT' signal as soon as it can.  And the child processes need to fork and setup the signal handler before that happens.

To make this work reliably:
1. the parent should pause before sending the 'CONT' signal
2. the parent should wake up and send a 'CONT' signal each time a child process is ready and has its signal handler setup.  

This doable, but I don't think the complexity is worth it.

lib/Rex/Shared/Var/Common.pm
kablamo: Thanks! :smile:

t/shared.t
kablamo: Added them back.

t/shared.t
ferki: Well, I have a strong feeling against hardcoded `sleep`s, so I'm trying hard to find an alternative without it :)

I think we can wait "just enough" time if we poll all the childs in a while loop with SIGZERO, and if they all return success, send SIGCONT. Plus there can be an `alarm` for 2 seconds, to abort the test with failure, instead of hanging.

I'm willing to spend more effort on this, but obviously I could use a second pair of eyes, and extra testing, as it seems we already have different results for the same code. I ran it 100 times both without and with your patches, and I could reproducibly make it fail or pass.

kablamo: Ok, I tried polling the children with sigzero (another signal I've never used).  
```
{
    @array = (0);
    my $parent = $$;
    print "parent: $$\n";
    my @pids;
    my $ready_pids;
    $SIG{CONT} = sub { print "child $$\n"; $ready_pids++ };

    for my $i (0..5) {
        my $pid = fork();
        if ($pid == 0) {
            # child
            POSIX::pause();
            print "after pause\n";
            push @array, 1;
            exit 0;
        }
        push @pids, $pid;
    }

    while (1) {
        my $success = kill ZERO => @pids;
        last if $success == 6; # the 6 children are ready to go
    }

    my $foo = kill 'CONT' => @pids;
    print "CONT: $foo\n";

    waitpid $_, 0 for @pids;

    cmp_deeply \@array, [qw/0 1 1 1 1 1 1/], 'race condition avoided';
}
```
It almost works, but it still hangs 30% of the time.  If I `sleep1; kill CONT => @pids` a second time, it doesn't hang.  So somehow sigzero is still too early I guess?

krimdomu:
also we should not forget that the tests also run on windows. maybe some of these special signals are not available there?

t/shared.t
kablamo: > If SIGNAL is zero, no signal is sent to the process, but "kill" checks whether it's possible to
> send a signal to it (that means, to be brief, that the process is owned by the same user, or
> we are the super-user).  This is useful to check that a child process is still alive (even if 
> only as a zombie) and hasn't changed its UID. 

  -- from `perldoc -f kill`

> "kill()" doesn't have the semantics of "raise()", i.e. it doesn't send a signal to the 
> identified process like it does on Unix platforms.  Instead "kill($sig, $pid)" terminates 
> the process identified by $pid, and makes it exit immediately with exit status $sig.  
> As in Unix, if $sig is 0 and the specified process exists, it returns true without 
> actually terminating it. (Win32)

  -- from `perldoc perlport`

ferki: Ok, how about merging it with `sleep`, but if possible, a shorter one (like 0.1), and marking it as a `#TODO: would be nice to handle it with POSIX::pause, see #774 for details`?

krimdomu: +1

kablamo:
I'm developing on a fairly beefy server so 0.1 seconds fails for me on the master branch like it should.  Not sure if it will reliably fail on lesser machines?  You may want to try it on your box.

kablamo:
I guess you want me to squash all the commits into one commit?

kablamo:
I rebased and did a force push to make the commit prettier.  I believe this is ready to merge.

ferki:
Thanks, @kablamo!

ferki:
Hmm..it doesn't seem like rebased on top of current master, and I find the `Avoid race condition when doing IPC.` commit message slightly confusing (as the content is a refactor + changing shared lock to exclusive lock), but let's get going :)

kablamo:
Noted for next time.  Glad to see it merged!  \o/

ferki:
Meh, plus it looks like `Rex::Shared::Var::Common` was not tidy.. :( Fixed.

Issue #622, opened 2015-04-07
Rex doesn't exit when there is a syntax error


kablamo:
It hangs forever and I have to do control-c.  A minor low priority thing, I know.

ferki:
@kablamo: thanks for the report! Could you provide an example for us to reproduce this? Also which version of Rex you used when experiencing this? There is already a syntax checking logic before trying to use modules/Rexfiles, it probably just misses a corner case we'd like to fix.

ferki:
@kablamo: do you have any more details on this? Can you share an example Rexfile which causes this on your machine, along with a debug output?

krimdomu:
When there is a syntax error, rex is doing a recursive scan over the directory where the Rexfile is and try to do a syntax check on all "Rexfile" and "*.pm" files to print out all syntax errors.

Maybe this is the problem in your case. Do you have a large directory tree?

kablamo:
Yeah it happens when I introduce any kind of syntax error.

Yes I do have a large directory tree.  I bet thats the problem.  I'm using Carton so I have a bunch of modules installed in ./local/lib/perl5.  Maybe rex should just look at the lib directory?  Hmm or it could skip ./local/lib.  Or it could skip anything in .gitignore.

ferki:
@kablamo: I believe this was fixed on #778. Can you confirm that?

ferki:
As far as I can tell, it should be already fixed on #778, so closing this. If there are any further troubles with syntax checking, please open a new issue.

PR #778, opened 2015-08-30
Better error handling when loading the Rexfile

kablamo:
The code for loading the Rexfile seems to be unnecessarily complicated and doesn't work well.  Here are some of the problems this fixes:

  1. Can't use `perl -d:Confess rex` to force a stack trace
  2. I use Carton and Rex appears to hang whenever there is a syntax error because Rex tries to syntax check every file in `./local/lib` which takes forever (see https://github.com/RexOps/Rex/issues/622).
  3. The current code seems to me way way more complicated than it needs to be.
  4. I also refactored out some logic from `Rex::CLI::_run__()`.  It was about 600 lines long.  Now its only 400 lines long.  Stuff I refactored out:
    1. Logic for loading the server ini file is now in `load_server_ini_file()`.
    1. Logic for loading the Rexfile is now in `load_rexfile()`.
    1. Logic for handling the lockfile is now in `handle_lockfile()`.

### DIFF
--- a/t/include.t
+++ b/t/include.t
@@ -1,0 +1,29 @@
+use strict;
+use warnings;
+use lib 't/lib';
+
+use Test::More tests => 2;
+use Test::Deep;
+
+use Rex::Commands;
+use Rex::Commands::Run;
+
+use t::tasks::cowboy;
+
+$::QUIET = 1;
+
+my $task_list = Rex::TaskList->create;
+
+my @task_names = $task_list->get_tasks;
+cmp_deeply
+    \@task_names,
+    [qw/t:tasks:cowboy:roundup/],
+    "found visible task";
+
+my @all_task_names = sort $task_list->get_all_tasks(qr/.*/);
+cmp_deeply 
+    \@all_task_names,
+    [qw/t:tasks:alien:negotiate t:tasks:cowboy:roundup/],
+    "found hidden task";
+
+done_testing();
--- a/lib/Rex/CLI.pm
+++ b/lib/Rex/CLI.pm
@@ -177,9 +177,7 @@ FORCE_SERVER: {
     Rex::Output->get( $opts{'o'} );
   }
 
-  # Load Rexfile before exec in order to suppport group exec
-  if ( -f $::rexfile ) {
-    Rex::Logger::debug("$::rexfile exists");
+    # handle lock file
     if ( $^O !~ m/^MSWin/ ) {
       if ( -f "$::rexfile.lock" && !exists $opts{'F'} ) {
         Rex::Logger::debug("Found $::rexfile.lock");
@@ -236,14 +234,6 @@ FORCE_SERVER: {
     if ($@) { print $@ . "\n"; exit 1; }
 
     load_rexfile($::rexfile);
-  }
-  else {
-    Rex::Logger::info( "No Rexfile found.", "warn" );
-    Rex::Logger::info(
-      "Please create a file named 'Rexfile' inside this directory,", "warn" );
-    Rex::Logger::info( "or specify the file you want to use with:", "warn" );
-    Rex::Logger::info( "   rex -f file_to_use task_to_run",         "warn" );
-  }
 
   #### check if some parameters should be overwritten from the command line
 CHECK_OVERWRITE: {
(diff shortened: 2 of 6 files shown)


kablamo:
Rebased to make the commits pretty.  I think the best way for you to review this is to look at each commit.  They are in nice bite sized chunks which makes this pr a lot easier to understand.

kablamo:
This is ready for review

ferki:
@kablamo: nice, love the refactors! I left two notes about preserving some functionality.

I'd also like to ping @krimdomu for opinion here, because I believe there was a reason to add the complexity in the first place, but that reason might have gone away by now.

kablamo:
I just tested a very simple rex project with PAR using this branch and it worked for me.  Not sure why that check for PAR is there.  If I comment it out, PAR works fine.  Hmmm... maybe its because if you specify a Rexfile via -f and use a relative path.  The PAR self-extracts into a /tmp/rex.build.$$ so relative paths for the Rexfile would not work.  That wouldn't be an issue with my solution.

ferki:
Yeah, probably would be fine. Just that amount of previous complexity is a bit weird, if it can be done with much less :)

kablamo:
ping @ferki @krimdomu.  you guys are busy i'm sure.  just don't want you to forget about me.  :)

ferki:
@kablamo: I'd like to confirm that we haven't forgotten about you at all ;)

kablamo:
Rebased on current master, made a few fixes, and force pushed.  Ready for review.

krimdomu:
i think this code looks good.
With this we can also remove some internal things that we had to add because the Rexfile gets parsed twice.
What we need to check is, that if there are syntax errors in modules, that this errors get displayed.

kablamo:
@krimdomu Not sure I understand what it is you want to test.  Here is what I tested:

Rexfile with syntax error
```
eric@va-bastion1:~/code/rex-src ⚡ cat > foo
use Rex -feature => [qw/ 1.3 /];
task test0 => sub { say run "pwd" };
print("hi"; # <-- syntax error here
1:
eric@va-bastion1:~/code/rex-src ⚡ perl bin/rex -f foo test0
[2015-11-27 21:30:14] ERROR - Compile time errors:
syntax error at foo line 3, near ""hi";"
Compilation failed in require at lib/Rex/CLI.pm line 713, <> line 6.
[2015-11-27 21:30:14] INFO - Exiting Rex...
[2015-11-27 21:30:14] INFO - Cleaning up...
```

Rexfile without syntax error
```
eric@va-bastion1:~/code/rex-src ⚡ cat > foo
use Rex -feature => [qw/ 1.3 /];
task test0 => sub { say run "pwd" };
1;
eric@va-bastion1:~/code/rex-src ⚡ perl bin/rex -f foo test0
[2015-11-27 23:42:28] INFO - Running task test0 on <local>
/home/eric/code/rex-src
[2015-11-27 23:42:28] INFO - All tasks successful on all hosts
```

kablamo:
@ferki this pr has been around a while, would be awesome to get it reviewed so i can merge.  I know you are busy which is fine.  Just saying itd be cool if you looked at this pr first.

t/include.t
```diff
@@ -0,0 +1,24 @@
+use Test::Most;
+use Test::Lib;
```
ferki: We need to make sure if `Test::Lib` is available on our usual targets as a package (otherwise we need to package it for those distros which are missing it).

If there's an easy way to test without `Test::Lib`, it might worth considering to replace it with either a perl core module or equivalent.

ferki:
Apart from the `Test::Lib` remark I left, this is working great for me for errors both in Rexfile or in an included module :+1:

It's ready to be merged once we made sure `Test::Lib` is available on our supported platforms (at least by building it for them), or we replaced it with equivalent functionality.

t/include.t
ferki: Perhaps `use lib 't/lib';` is good enough?

kablamo:
updated `t/include.t` and force pushed.

ferki:
@kablamo: this is now merged. Great job, thanks!

ferki:
ps.: I fixed up some small perltidy issues afterwards, I need to talk to Jenkins to detect those more reliably.

Also it looks like I missed `Test::Deep` as an additional non-core dependency, but looks like we are already started to build that for our package repositories, so I guess it's OK (and I'm sure we don't really want to build the same functionality on our own :).

I think we can think about those dependencies more once we released 1.4.0. I'd love to use more modules from CPAN too, and I believe it's largely just a question of finding good balance between "ease of development" and "ease of packaging/maintenance".

kablamo:
Yay! :balloon:  Thanks!  

Yeah not being able to use whatever cpan dependency is useful is not fun and I'd love to find a solution for that.  I really like `Exporter::Shiny` and `Path::TIny` and tend to use them everywhere.  And `Moo`.  etc.  Anyway thats a distraction.  We can talk about that another time.

PR #779, opened 2015-08-31
Better argument parsing to support task chaining

kablamo:
# Overview

Currently rex has very limited support for task chaining.  The following works:
```
rex task1 task2 task3
```

But as soon as you start passing options to tasks, things break down in a couple of ways.  The goal of this pr is to fix this.

# Concepts

A `Rex::TaskList` object holds a *hash* defining every task defined in a user's Rexfile.

A `Rex::RunList` object holds a *list* of all the `Rex::Task` objects that are going to get executed in a single run of rex.  In other words, it contains all the tasks specified by the user on the command line.  Note that `Rex::Task` objects now contain `args` and `opts` attributes.

# Fixes

  * `Rex::Args` parses @ARGV every time it was used.  Moved code out of `import()`.
  * Users can pass in multiple tasks to rex on the command line.  But `Rex::Args` assumes all options are for the first task.  Argument parsing is now fixed in `Rex::RunList->parse_opts()`.  Example:
```
rex lib:task1 --option1=value1 lib:task2 --option1=value2
```
  * Even if Rex::Args parses the arguments correctly, rex can't run a task more than once with different options using only the `Rex::TaskList` concept because it contains a hash of all the tasks (and hashes don't allow duplicate keys).  Fixed this by adding the `Rex::RunList` concept which is based around an arrayref instead of a hashref.  Example:
```
rex lib:task1 --option=value1 lib:task2 --option=value2
```
  * Tasks can now accept arguments as well as options. Example:
```
rex lib:task1 <arg> --option=value
```

# Note

I structured my commits neatly (I hope).  It may help to review them one at a time to understand this pr.

### DIFF
--- a/lib/Rex.pm
+++ b/lib/Rex.pm
@@ -786,6 +786,12 @@ sub import {
         $found_feature = 1;
       }
 
+      if ( $add eq "task_chaining_cmdline_args" ) {
+        Rex::Logger::debug("Enabling task_chaining_cmdline_args feature");
+        Rex::Config->set_task_chaining_cmdline_args(1);
+        $found_feature = 1;
+      }
+
       if ( $found_feature == 0 ) {
         Rex::Logger::info(
           "You tried to load a feature ($add) that doesn't exists in your Rex version. Please update.",
--- a/lib/Rex/RunList.pm
+++ b/lib/Rex/RunList.pm
@@ -10,7 +10,6 @@ use strict;
 use warnings;
 
 use Rex::Logger;
-use Rex::Task;
 use Rex::TaskList;
 
 # VERSION
--- a/lib/Rex/RunList.pm
+++ b/lib/Rex/RunList.pm
@@ -1,0 +1,119 @@
+#
+# (c) Jan Gehring <jan.gehring@gmail.com>
+#
+# vim: set ts=2 sw=2 tw=0:
+# vim: set expandtab:
+
+package Rex::RunList;
+
+use strict;
+use warnings;
+
+# VERSION
+
+use Rex::Logger;
+use Rex::Task;
+use Rex::TaskList;
+#use Rex::Config;
+#use Rex::Interface::Executor;
+#use Rex::Fork::Manager;
+#use Rex::Report;
+#use Rex::Group;
+#use Time::HiRes qw(time);
+#use POSIX qw(floor);
+
+my $INSTANCE;
+
+sub new {
+  my ($class, %args) = @_;
+  return bless \%args, $class;
+}
+
+# returns a singleton
+sub instance {
+  my $class = shift;
+  return $INSTANCE if $INSTANCE;
+  $INSTANCE = $class->new(@_);
+}
+
+sub add_task {
+  my ($self, $task_name, $task_args, $task_opts) = @_;
+  $task_args ||= [];
+  $task_opts ||= {};
+  my $task = $self->task_list->get_task($task_name)->clone;
+  $task->set_args(@$task_args);
+  $task->set_opts(%$task_opts);
+  push @{ $self->{tasks} }, $task;
+}
+
+sub current_index {
+  my ($self) = @_;
+  return $self->{current_index} || 0;
+}
+
+sub increment_current_index {
+  my ($self) = @_;
+  return $self->{current_index} += 1;
+}
+
+sub current_task {
+  my $self = shift;
+  my $i = $self->current_index;
+  $self->{tasks}->[$i];
+}
+
+sub tasks { @{ shift->{tasks} } }
+
+sub task_list    { 
+  my $self = shift;
+  return $self->{task_list} if $self->{task_list};
+  $self->{task_list} = Rex::TaskList->create;
+}
+
+sub run_tasks {
+  my ($self) = @_;
+  
+  for my $task ($self->tasks) {
+    $_->($task) for @{ $task->{before_task_start} };
+    $self->task_list->run($task);
+    $_->($task) for @{ $task->{after_task_finished} };
+    $self->increment_current_index;
+  }
+}
+
+# Parse @ARGV to get tasks, task args, and task opts.  Use these values to
+# generate a list of tasks the user wants to run.
+sub parse_opts {
+  my ($self, @params) = @_;
+
+  while (my $task_name = shift @params) {
+    die "Expected a task name but found '$task_name' instead\n"
+      unless $self->task_list->is_task($task_name);
+
+    my @args;
+    my %opts;
+
+    while (my $param = shift @params) {
+      if ($self->task_list->is_task($param)) {
+        unshift @params, $param;
+        last;
+      }
+
+      if ($param =~ /^--/) {
+        my ($key, $val) = split /=/, $param, 2;
+        $key =~ s/^--//;
+
+        $opts{$key} = defined $val ? $val : 1;
+      }
+      else {
+        push @args, $param;
+      }
+    }
+
+    $self->add_task($task_name, \@args, \%opts);
+  }
+
+  return $self;
+}
+
+1;
--- a/lib/Rex/RunList.pm
+++ b/lib/Rex/RunList.pm
@@ -78,6 +78,9 @@ sub run_tasks {
 sub parse_opts {
   my ($self, @params) = @_;
 
+  return $self->deprecated(@params) 
+    unless Rex::Config->get_task_chaining_cmdline_args;
+
   while (my $task_name = shift @params) {
     die "Expected a task name but found '$task_name' instead\n"
       unless $self->task_list->is_task($task_name);
@@ -106,6 +109,40 @@ sub parse_opts {
   }
 
   return $self;
+}
+
+sub deprecated {
+  my ($self, @params) = @_;
+
+  my $msg = <<EOF;
+The default way command line arguments work will change somewhat in a future
+release.  Major changes include:
+ - Fixes for argument handling for multiple tasks (aka chained tasks).  
+ - Tasks accept arguments as well as options.  
+Use the feature 'task_chaining_cmdline_args' to remove this warning and enable
+the new features.  For more info see:
+http://www.rexify.org/docs/other/a_word_on_backward_compatibility.html
+EOF
+  Rex::Logger::info($msg, 'warn');
+
+  #### parse task options
+  my %task_opts;
+
+  for my $p (@params) {
+    my ( $key, $val ) = split( /=/, $p, 2 );
+
+    $key =~ s/^--//;
+
+    if ( defined $val ) { $task_opts{$key} = $val; next; }
+    $task_opts{$key} = 1;
+  }
+
+  for my $task_name (@params) {
+    next if $task_name =~ m/^\-\-/ || $task_name =~ m/=/;
+    next unless $self->task_list->is_task($task_name);
+
+    $self->add_task($task_name, [], \%task_opts);
+  }
 }
 
 1;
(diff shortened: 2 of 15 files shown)


kablamo:
Nice to have: save task params/args to the task object in `Rex::Task->run()` so that hooks have access to the them even when `run_task()` is used.

ferki:
It looks like breaking the current interface. I think we would need at least a feature flag to activate these things.

As a slightly related note, we also consider replacing Rex's argument parsing with some external module if possible (instead of brewing our own).

kablamo:
The tests pass now and this is ready for review.

Can you give me a little checklist of what things you view as breaking backwards compatibility so we need a feature flag?  I thought I did not break backwards compatibility.  For example, I did change Rex::Args, but that has no pod and is not documented.  I thought that is the standard we use to determine what is safe to change?

ferki:
Well, at least:
- [ ] failing tests are good indicators at first (but worth doublecheking if the tests are still correct)
- [ ] changing the public API (the Rex DSL), other than adding new options/commands, is likely to break things, so extra checks might needed to ensure backwards compat (e.g. removing an option or a command is a breaking change, because someone might rely on it, even if it's considered bad practice; if we add a new command or option, then it's usually very hard to break existing rex task, because nobody uses them yet)
- [ ] leaving the public API alone, but changing the internal implementation in a way that gives different results, is also likely to break backwards compat (unless it's a bugfix, but lots of times it's not that clear, e.g. we thought allowing rex to continue running with a non-existing environment specified on the command line was a bug, then it turned out there's a valid use-case for it, so we went back to just warn about it and recommend a setting to suppress the warning; or the recent MD5 checksum calculation change)

Or if in doubt, it's always safe to be paranoid and assume that it breaks backwards compat, and add a feature flag :) Feature flags are the way with which we can introduce breaking changes and let people opt-in/opt-out for the new behaviour. We plan to pair that up with planned deprecation, e.g. in cases when the old implementation is clearly bad/wrong/unmaintained, it's good to include a deprecation warning in the next release for it, so people still using that can upgrade, then remove the deprecated part in a later release. We don't yet do this, but the option is there and would allow us to drop some ballast.

kablamo:
I appreciate the response.  That is helpful.  I don't believe this pr violates any of those 3 situations.  What do you think?

krimdomu:
hi @kablamo 
i like this pr much and it will make the cli parsing better but sadly it break some things. We could add this if we also integrate #801 so that we can check for a feature flag. (after thinking a bit longer, it might be possible to add this without #801)

For example what it breaks (which can be fixed with a feature flag without the need of #801):
There are 2 tasks which just do `print Dumper $_[0]` inside the tasks.

With your pr:
```
> rex test1 test2 --name=foo
[2015-10-09 12:09:45] INFO - Running task test1 on <local>
$VAR1 = {};
[2015-10-09 12:09:46] INFO - Running task test2 on <local>
$VAR1 = {
          'name' => 'foo'
        };
```

Current master:
```
> rex test1 test2 --name=foo

[2015-10-09 12:09:58] INFO - Running task test1 on <local>
$VAR1 = {
          'name' => 'foo',
          'test2' => 1
        };
[2015-10-09 12:09:58] INFO - Running task test2 on <local>
$VAR1 = {
          'name' => 'foo',
          'test2' => 1
        };
```

The output of "test2 => 1" is a bug which i didn't fixed yet because it doesn't hurt. It just added all the other task names into the parameter list. (`rex task1 task2` will result in an task opton "task2 => 1")

But some people relay on the current behaviour that all tasks gets the provided parameters.

I'm going to think a bit more about it. I just wanted to give a feedback.

kablamo:
Thanks for the review @krimdomu.  

1. You want to provide backwards compatibility for a bug?  How many users do this?  How do you know users do this?
2. I can add a feature flag which handles the above scenario.  In `Rex::RunList::parse_opts()`. I'll add an `if` statement which only uses the last set of params and throws away anything before that to recreate the bug.
3. Is that the only obstacle here?

krimdomu:
hi @kablamo 
only the `test2 => 1` in the parameters is a bug. It is not a bug that every task gets the same parameters.
And yes, me is using this in many places. ;-)

kablamo:
Oh right!  I said the wrong thing.  Thanks for the correction.   I guess you also want to preserve all of the bad old behavior:

```
~/code/rex-src ⚡ bin/rex -f Rexfile test1 --boo=hoo --a=1 test2 --name=foo --a=2
[2015-10-14 08:38:35] INFO - Running task test1 on <local>
$VAR1 = {
          'boo' => 'hoo',
          'a' => '2',
          'name' => 'foo',
          'test2' => 1
        };
[2015-10-14 08:38:35] INFO - Running task test2 on <local>
$VAR1 = {
          'boo' => 'hoo',
          'a' => '2',
          'name' => 'foo',
          'test2' => 1
        };
```

I will change `Rex::RunList::parse_opts()` to do this behavior depending on whether the feature is enabled/disabled.  

FWIW though it seems a bit wasteful to provide backwards compatibility for a bug.  We could spend that time adding features and fixing bugs.  Just a thought.

krimdomu:
@kablamo 
i don't think you said the wrong thing. If my answer was a bit harsh i'm sorry for this.

I think it depends on the point of view of the user if it is a bug or not. And the current behavior is in use, so if we change it from one version to another the users might get problems.

But you are right, spending time on backward compatibility consumes time. I don't want to maintain backward compatibility for all times, but we need it a least for some versions. Currently we don't have a strategy to drop old code. We need to think about this. And then we can drop old code in a way the users know.

ferki:
I guess this is aperfect opportunity to practice planned deprecation. The argument parsing is like "as it is" since a long time, it's safe to assume that there are people who actually rely on the current behaviour (even if we all agree it's a bug, and document it, and make it clear in the docs, tweet about it, etc. :).

IMHO, it's better to make this obvious and somewhat annoying, by e.g. adding a warning level log message to the output whenever there are multiple tasks passed to rex with task parameters, and mentioning that "this interface will change soon(tm), make sure your code works with feature flag X, if you find regressions, please report, and this will be default in the next release".

People will learn about it, have instructions what to check, and what to do, and will possibly report use cases we never ever dreamt of. We can then decide when to enable this as default, and just drop the old behaviour.

I think that's a clean and relatively painless approach, but we need to start planned deprecations to drop some ballast.

krimdomu:
@ferki :+1:

kablamo:
Cool.  Sounds reasonable.  I think deprecation is a great way to go.  Reduces complexity and makes the code easier to change.  I'll add some deprecation warnings and a feature flag.

ferki:
There's already `Rex::deprecated()` which can help us.

Though I would not make any assumptions about which rex version will drop a specific deprecated feature. Let's just use something like "This is marked as deprecated and will be removed in the future. Please do _this and this_"). And maybe bump the severity to `warn`, so it will be more obvious in the output.

I think the changes might be small enough to include here, but let me know if I should handle this on a separate PR.

kablamo:
Ready for review.

krimdomu:
@kablamo 
thanks for the enhancement. This looks good. :+1: 

Currently the tests are failing because i think the feature flag is missing in the tests:
```perl
Rex::Config->set_task_chaining_cmdline_args(1);
```
Could you please add this? And run perltidy over the files?

Then i can merge this pull request.

Thanks,

Jan

kablamo:
Fixed test + rebased on current master + ran perltidy

krimdomu:
Hi @kablamo 
thank you very much. I've just merged this request into master branch :+1:  :-)

kablamo:
\o/ :balloon:

ferki:
Hmm..the overall commit message became a bit confusing, but this PR is nevertheless a great job! :+1:

lib/Rex/RunList.pm
```diff
@@ -0,0 +1,148 @@
...
+    $self->increment_current_index;
+  }
+}
+
+# Parse @ARGV to get tasks, task args, and task opts.  Use these values to
+# generate a list of tasks the user wants to run.
+sub parse_opts {
+  my ( $self, @params ) = @_;
+
+  return $self->deprecated(@params)
```
ferki: Why don't we use `Rex::deprecated()` here? Why we implemented `Rex::RunList::deprecated()` instead?

ferki: Ok, so `Rex::RunList::deprecated` is only executed if the `task_chaining_cmdline_args` feautre flag is not enabled. But the naming is still confusing. IMHO `Rex::deprecated` should be used to write out deprecation messages.

lib/Rex/RunList.pm
```diff
@@ -0,0 +1,148 @@
...
+
+  my $msg = <<EOF;
+The default way command line arguments work will change somewhat in a future
+release.  Major changes include:
+ - Fixes for argument handling for multiple tasks (aka chained tasks).  
+ - Tasks accept arguments as well as options.  
+Use the feature 'task_chaining_cmdline_args' to remove this warning and enable
+the new features.  For more info see:
+http://www.rexify.org/docs/other/a_word_on_backward_compatibility.html
+EOF
```
ferki: Can we find a better message here? For me as a user, it's not clear what will change, and whether I should change something. The linked compatibility page doesn't have any info on the mentioned feature flag. Can we link to a documentation of task arguments instead? Is it wise to use links here? What if the link structure changes and we start pointing to non-existing resources?

lib/Rex.pm
```diff
@@ -786,6 +786,12 @@ sub import {
         $found_feature = 1;
       }
 
+      if ( $add eq "task_chaining_cmdline_args" ) {
```
ferki: It'd be nice if we could add a counterpart for this, like `no_task_chaining_cmdline_args`

ferki:
I'm sorry for being a bit late to the party, and just seemingly ranting around randomly, but I find the way how it was merged very confusing (i.e. everything squashed into a single big commit, or how depreacation is handled in th end). Since it's in master, it's too late to cleanup/change, so I think I'll just refactor some of the stuff included here.

krimdomu:
Hi Ferki,

i think it is better to open new tickets for these changes and not to comment in the old prs. Also i think the squashing is the right way, but we can talk on this in irc.

krimdomu:
And yes, we need a no_* and also we need to enable this only if 1.4.0 feature is set and not by default

ferki:
You're right, I just got frustrated because it's hard to understand what's going on in the big commit that appeared on master, and rex started to throw big warnings on every call :) Sorry about that.

lib/Rex/RunList.pm
kablamo: I agree with you.  But I would have had to change how `Rex::deprecated()` works and I felt that was out of scope of what this task is about.  My prs are too big which makes them take forever to get merged and I need to get them to be smaller.

lib/Rex/RunList.pm
kablamo: 1. It doesn't matter to me where it sends people.  You guys always send me to the compatibility page on irc so I figured that was the place to go.  
2. How is it not clear what changed?  It tells you the default is changing and then lists the 2 major changes.  Maybe the descriptions are not clear?  
3. If you have any suggested solutions, I'm happy to hear what they are.

kablamo: I think it would be great if we had a BackwardsCompatability.pm file which listed the changes as pod.  And then the website prettified that and displayed that.  That way people could view those on metacpan.

kablamo:
No worries @ferki .  I get frustrated some days too.  I do think its good that you are communicating what you want.  I'm glad to have feedback.  

I thought you said above that you wanted the message promoted to a warning?  https://github.com/RexOps/Rex/pull/779#issuecomment-148050897

PR #787, opened 2015-09-07
Ask FreeBSD system where the rc.conf files and rc.conf.d dirs are.

timp87:
Use same approach as FreeBSD's rc initialization code. That's how FreeBSD boots.
https://github.com/freebsd/freebsd/blob/master/etc/rc.subr#L1336-L1354
It looks really ugly for me, but IMO it's more right than just assuming where the right files and dirs are.

Feel free to criticize and even skip this approach.

P.S. I'm not sure if these are the best names for vars.
P.S.S. I think this is almost the last peace of code I'd like to change in Service/FreeBSD.pm (service existence check is left, not sure where ;)).

### DIFF
--- a/lib/Rex/Service/FreeBSD.pm
+++ b/lib/Rex/Service/FreeBSD.pm
@@ -43,41 +43,71 @@ sub ensure {
 
   my $what = $options->{ensure};
 
-  my $rccom = "/usr/sbin/service $service rcvar";
-  my $rcout = i_run $rccom;
+  my $get_rc_out = "/usr/sbin/service $service rcvar";
+  my $rc_out     = i_run $get_rc_out;
   if ( $? != 0 ) {
-    Rex::Logger::info( "Running `$rccom` failed", "error" );
+    Rex::Logger::info( "Running `$get_rc_out` failed", "error" );
     return 0;
   }
 
-  my ( $rcvar, $rcvalue ) = $rcout =~ m/^\$?(\w+)="?(\w+)"?$/m;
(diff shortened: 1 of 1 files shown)


lib/Rex/Service/FreeBSD.pm
```diff
@@ -43,41 +43,71 @@ sub ensure {
...
+    Rex::Logger::info( "Running `$get_rcconf_out` failed", "error" );
+    return 0;
+  }
+  my @rcconf = split /\s+/, $rcconf_out;
+  unless (@rcconf) {
+    Rex::Logger::info( "Error getting rc.conf files", "error" );
+    return 0;
+  }
+
+  my $get_rcconfd_out =
```
timp87: maybe this way?
```perl
-  my $get_rcconfd_out =
-    "/bin/sh -c '. /etc/rc.subr; load_rc_config 'XXX'; for dir in /etc \$local_startup; do dir=\${dir\%/rc.d}; echo \${dir}/rc.conf.d; done'";
+  my $get_rcconfd_out = "/bin/sh -c ' \\
+    . /etc/rc.subr; \\
+    load_rc_config 'XXX'; \\
+    for dir in /etc \$local_startup; \\
+    do \\
+      dir=\${dir\%/rc.d}; \\
+      echo \${dir}/rc.conf.d; \\
+    done \\
+  '";
```

timp87: rename every rcconfd -> rcconf_d?

timp87:
Let's keep (R)?ex code simple.
FreeBSD is actually mature OS and nothing much has chaged in this area for a long time .

Issue #788, opened 2015-09-09
first call of run_task might run on the wrong host


krimdomu:
(just a reminder, i need to figure out the circumstances)

PR #789, opened 2015-09-09
fixes the problem of run_task connecting to the wrong host #788

krimdomu:
need to write some tests for it
references #788


ferki:
I created a rex-build issue for it, but I'd like to merge it to fix this bug.

Issue #792, opened 2015-09-15
Rex / Net::OpenSSH guesses wrong username 


ipcjk:
I am using su - to switch to my deployment account. For whatever reason rex/net::openssh (0.64) is pulling out my original ssh connection username, even though I have an clear environment. 

joerg:~$ sshg3 joerg@linde
Authentication successful.
joerg@linde:~$ su - sysop-rex
Password: 
sysop-rex@linde:~$ rex  -dd -H linde2 uptime
[2015-09-15 10:44:09] DEBUG - This is Rex version: 1.3.2
[2015-09-15 10:44:09] DEBUG - Command Line Parameters
[2015-09-15 10:44:09] DEBUG - 	d = 2
[2015-09-15 10:44:09] DEBUG - 	H = linde2
[2015-09-15 10:44:09] DEBUG - Rexfile exists
[2015-09-15 10:44:09] DEBUG - Checking Rexfile Syntax...
[2015-09-15 10:44:09] DEBUG - Creating lock-file (Rexfile.lock)
[2015-09-15 10:44:09] DEBUG - Including/Parsing Rexfile
[2015-09-15 10:44:09] DEBUG - Disabling usage of a tty
[2015-09-15 10:44:09] DEBUG - Activating autodie.
[2015-09-15 10:44:09] DEBUG - Using Net::OpenSSH if present.
[2015-09-15 10:44:09] DEBUG - Add service check.
[2015-09-15 10:44:09] DEBUG - Setting set() to not append data.
[2015-09-15 10:44:09] DEBUG - Registering CMDB as template variables.
[2015-09-15 10:44:09] DEBUG - activating featureset >= 0.51
[2015-09-15 10:44:09] DEBUG - activating featureset >= 0.40
[2015-09-15 10:44:09] DEBUG - activating featureset >= 0.35
[2015-09-15 10:44:09] DEBUG - activating featureset >= 0.31
[2015-09-15 10:44:09] DEBUG - Creating new distribution class of type: Base
[2015-09-15 10:44:09] DEBUG - new distribution class of type Rex::TaskList::Base created.
[2015-09-15 10:44:09] DEBUG - Registering task: Rex::CLI::uptime
[2015-09-15 10:44:09] DEBUG - Returning existing distribution class of type: Rex::TaskList::Base
[2015-09-15 10:44:09] DEBUG - Creating task: uptime
[2015-09-15 10:44:09] DEBUG - 	server: linde2
[...]

PR #793, opened 2015-09-15
Change the order for ssh username guesses, prefer current uid 

ipcjk:
Change the order for ssh username guesses, prefer current uid over getlogin

### DIFF
--- a/lib/Rex/Config.pm
+++ b/lib/Rex/Config.pm
@@ -441,7 +441,7 @@ sub get_user {
     return $user;
   }
 
-  return getlogin || getpwuid($<) || "Kilroy";
+  return getpwuid($<) || getlogin || "Kilroy";
 }
 
 sub get_password {


kablamo:
The original version was cut n paste from the `getlogin()` perldoc.  But I think this pr is more correct for what we are trying to do.  This change looks good to me.  :+1:   

Background: https://github.com/RexOps/Rex/issues/792

ferki:
@ipcjk: thanks for the patch!
@kablamo: thanks for the review! :)

I agree with the ordering. I think this also would make it possible to use "user-less" Rexfiles in some cases, making it easier to share them.

I have two questions:
 1. perldoc says: `Do not consider getlogin for authentication`. Do you think we can just drop `getlogin` completely?
 2. perlport says `getpwuid` is not implemented on Win32. Maybe we should only try to call it when running on non-Win32?

ipcjk:
getlogin shall be dropped because its confusing way to read utmp files. Also may not return what will be expected.

I think getpwuid(getuid()) may be the way to go on *nix. In general the question shall be how rexfiles are supposed to be called. If you can call it without specifying an username per host , it is a fallback feature to use getpwuid  and makes rexfiles or hostgroupfiles more tidier  for people that are using the same username all over for doing things.

ferki:
@ipcjk: are you still planning to work on this?

krimdomu:
the problem with changing `getpwuid` and `getlogin` is, that on MSWin systems it will fail with `The getpwuid function is unimplemented`. This is why the `getlogin` function is called first.
But if we make a distinction between MSWin and *nix we can rewrite it to something like:

```perl
if($^O =~ m/^MSWin/) {
  return getlogin;
}
else {
   return getpwuid($<);
}
```

ferki:
yup :+1: I think we can just commit that.

Issue #578, opened 2015-02-28
perl-5.8.8 compatibility


ferki:
## TL:DR version

We continue to support running Rex on CentOS/RHEL 5, but general usage with perl-5.8.8 is not supported (yet?).
## Long story

Since CentOS/RHEL 5 is supported by Rex, and those OSes are still shipping perl-5.8.8, we need to ensure that Rex can actually be _run_ on those systems (please note that this does not the same as the ability to remotely _manage_ those OSes).

Unfortunately 5.8.8 has a bug related to the `/m` regexp modifier. For more details, see the following upstream perl bugs:
- [RT#3038](https://rt.perl.org/Public/Bug/Display.html?id=3038)
- [RT#7781](https://rt.perl.org/Public/Bug/Display.html?id=7781)
- [RT#10054](https://rt.perl.org/Public/Bug/Display.html?id=10054)

For general 5.8.8 compatibility, all the current regular expressions that rely on the correct `/m` modifier behaviour needs to be replaced. Proposed [workaround from this Stack Overflow thread](http://stackoverflow.com/questions/11231441/regex-ms-not-behaving-as-expected-perl-5-8-8):
- `qr/^/m` is equivalent to `qr/(?:^|(?<=\n))/`
- `qr/$/m` is equivalent to `qr/(?=\n|\z)/`

Fortunately CentOS/RHEL 5 is not affected by this bug because they backported the upstream fix into `perl-5.8.8-6` (see [Changelog](http://rpm.pbone.net/index.php3/stat/22/idpl/27814152/com/changelog.html)).

ferki:
With the "End of Maintenance Support 2" of CentOS/RHEL 5 on 2017-03-31, we already bumped minimum supported perl version for _running_ Rex to 5.10.1.

The related commit is c4e35d6 as part of #1222.

Issue #340, opened 2014-04-10
Support for Authentication in http downloads


dirkcjelli:
It would be great if authentication info could be passed on to Download.pm to facilitate authenticated http downloads.

task "download", sub {
    download "username","password","http://www.rexify.org/index.html", "localfile.html";
 };

or similar.

PR #795, opened 2015-09-16
 Add support for ip6tables

kopytov:
Hello!

I've tried to add support for IPv6 in Rex::Commands::Iptables functions. Version of IP can be specified in the first argument of any function: -4 or -6 (like popular parameters in network tools). If no version specified, -4 implies (for full backward compatibility) .

Examples:

    iptables_clear -6;

    open_port -6, [22, 80];
    close_port -6, "all";
    redirect_port -6, 80 => 10080;
    default_state_rule -6;
    is_nat_gateway -6;

    iptables -6, "flush";
    iptables -6,
          t     => "filter",
          A     => "INPUT",
          i     => "eth0",
          m     => "state",
          state => "RELATED,ESTABLISHED",
          j     => "ACCEPT";

Please review this patch.

### DIFF
--- a/lib/Rex/Commands/Iptables.pm
+++ b/lib/Rex/Commands/Iptables.pm
@@ -83,6 +83,7 @@ package Rex::Commands::Iptables;
 
 use strict;
 use warnings;
+use version;
 
 # VERSION
 
@@ -96,6 +97,7 @@ use vars qw(@EXPORT);
 use Rex::Commands::Sysctl;
 use Rex::Commands::Run;
 use Rex::Commands::Gather;
+use Rex::Helper::Run;
 
 use Rex::Logger;
 
@@ -205,8 +207,8 @@ sub redirect_port {
   my @params     = @_;
   my $ip_version = _get_ip_version( \@params );
   if ( $ip_version == -6 ) {
-    my $iptables_version = _iptables_version();
-    if ( $iptables_version < 1_004_018 ) {
+    my $iptables_version = _iptables_version($ip_version);
+    if ( $iptables_version < v1.4.18 ) {
       Rex::Logger::info("iptables < v1.4.18 doesn't support NAT for IPv6");
       die("iptables < v1.4.18 doesn't support NAT for IPv6");
     }
@@ -371,7 +373,7 @@ sub is_nat_gateway {
 
     if ( $ip_version == -6 ) {
       die "NAT for IPv6 supported by iptables >= v1.4.18"
-        if _iptables_version() < 1_004_018;
+        if _iptables_version($ip_version) < v1.4.18;
       sysctl "net.ipv6.conf.all.forwarding",     1;
       sysctl "net.ipv6.conf.default.forwarding", 1;
       iptables $ip_version,
@@ -502,7 +504,7 @@ sub iptables_clear {
     next
       if $table eq "nat"
       && $ip_version == -6
-      && _iptables_version() < 1_004_018;
+      && _iptables_version($ip_version) < v1.4.18;
 
     iptables $ip_version, t => $table, F => '';
     iptables $ip_version, t => $table, X => '';
@@ -620,10 +622,15 @@ sub _get_executable {
 }
 
 sub _iptables_version {
-  my $version = join "", map { sprintf "%03d", $_ } split /[.]/,
-    substr run("iptables -V"), 10;
-  $version =~ s/^0+//;
-  return $version;
+  my (@params) = @_;
+  my $iptables = _get_executable( \@params );
+  my $out      = i_run("$iptables -V");
+  if ( $out =~ /v([.\d]+)/ms ) {
+    return version->parse($1);
+  }
+  else {
+    die "Can't parse `$iptables -V' output `$out'";
+  }
 }
 
 1;
--- a/lib/Rex/Commands/Iptables.pm
+++ b/lib/Rex/Commands/Iptables.pm
@@ -55,7 +55,25 @@ Only I<open_port> and I<close_port> are idempotent.
          accept          => "POSTROUTING",
          "out-interface" => "eth0",
          jump            => "MASQUERADE";
- 
+
+   # Version of IP can be specified in the first argument
+   # of any function: -4 or -6 (if no version specified,
+   # -4 implies)
+   iptables_clear -6;
+
+   open_port -6, [22, 80];
+   close_port -6, "all";
+   redirect_port -6, 80 => 10080;
+   default_state_rule -6;
+
+   iptables -6, "flush";
+   iptables -6,
+         t     => "filter",
+         A     => "INPUT",
+         i     => "eth0",
+         m     => "state",
+         state => "RELATED,ESTABLISHED",
+         j     => "ACCEPT";
  };
 
 =head1 EXPORTED FUNCTIONS
@@ -109,8 +127,9 @@ Open a port for inbound connections.
 =cut
 
 sub open_port {
-
-  my ( $port, $option ) = @_;
+  my (@params) = @_;
+  my $ip_version = _get_ip_version( \@params );
+  my ( $port, $option ) = @params;
 
   my %option_h;
   if ( ref $option ne "HASH" ) {
@@ -126,7 +145,8 @@ sub open_port {
     delete $option_h{only_if};
     $option = {%option_h};
   }
-  _open_or_close_port( "i", "I", "INPUT", "ACCEPT", $port, $option );
+  _open_or_close_port( $ip_version, "i", "I", "INPUT", "ACCEPT", $port,
+    $option );
 
 }
 
@@ -145,8 +165,9 @@ Close a port for inbound connections.
 =cut
 
 sub close_port {
-
-  my ( $port, $option ) = @_;
+  my (@params) = @_;
+  my $ip_version = _get_ip_version( \@params );
+  my ( $port, $option ) = @params;
 
   my %option_h;
   if ( ref $option ne "HASH" ) {
@@ -163,7 +184,7 @@ sub close_port {
     $option = {%option_h};
   }
 
-  _open_or_close_port( "i", "A", "INPUT", "DROP", $port, $option );
+  _open_or_close_port( $ip_version, "i", "A", "INPUT", "DROP", $port, $option );
 
 }
 
@@ -182,7 +203,9 @@ Redirect $in_port to another local port.
 =cut
 
 sub redirect_port {
-  my ( $in_port, $option ) = @_;
+  my (@params) = @_;
+  my $ip_version = _get_ip_version( \@params );
+  my ( $in_port, $option ) = @params;
 
   my @opts;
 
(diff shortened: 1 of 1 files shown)


ferki:
@kopytov: thanks for the PR! Unfortunately it seems to fail on perl versions < 5.13.2 because of line 631. I'll leave review notes as line comments.

lib/Rex/Commands/Iptables.pm
```diff
@@ -548,4 +610,25 @@ sub _rule_exists {
...
+
+sub _get_executable {
+  my ($params) = @_;
+  return _get_ip_version($params) == -6 ? "ip6tables" : "iptables";
+}
+
+sub _iptables_version {
+  my $version = join "", map { sprintf "%03d", $_ } split /[.]/,
+    substr run("iptables -V"), 10;
+  return $version =~ s/^0+//r;
```
ferki: This syntax looks like introduced in perl-5.13.2. Can you modify it to run on perl-5.8.9?

You can find more info on perl compatibility goals on #578. You can also try to catch those errors early by running `dzil test --no-author --release` on your machine (author tests are being fixed, but right now you can skip perlcritic warnings woth `--no-author`).

lib/Rex/Commands/Iptables.pm
```diff
@@ -55,7 +55,25 @@ Only I<open_port> and I<close_port> are idempotent.
          accept          => "POSTROUTING",
          "out-interface" => "eth0",
          jump            => "MASQUERADE";
- 
+
+   # Version of IP can be specified in the first argument
+   # of any function: -4 or -6 (if no version specified,
+   # -4 implies)
```
ferki: Kudos for the documentation bits! I think `defaults to -4` might be slightly better wording here.

lib/Rex/Commands/Iptables.pm
```diff
@@ -109,12 +127,13 @@ Open a port for inbound connections.
 =cut
 
 sub open_port {
-
-  my ( $port, $option ) = @_;
+  my (@params) = @_;
```
ferki: Is the parentheses around `@params` needed for a special reason here?

lib/Rex/Commands/Iptables.pm
```diff
@@ -182,8 +203,17 @@ Redirect $in_port to another local port.
...
 
 sub redirect_port {
-  my ( $in_port, $option ) = @_;
+  my (@params) = @_;
+  my $ip_version = _get_ip_version( \@params );
+  if ( $ip_version == -6 ) {
+    my $iptables_version = _iptables_version();
+    if ( $iptables_version < 1_004_018 ) {
+      Rex::Logger::info("IPtables < v1.4.18 doesn't support NAT for IPv6");
+      die("IPtables < v1.4.18 doesn't support NAT for IPv6");
```
ferki: I think the official name is `iptables` (vs. `IPtables`), based on a quick glance at their [homepage](http://www.netfilter.org/projects/iptables/). If you agree, can you update the messages here, please?

lib/Rex/Commands/Iptables.pm
```diff
@@ -338,13 +372,28 @@ sub is_nat_gateway {
...
+    my ($default_line) = run "/sbin/ip $ip_version r |grep ^default";
     my ($dev)          = ( $default_line =~ m/dev ([a-z0-9]+)/i );
     Rex::Logger::debug("Default GW Device is $dev");
 
-    sysctl "net.ipv4.ip_forward" => 1;
-    iptables t => "nat", A => "POSTROUTING", o => $dev, j => "MASQUERADE";
-
+    if ( $ip_version == -6 ) {
+      sysctl "net.ipv6.conf.all.forwarding",     1;
+      sysctl "net.ipv6.conf.default.forwarding", 1;
```
ferki: Would it make sense to wrap these `sysctl` commands into the block after the upcoming check for iptables version on the next line (i.e. only execute them if iptables actually supports IPv6 NAT)?

lib/Rex/Commands/Iptables.pm
```diff
@@ -338,13 +372,28 @@ sub is_nat_gateway {
...
+      if ( _iptables_version() < 1_004_018 ) {
+        Rex::Logger::info("NAT for IPv6 supported by iptables >= v1.4.18");
```
ferki: Is it OK to continue in this case without applying the command in the `else` branch? Why we don't need a `die` in this case?

lib/Rex/Commands/Iptables.pm
```diff
@@ -449,20 +503,28 @@ Remove all iptables rules.
...
+  my (@params) = @_;
+  my $ip_version = _get_ip_version( \@params );
 
   for my $table (qw/nat mangle filter/) {
-    iptables t => $table, F => '';
-    iptables t => $table, X => '';
+
+    # NAT for IPv6 supported by iptables >= v1.4.18
+    if ( $table eq "nat" && $ip_version == -6 ) {
+      next if _iptables_version() < 1_004_018;
```
ferki: These three conditions might be unified intoa single `next if A && B && C`. What do you think?

lib/Rex/Commands/Iptables.pm
```diff
@@ -548,4 +610,25 @@ sub _rule_exists {
...
+    if ( $params->[0] eq "-4" || $params->[0] eq "-6" ) {
+      return shift @$params;
+    }
+  }
+  return -4;
+}
+
+sub _get_executable {
+  my ($params) = @_;
+  return _get_ip_version($params) == -6 ? "ip6tables" : "iptables";
```
ferki: I think we can extend this subroutine to return the full path of the executable, instad of just a name. Like:
```perl
my $binary = _get_ip_version($params) == -6 ? "ip6tables" : "iptables";
return can_run($binary);
```

This way the check in the `iptables` subroutine (line #340) might also be simplified. I'll leave a note there too.

lib/Rex/Commands/Iptables.pm
```diff
@@ -306,8 +337,8 @@ sub iptables {
     }
   }
 
-  if ( can_run("iptables") ) {
-    my $output = run "iptables $cmd";
+  if ( can_run($iptables) ) {
```
ferki: If `_get_executable` returns the full path to the executable (commented below), this check might be simpified or pushed into `_get_executable` fully (also the `else` branch).

lib/Rex/Commands/Iptables.pm
```diff
@@ -548,4 +610,25 @@ sub _rule_exists {
...
+}
+
+sub _get_executable {
+  my ($params) = @_;
+  return _get_ip_version($params) == -6 ? "ip6tables" : "iptables";
+}
+
+sub _iptables_version {
+  my $version = join "", map { sprintf "%03d", $_ } split /[.]/,
+    substr run("iptables -V"), 10;
```
ferki: If you need to `run` something for internal use, please use `i_run` for that.

Do you think we can simplify this version retrieval? It looks correct, but it also seems like doing a lot of hoops just to retrieve, parse and normalize the version string (which can be a PITA, I know..).

Do we need to check for `iptables` or `ip6tables` version (i.e. calling `_get_executable` first to see what needs checking) depending on which mode is requested?

kopytov:
I've fixed compatibility problem and now tests successfully passs on my CentOS 5 box with perl 5.8.8. But something still goes wrong on Jenkins.

lib/Rex/Commands/Iptables.pm
kopytov: Sorry for my poor English. Fixed.

lib/Rex/Commands/Iptables.pm
kopytov: No, I just met this code:

    sub _iptables_list {
       my ( %tables, $ret );
       my (@lines) = @_; 

And supposed that this is author's code style. Also fixed.

ferki:
Thanks for the updates! During my second look, I noticed there's no tests yet for IPv6 support. Otherwise it's shaping up nicely for inclusion! :+1:

I think it is currently failing, because master branch had some accidentally committed non-tidy code. This has been fixed since then, so I think if you rebase on top of current master, this branch should also pass.

kopytov:
I've tried to make code of _iptables_version more simpler and stabler by using core's version parser & regex for parsing ip6?tables -V. Please review.

Function ``_get_executable`` (earlier ``iptables``) runs ``can_run($iptables)`` every time will call it. We can get some optimization if Rex will cache iptables path (and may be an iptables version). Something like this:

```perl
my %CACHE;

sub _get_executable {
  my ($params) = @_;
  my $hostname = connection->{server}->{name};
  $CACHE{executable}{$hostname}{$ip_version} ||= do {
    my $binary = _get_ip_version($params) == -6 ? "ip6tables" : "iptables";
    my $hostname = connection->{server}->{name};
    my $executable = can_run($binary);
    die "Can't find $binary in PATH" if $executable eq '';
    $executable;
  };
  return $CACHE{executable}{$hostname}{$ip_version};
}
```
What do you think about it?

ferki:
@kopytov: on one hand, and long term, I was thinking of reusing some CPAN module to do the iptables rule management (instead of duplicating efforts), and just make Rex DSL a wrapper around that (at least for backwards compatibility for now). One such module is for example [`IPTables::ChainMgr`](https://metacpan.org/pod/IPTables::ChainMgr).

On the other hand, if you'd like to use some caching, Rex already has `Rex::Interface::Cache` (and there's a parallel plan to improve that by reusing, again, appropriate caching modules from CPAN).

krimdomu:
@kopytov 
To use the internal cache this example code might help you:

```perl
my $cache = Rex::get_cache();
if ( $cache->valid("cache_key_name") ) {
  return $cache->get("cache_key_name");
}
$cache->set( "cache_key_name", $data );
```

@ferki i think using the module you mentioned is a good idea. But i think we should merge this pr and doing a separate refactoring to use the module.

ferki:
@krimdomu: yes, of course, switching to `IPTables::ChainMgr` is outside the scope of this PR.

kopytov:
Hello guys. I'm sorry for delay with it. Please review my last commit which enables cache. I've tested it on my default set of rules (about 100 rules), it shows good result - about 30 % faster.

lib/Rex/Interface/Connection/OpenSSH.pm
```diff
@@ -82,7 +82,10 @@ sub connect {
   $ssh_opts{ConnectTimeout} = $timeout;
 
   my %net_openssh_constructor_options = (
-    exists $ssh_opts{initialize_options} ? $ssh_opts{initialize_options} : () );
+    exists $ssh_opts{initialize_options} ? %{ $ssh_opts{initialize_options} } : () );
```
ferki: I'm not sure how this ended up as part of this PR - can you rebase your branch on top of current master, and push again, please?

kopytov:
@ferki done — rebased on fresh master.

ferki:
I'll try to squash the commits upon merge into logical changesets and/or rewrite commit messages. E.g. I don't see no reason putting the documentation into the commit message body.

@kopytov: if you can beat me to it, feel free to do the pre-merge cleanup ;)

ferki:
@kopytov: thanks, this is now merged manually after squashing the commits together!

PR #797, opened 2015-09-27
Fatal exceptions if git switching fail

alex1line:
There was a silent checkout even if the tag does not exists. Now throws the exception "ERROR - Error switching to tag."


krimdomu:
Hi @alex1line 
thanks for this patch. It looks good and i'm going to merge it.

ferki:
Shouldn't we increase the logging severity level to `error` from `warn`, if Rex is going to `die` two lines later in the same block?

PR #798, opened 2015-09-30
this prevents a warning message if Rex::Commands is loaded with require

krimdomu:


ferki:
@krimdomu: can you write an example for this issue please? I'm not entirely sure what case we're trying to fix here., but I'd like to merge it soon.

krimdomu:
@ferki i just updated the title. You'll see this the problem if you run
```
perl -e 'require Rex::Commands'
```

This is because $caller_pkg is not initialized.

ferki:
Thanks for the explanation, I get it now!

PR #800, opened 2015-10-02
Rex arch linux provider

stingA0815:
Hey folks,

I've tried to add ArchLinux support. Are there any objections or suggestions for improvement?

kind regards


stingA0815:
you can get some base arch Boxes here http://boxes.stinga.org/
an quick and dirty test env here https://github.com/stingA0815/rex-arch-test

krimdomu:
Hi @stingA0815 

thanks much for this patch. I think this looks good. 
It would be nice if you can squash these commits together to one. When you do a force push into your branch this pull requests gets updated automatically (as far as i know)

stingA0815:
done :-)

lib/Rex/Service.pm
```diff
@@ -61,6 +61,8 @@ sub get {
 
     # this also counts for Ubuntu
     $class = "Rex::Service::Debian::systemd";
+  }elsif ( is_archlinux($operatingsystem) ) {
```
ferki: Watch out on merge! I'm pretty sure this line is not tidy. I'm not sure yet why `perltidy` and/or the tidy tests missed this one.

lib/Rex/Pkg/ArchLinux.pm
```diff
@@ -0,0 +1,100 @@
...
+        arch    => $tmp_arch,
+      }
+    );
+  }
+
+  return @pkgs;
+}
+
+sub add_repository {
+  #no suitable repo management use template/file for pacman.conf
```
ferki: Would it make sense to print that message out as a warning, instead of only having it as a code comment, and silently returning success? Same question applies below in `rm_repository` too.

lib/Rex/Pkg/ArchLinux.pm
```diff
@@ -0,0 +1,100 @@
...
+  my $pkg_query = 'pacman -Qq';
+  if ($pkg) {
+    $pkg_query .= " " . $pkg;
+  }
+  my $pkg_query_details = "pacman -Qi";
+
+  my @installed_packages = i_run $pkg_query;
+  for my $package (@installed_packages) {
+    my @lines = i_run $pkg_query_details;
+    my ($tmp_name, $tmp_version, $tmp_release, $tmp_arch);
```
ferki: What does the `tmp_` prefix means here?

lib/Rex/Pkg/ArchLinux.pm
```diff
@@ -0,0 +1,100 @@
...
+  my @pkgs;
+  my $pkg_query = 'pacman -Qq';
+  if ($pkg) {
+    $pkg_query .= " " . $pkg;
+  }
+  my $pkg_query_details = "pacman -Qi";
+
+  my @installed_packages = i_run $pkg_query;
+  for my $package (@installed_packages) {
+    my @lines = i_run $pkg_query_details;
```
ferki: If I understand it correctly, here we get the full list of packages first, then trying to iterate through all of them, and query detailed info for each package in the inner loop.

Are there a simpler way to list detailed package information, than doing it in a nested loop?

Also, I don't understand how `my @lines = i_run $pkg_query_details;` knows which package to query for more information without passing `$package` to it. Is it maybe just missing from the end of the command?

lib/Rex/Pkg/ArchLinux.pm
```diff
@@ -0,0 +1,100 @@
...
+      if ( $line =~ m/^Name\s+\:\s+([^\s]+)$/ ) {
+        $tmp_name = $1;
+      }
+      if ( $line =~ m/^Version\s+\:\s+([^\s]+)-([^\s]+)$/ ) {
+        $tmp_version = $1;
+        $tmp_release = $2;
+      }
+      if ( $line =~ m/^Architecture\s+\:\s+([^\s]+)$/ ) {
+        $tmp_arch = $1;
+      }
```
ferki: Would it make sense to just restrict, or prefilter the output of `pacman -Qi` for only these information that we want to parse out later? Like `pacman -Qi | grep -e '^Name' -e '^Version' -e '^Architecture'`. I think that would be easier to work with.

lib/Rex/Pkg/ArchLinux.pm
```diff
@@ -0,0 +1,100 @@
...
+    $pkg_query .= " " . $pkg;
+  }
+  my $pkg_query_details = "pacman -Qi";
+
+  my @installed_packages = i_run $pkg_query;
+  for my $package (@installed_packages) {
+    my @lines = i_run $pkg_query_details;
+    my ($tmp_name, $tmp_version, $tmp_release, $tmp_arch);
+    for my $line (@lines) {
+      if ( $line =~ m/^Name\s+\:\s+([^\s]+)$/ ) {
```
ferki: I think `:` does not need to be escaped inside regeular expressions, and `[^\s]` could be written simply as `\S`.

ferki:
@stingA0815: nice, thanks! I left some notes and questions directly on the commit.

@krimdomu: we can add an Arch box to our jenkins tests from @stingA0815's site: http://boxes.stinga.org

stingA0815:
updated

ferki:
@stingA0815: thanks for the updates! I'll go over the PR again, and also run functional tests on our [build server](https://build.rexify.org) before merging.

stingA0815:
tried to update ...

ferki:
@krimdomu @stingA0815: I've added [Arch tests to Jenkins](https://build.rexify.org/job/Test%20-%20Arch%2020151001%20(amd64)/) (thanks for the image!). Unfortunately they don't pass, neither for service or package management for this branch.

This branch also can't be merged cleanly into master, so it needs rebase on top of current master and conflict resolution.

ferki:
@stingA0815: I've merged and squashed your changes into our repository's `arch` branch, resolving the merge conflicts, and doing minor fixes (like removing unrelated trailing spaces on a separate commit, and adding missing `Rex::Pkg::Arch` to pp.args).

I'll try to make it pass the arch VM tests.

stingA0815:
Hey @ferki ,

ty! I'm not realy familar with that squash stuff. Is there anything i schould do?

kind regards
Harm

ferki:
@stingA0815: git squashing is just "flattening" multiple commits into one (e.g. instead of keeping the history for all the experimental commits that lead to a solution, we can squash them together in order to keep a more readable git history). You don't really need to do anything about that.

If you have ideas about why package installation and service management still fails with the patches so far, feel free to send a second PR against the `arch` branch. But now we have an arch test VM, I guess we can track that down on our own too from here.

I think I'll keep this PR open till we resolve this.

stingA0815:
Hey @ferki , ich checked my little test env. Seems 2 packages where missing.

perl-devel-caller
perl-padwalker

kind regards

krimdomu:
Hi @stingA0815 

thank you very much for your patch. I've run perltidy over the code, squashed the commits and merged it into master branch.

Issue #623, opened 2015-04-07
Passing args to tasks is awkward


kablamo:
I'd like to limit the number of tasks Rex can do at a time to one.  That way all arguments after the task name can be passed to the task.  For example:
```
rex Unix:ls -l /etc
```
would result in `-l /etc' being passed to the task Unix::ls() in some way.  The only way to do this currently is to do something like:
```
rex Unix:ls --l --dir /etc
```
which is really awkward.

How would you feel about a pull request that did something like this.  I realize there are a lot of details that need to be worked out here.

krimdomu:
Hi @kablamo 

if this is possible without changing the current behaviour we welcome any patches. 
There is already a task on the backlog (without a ticket yet) to refactor the argument parsing. Maybe we can do this together in one step.
Maybe there is a nice CPAN module that can do this?

kablamo:
Currently Rex (sort of) allows you to do more than one task at a time.  So for example:
```
rex -G group task1 task2
```
task1 runs for each server in "group".  However task2 seems to get executed on the local machine only.  

I want to change that to only allow Rex to do one task at a time.  So in the above example, the string "task2" would get passed as an argument to task1 (in some backwards compatible way) and task2 would not run.  

Does that count as changing current behavior?  <-- The part about only doing one task at a time, I mean.

kablamo:
Hmm after reading some more of the documentation I think that would count as breaking backwards compatibility.

kablamo:
I'm interested in getting involved in refactoring argument parsing.  What specific goals did you have in mind?

One way to break backwards compatibility without affecting any users would be to create a new command line util with a different name.  Or maybe just something that wraps the 'rex' script.

krimdomu:
Well, 

i think we can add a new feature flag for example `1.3` (if the new arg parser is part of version 1.3) which will enable this new arg parser. So there is no problem for people who are using an older feature flag. This would be a little bit of a hack, because currently the arguments are parsed before the Rexfile is parsed. So we would ne to parse the Rexfile twice. The first time to check which feature flag is enabled, then parse the arguments and then load the Rexfile for real.

Another option would be to try to autodetect which parser should be used (i would vote for this one).

For example if we define that we wan't to have a more generic cli interface:

```
Syntax:
rex $action @options
```

To execute a task
```
rex execute $task
rex execute $task -u user -p password --foo=bar --action=fun
```

To query modules.rexify.org
```
rex search apache
```

[...]

PR #801, opened 2015-10-03
added plugable interface - #623

krimdomu:
This is a pull request to allow the creation of new cli interfaces as discussed in #623 
This code tries to detect with `PPI` which cli interface it should load (by the help of a feature flag `cli_XYZ`)

So with this code it would be possible to experiment with new command line interfaces without compromising the current code.


bin/rex
```diff
@@ -11,7 +11,14 @@ use warnings;
 
 # VERSION
 
-use Rex::CLI;
+use Rex::DetectCLI;
+my $cli = detect_cli;
+
+eval "require 'Rex/CLI/$cli.pm'";
```
kablamo: https://metacpan.org/pod/Module::Load?   Just a suggestion

lib/Rex/DetectCLI.pm
```diff
@@ -0,0 +1,89 @@
...
+use vars qw(@EXPORT);
+
+use List::MoreUtils qw'firstidx';
+use Data::Dumper;
+
+use PPI;
+
+@EXPORT = qw(detect_cli);
+
+sub detect_cli {
```
kablamo: This is pretty cool.  I mean its also insane.  I don't love it.  But I don't have any better ideas atm.  I thought PPI would be really slow.  Its not as bad as I thought though...

krimdomu: yes, i know, i also feel not that good with it but it seems to work pretty good.

bin/rex
krimdomu: you're right. and there is also: https://github.com/RexOps/Rex/blob/master/lib/Rex/Require.pm which is doing the same.

krimdomu:
This can be closed in favor of Rex 2.0 which will implement a new cli

PR #802, opened 2015-10-03
Fix managing a user's own crontab via cron_entry

ferki:
`cron_entry` fails without passing a user to it (correctly), however this enforces calling `crontab` CLI command on the remotes with the `-u` option, which in turn requires superuser privileges.

This patch resolves that situation, by checking if the effective remote username is the same as the username for which the crontab operation is requested. If yes, it means the user manages his/her own crontab, so Rex avoids using the `-u` CLI option.

### DIFF
--- a/lib/Rex/Cron/Base.pm
+++ b/lib/Rex/Cron/Base.pm
@@ -174,6 +174,7 @@ sub read_user_cron {
 
   my $command = 'crontab -l';
   $command .= " -u $user" if defined $user;
+  $command .= ' 2> /dev/null';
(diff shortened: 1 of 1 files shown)


ferki:
I'd like to push one more cron-related commit here about fixing crontab comments multiplication.

ferki:
Ok, that's not a rex issue, but how `crontab` command with `vixie-cron` behaves (at least on gentoo).

krimdomu:
from just reading the code: looks good

PR #803, opened 2015-10-13
Fix undef warnings if calling get() without param

abeverley:
When using the firewall module, lots of ```Use of uninitialized value $var in exists at /usr/local/share/perl/5.18.2/Rex/Config.pm line 870``` warnings are generated. The offending code is here: https://github.com/RexOps/rex-recipes/blob/1.3/Rex/Ext/ParamLookup/__module__.pm#L41 (```Rex::CMDB::cmdb($cmdb_key)``` is often undefined).

### DIFF
--- a/lib/Rex/Config.pm
+++ b/lib/Rex/Config.pm
@@ -865,6 +865,7 @@ sub unset {
 
 sub get {
   my ( $class, $var ) = @_;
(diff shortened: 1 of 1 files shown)


krimdomu:
hi @abeverley 
thanks for this patch.

PR #804, opened 2015-10-13
added ParamLookup, also for RexOps/Rex#545

krimdomu:
This pr adds the Rex::Ext::ParamLookup module, because this module is the prefered way to handle task parameters.

### DIFF
--- a/lib/Rex/Helper/Rexfile/ParamLookup.pm
+++ b/lib/Rex/Helper/Rexfile/ParamLookup.pm
@@ -69,8 +69,7 @@ sub param_lookup {
     $ret = $default;
   }
 
-  # make the variable also known to templates
-  Rex::Commands::set( $key => $ret );
+  Rex::Commands::task()->set_parameter( $key => $ret );
 
   return $ret;
 }
@@ -90,7 +89,7 @@ This module also looks inside a CMDB (if present) for a valid key.
 
 =head1 SYNOPSIS
 
- task "setup", make {
+ task "setup", sub {
    my $var = param_lookup "param_name", "default_value";
  };
 
--- a/lib/Rex/Helper/Rexfile/ParamLookup.pm
+++ b/lib/Rex/Helper/Rexfile/ParamLookup.pm
@@ -1,0 +1,103 @@
+#
+# (c) Jan Gehring <jan.gehring@gmail.com>
+#
+# vim: set ts=3 sw=3 tw=0:
+# vim: set expandtab:
+
+package Rex::Helper::Rexfile::ParamLookup;
+
+use strict;
+use warnings;
+
+# VERSION
+
+use Devel::Caller;
+use Data::Dumper;
+require Rex::Exporter;
+require Rex::Commands;
+
+use base qw(Rex::Exporter);
+use vars qw (@EXPORT);
+
+@EXPORT = qw(param_lookup);
+
+sub param_lookup {
+  my ( $key, $default ) = @_;
+
+  my $ret;
+
+  my ($caller_pkg) = caller(0);
+
+  my @args = Devel::Caller::caller_args(1);
+  if ( ref $args[0] eq "HASH" ) {
+    if ( exists $args[0]->{$key} ) {
+      $ret = $args[0]->{$key};
+    }
+  }
+
+  if ( !$ret ) {
+
+    # check if cmdb is loaded
+    my ($use_cmdb) = grep { m/CMDB\.pm/ } keys %INC;
+    if ($use_cmdb) {
+
+      # look inside cmdb
+      my $cmdb_key = "${caller_pkg}::$key";
+      $ret = Rex::Commands::get( Rex::CMDB::cmdb($cmdb_key) );
+
+      if ( !$ret ) {
+
+        # check in resource
+        if ($Rex::Resource::INSIDE_RES) {
+          $cmdb_key =
+              "${caller_pkg}::"
+            . $Rex::Resource::CURRENT_RES[-1]->display_name
+            . "::$key";
+          $ret = Rex::Commands::get( Rex::CMDB::cmdb($cmdb_key) );
+        }
+      }
+
+      if ( !$ret ) {
+
+        # check in global namespace
+        $ret = Rex::Commands::get( Rex::CMDB::cmdb($key) );
+      }
+    }
+  }
+
+  if ( !$ret ) {
+    $ret = $default;
+  }
+
+  # make the variable also known to templates
+  Rex::Commands::set( $key => $ret );
+
+  return $ret;
+}
+
+1;
+
+=pod
+
+=head1 NAME
+
+Rex::Helper::Rexfile::ParamLookup - A command to manage task parameters.
+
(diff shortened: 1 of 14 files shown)


krimdomu:
added tests

lib/Rex/Helper/Rexfile/ParamLookup.pm
```diff
@@ -0,0 +1,103 @@
...
+Rex::Helper::Rexfile::ParamLookup - A command to manage task parameters.
+
+A command to manage task parameters. Additionally it register the parameters as template values.
+
+This module also looks inside a CMDB (if present) for a valid key.
+
+
+=head1 SYNOPSIS
+
+ task "setup", make {
```
kablamo: Maybe I'm missing something obvious... what is this 'make' keyword?

lib/Rex/Helper/Rexfile/ParamLookup.pm
```diff
@@ -0,0 +1,103 @@
...
+      }
+    }
+  }
+
+  if ( !$ret ) {
+    $ret = $default;
+  }
+
+  # make the variable also known to templates
+  Rex::Commands::set( $key => $ret );
```
kablamo: I assume this value persists across tasks.  What happens when you have more than one task?

lib/Rex/Helper/Rexfile/ParamLookup.pm
```diff
@@ -0,0 +1,103 @@
...
+@EXPORT = qw(param_lookup);
+
+sub param_lookup {
+  my ( $key, $default ) = @_;
+
+  my $ret;
+
+  my ($caller_pkg) = caller(0);
+
+  my @args = Devel::Caller::caller_args(1);
```
kablamo: Neat!  A bit magical.  Does it still work if the user does this in the task:
```
my $x = shift;
my $y = param_lookup 'key', 'default';
```

lib/Rex/Helper/Rexfile/ParamLookup.pm
```diff
@@ -0,0 +1,103 @@
...
+
+A command to manage task parameters. Additionally it register the parameters as template values.
+
+This module also looks inside a CMDB (if present) for a valid key.
+
+
+=head1 SYNOPSIS
+
+ task "setup", make {
+   my $var = param_lookup "param_name", "default_value";
```
kablamo: I wanted to add the ability for tasks to have args as well as --params.  Any thoughts on how that might work?  Could have an 'args_lookup'  I guess.

```
my @args = args_lookup();
```

lib/Rex/Helper/Rexfile/ParamLookup.pm
kablamo: I like this, but should it happen in the coderef/wrapper around the task that runs when the task is run?

lib/Rex/Helper/Rexfile/ParamLookup.pm
krimdomu: the `make` keyword is the same as `sub`. I'll need to update the docs and change it to sub.
i added `make` in the past for a colleague (kind of a joke), because he was totally in love with "make". And sometimes i use it instead of `sub`

lib/Rex/Helper/Rexfile/ParamLookup.pm
krimdomu: You are talking about line 73?

lib/Rex/Helper/Rexfile/ParamLookup.pm
krimdomu: i have to test it, but i think if someone `shift` it doesn't work anymore. This is mostly for users which don't know perl and it is a bit hard to teach them "\@_"

ferki:
Great \o/ I like to see it finding its way into core ^^ :+1: 

ps.: the test failure here is introduced by #803, which shifted an existing perlcritic warning by a line and jenkins is hypersensitive on that. I'll cleanup the test state later.

lib/Rex/Helper/Rexfile/ParamLookup.pm
kablamo: Oh weird.  I view people who like `make` with suspicion :eyes: .

lib/Rex/Helper/Rexfile/ParamLookup.pm
kablamo: Yes.  And the coderef/wrapper I'm referring to is ... I guess... here: https://github.com/RexOps/Rex/blob/master/lib/Rex/TaskList/Parallel_ForkManager.pm#L62

lib/Rex/Helper/Rexfile/ParamLookup.pm
kablamo: :+1:

lib/Rex/Helper/Rexfile/ParamLookup.pm
krimdomu: if i get it right, i don't think that we can add it there, because at this moment we doesn't know what the user is doing with the parameters.
the parameters only persist in subtasks, because tasks run inside a fork.

kablamo: Sorry if I'm being a pain!  This pr looks really cool.  But I'm thinking it would be nice if people could get their task params added to the template automatically without needing to use this module or do anything.

What is it that we know here on line 73 of ParamLookup.pm that we don't know on line 62 of Parallel_ForkManager.pm?

Also, to solve the problem of parameters persisting in subtasks perhaps you could do this:
```
Rex::Commands::set( task_params => $task_params_hashref );
```

krimdomu: no problem, keep writing :+1:  it is pretty cool to have people like you.

At this point we don't know which default values we had to set if the parameter is not given via command line or other possibilities.

But the idea of automatically adding known parameters to the template is good. and i'm going to open another task to do this.

I think the problem that the parameters persist in subtask must be solved in another way. Currently we have a task object, and we can add those parameters to this task object. So with every new task we also do have a new object which then is clean.

I'm going to take a look at this.

krimdomu: i'm not going to open a new ticket, because this can be done in companion with the parameter persistance thing

lib/Rex/Helper/Rexfile/ParamLookup.pm
```diff
@@ -0,0 +1,102 @@
...
+        $ret = Rex::Commands::get( Rex::CMDB::cmdb($key) );
+      }
+    }
+  }
+
+  if ( !$ret ) {
+    $ret = $default;
+  }
+
+  Rex::Commands::task()->set_parameter( $key => $ret );
```
kablamo: @krimdomu I don't get what this does.  I thought `task0` will create a new task?

kablamo: Oh I see it now.  `task()` with no params returns the task from the current connection.

PR #805, opened 2015-10-13
Feature firewall

krimdomu:
This pr adds the Firewall module from rex-recipes to core.

This is the first module which will use the `resource` syntax which will automatically add `on_change` options and reporting.

this pr has a dependency to #804

### DIFF
--- a/lib/Rex.pm
+++ b/lib/Rex.pm
@@ -529,6 +529,9 @@ sub import {
 
     require Rex::Helper::Rexfile::ParamLookup;
     Rex::Helper::Rexfile::ParamLookup->import( register_in => $register_to );
+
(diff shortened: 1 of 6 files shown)


krimdomu:
well, it doesn't have a dependency, it includes it.. *grml*
i'm going to cherry-pick this at the end

ferki:
Nice! :+1:

If I understand correctly, it requires ParamLookup, so that needs to be merged first (#804) I guess after that, we can rebase/rework this branch easily to have a clean test-merge cycle here too.

krimdomu:
@ferki right

abeverley:
Awesome, thanks very much!

PR #809, opened 2015-10-19
Pkg/Gentoo.pm: Avoid needless rebuilds

patricklauer:
Using emerge -u will install package if absent, update if needed, and noop if already the newest version.
This avoids rebuilding in the last case, where before it always re-built the package.


ferki:
@patricklauer: many thanks for this patch! I took the liberty to rebase your commits, and squash them together, slightly reword the commit message, and add you to the list of contributors :)

Issue #470, opened 2014-11-05
Refactor mkfs/mkswap usage


ferki:
Rex already tries to create filesystems with `mkfs`/`mkswap` (in Rex::Commands::LVM and in Rex::Commands::Partition), so the logic could be extracted to a single place. It would also make a dedicated `mkfs` command possible.

It could also be useful for example when the partition table is created without creating filesystems yet, e.g. because a RAID array needs to be created first.

protoCall7:
Extracted the mkfs logic from LVM.pm and Partitions.pm into Mkfs.pm.

PR #810, opened 2015-10-20
Mkfs refactor

protoCall7:
Extracted and cleaned up mkfs logic from LVM.pm and Partitions.pm per issue #470 

Any input would be greatly appreciated.  I have not had a chance to spin up a test machine to try mkfs() on yet, however 'dzil test --no-author' is passing all tests.

### DIFF
--- a/lib/Rex/Commands/Mkfs.pm
+++ b/lib/Rex/Commands/Mkfs.pm
@@ -45,32 +45,31 @@ Create a filesystem on device $devname.
 =cut
 
 sub mkfs {
-    my ( $devname, %option ) = @_;
+  my ( $devname, %option ) = @_;
 
-    if ( exists $option{fstype} ) {
-        if ( $option{fstype} eq "swap" ) {
-            Rex::Logger::info("Creating swap space on /dev/$devname");
-            run "mkswap -f /dev/$devname";
-        }
-        elsif ( can_run("mkfs.$option{fstype}") ) {
-            Rex::Logger::info(
-                "Creating filesystem $option{fstype} on /dev/$devname");
-
-            my $add_opts = "";
-
-            if ( exists $option{label} || exists $option{lable} ) {
-                my $label = $option{label} || $option{lable};
-                $add_opts .= " -L $label ";
-            }
-
-            run "mkfs.$option{fstype} $add_opts /dev/$devname";
-        }
+  if ( exists $option{fstype} ) {
+    if ( $option{fstype} eq "swap" ) {
+      Rex::Logger::info("Creating swap space on /dev/$devname");
+      run "mkswap -f /dev/$devname";
     }
-    else {
-        croak("Can't format partition with $option{fstype}");
-    }
+    elsif ( can_run("mkfs.$option{fstype}") ) {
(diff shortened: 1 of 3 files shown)


ferki:
Thanks for this! I only accidentally close-reopened it :) I'll do a proper review once I'm at my laptop!

protoCall7:
Thanks @ferki!

If you don't have a handy machine to do proper testing with, I will be spinning up a few VM's to test on this evening, and can report my results here.

lib/Rex/Commands/Mkfs.pm
```diff
@@ -0,0 +1,75 @@
...
+=cut
+
+sub mkfs {
+  my ( $devname, %option ) = @_;
+
+  if ( exists $option{fstype} ) {
+    if ( $option{fstype} eq "swap" ) {
+      Rex::Logger::info("Creating swap space on /dev/$devname");
+      run "mkswap -f /dev/$devname";
+    }
```
ferki: I think it's not possible anymore to specify a label for a swap partition. Can you make sure this possibility is not lost?

ferki:
@protoCall7: looks good, thanks! 

I'm not sure if the mkfs logic needs to have its own module, as it contains only a single function. I guess it can be kept in `Rex::Commands::Fs` (as in "that module contains filesystem related functions"), but it's not a very strong preference, just an idea. @krimdomu: do you have any preferences on this?

Apart from that, I left only one small comment about swap + labels. I think we can merge this once it is fixed + all commits squashed into one.

protoCall7:
Updated the mkswap portion to support -L as well.

lib/Rex/Commands/Partition.pm
```diff
@@ -225,42 +226,8 @@ sub partition {
...
-      $add_opts .= " -L $label ";
-    }
-
-    run "mkswap $add_opts /dev/$disk$part_num";
-  }
-  else {
-    die("Can't format partition with $option{fstype}");
-  }
-
+  mkfs "$disk$part_num";
```
ferki: How `Rex::Commands::Fs::mkfs` would know about the filesystem options (e.g. what file system type to create)? Would it still skip creating a file system if `non-fs`, `none` or `''` is passed as requested filetype?

ferki:
ping:

>I'm not sure if the mkfs logic needs to have its own module, as it contains only a single function. I guess it can be kept in Rex::Commands::Fs (as in "that module contains filesystem related functions"), but it's not a very strong preference, just an idea.

@krimdomu: do you have any preferences on this?

ps.: if we're going to put it into a new module, we need to record it in `bin/rexify` and `docs/pp.args` too.

krimdomu:
@ferki 
i think this is okay. Also it would be nice to have an "ensure" option, so that this function get idempotent. (checks partition with some command which fs it has, and if it is the requested one, do nothing)
But this can be done with a later code change.

krimdomu:
Hi @protoCall7 

Thank you for this patch. I've just merged it into master branch.

lib/Rex/Commands/Partition.pm
ferki: This PR looks like was merged for some reason without resolving this part of the problem, and lead us to #905 :(

PR #812, opened 2015-10-29
Fix GetRex link url

hiroraba:


ferki:
Thanks for spotting this, @hiroraba! I've also added you to the list of contributors ;)

PR #813, opened 2015-10-30
Updated pod for bin/rex

kablamo:
The goal was to update the command line options since that was out of date.  But then I got carried away and made a lot of other changes to the pod to improve readability on metacpan.


krimdomu:
Hi @kablamo 
thanks for this patch. I've just merged it into master and slightly updated the commit message.

PR #814, opened 2015-10-30
fix wrong detection for pid exist checking

petersonchen:
for example
there is a pid 1492 in system  
and the previous rex pid is 492 , without this patch , It will detect Rexfile is in use by 492.


krimdomu:
Hi @petersonchen 
thank you much for this patch. I've just merged it into master branch and slightly reworded the commit message.

PR #815, opened 2015-10-31
Switch from dzil [autoprereqs] to [PrereqsFromCPANfile]

kablamo:
# The problem

Rex complains Net::SSH2 is not installed when I run it.  This happened
 - because I didn't have Net::OpenSSH installed
 - which happened because `dzil listdeps` doesn't list Net::OpenSSH as dependency for me even though I'm on Ubuntu. 
 - which happened because `[OSPrereqs]` needs `[MakeMaker]` or `[ModuleBuild]` but instead we are using `[MakeMaker::Awesome]`.

# A solution

I'm sure there is more than one solution here.  I'm kind of going with what I know.  I'm a big fan of Carton and Minilla.  My solution is to use Miyagawa's `[Prereqs::FromCPANfile]` module to replace `[OSPrereqs]` and `[Autoprereqs]`.

**Pros**
 - `dzil listdeps` only takes 12 seconds to run instead of 32 seconds.
 - Contributors don't have to install dzil to develop rex.  They can optionally use Carton which is simpler to use and *way* faster to install.  I think its good to reduce barriers.  I've even heard of developers saying they won't contribute to projects which use dzil.  Which I think is kind of extreme, but I get where they are coming from.
 - cpanfile is more flexible and easier to work with than dist.ini.

**Cons**
 - We have to manually maintain dependencies in the cpanfile.  We could run [scan-prereqs-cpanfile](https://metacpan.org/pod/distribution/App-scan_prereqs_cpanfile/script/scan-prereqs-cpanfile) to automatically find dependencies, but there isn't a dzil plugin for it atm that I know of.  And if there was that would make things slower again.

### DIFF
--- a/cpanfile
+++ b/cpanfile
@@ -1,0 +1,62 @@
+requires 'attributes';
+requires 'base';
+requires 'Carp';
+requires 'constant';
+requires 'Cwd';
+requires 'Data::Dumper';
+requires 'Devel::Caller';
+requires 'Digest::MD5';
+requires 'Exporter';
+requires 'ExtUtils::MakeMaker';
+requires 'Fcntl';
+requires 'File::Basename';
+requires 'File::Spec';
+requires 'File::Spec::Unix';
+requires 'File::Spec::Win32';
+requires 'File::Temp';
+requires 'FindBin';
+requires 'Hash::Merge';
+requires 'HTTP::Request';
+requires 'HTTP::Request::Common';
+requires 'IO::File';
+requires 'IO::Socket';
+requires 'IO::String';
+requires 'IPC::Open3';
+requires 'JSON::XS';
+requires 'List::MoreUtils';
+requires 'List::Util';
+requires 'LWP::UserAgent';
+requires 'MIME::Base64';
+requires 'overload';
+requires 'POSIX';
+requires 'Sort::Naturally';
+requires 'Storable';
+requires 'strict';
+requires 'String::Escape';
+requires 'Symbol';
+requires 'Term::ReadKey';
+requires 'Test::Builder::Module';
+requires 'Test::Deep';
+requires 'Test::More';
+requires 'Test::Pod';
+requires 'Test::UseAllModules';
+requires 'Text::Glob';
+requires 'Text::Wrap';
+requires 'Time::HiRes';
+requires 'UNIVERSAL';
+requires 'URI';
+requires 'URI::QueryParam';
+requires 'vars';
+requires 'warnings';
+requires 'XML::LibXML';
+requires 'XML::Simple';
+requires 'YAML';
+
+if ($^O =~ /MSWin/) {
+    requires 'Net::SSH2';
+}
+else {
+    requires 'IO::Pty';
+    requires 'Net::OpenSSH';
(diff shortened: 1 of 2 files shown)


ferki:
I think `OSPrereqs` is compatible with `MakeMaker::Awesome` (the latter even states it's 100% compatible, + judging by the source it's just subclassing `MakeMaker`). Both of them are just writing to `Makefile.PL`. We use `OSPrereqs` to describe platform-specific dependencies.

As far as I can judge by its code, `dzil listdeps` only shows dependencies based on the project's codebase, i.e. it does not generate and evaluate the generated `Makefile.PL`, which has the following lines:
```perl
if ( $^O !~ /MSWin/i ) {
	$WriteMakefileArgs{PREREQ_PM}{'IO::Pty'} = '0';
	$WriteMakefileArgs{PREREQ_PM}{'Net::OpenSSH'} = '0';
	$WriteMakefileArgs{PREREQ_PM}{'Net::SFTP::Foreign'} = '0';
}

if ( $^O =~ /MSWin/i ) {
	$WriteMakefileArgs{PREREQ_PM}{'Net::SSH2'} = '0';
}
```

So they become install-time dependencies, instead of "development-time" dependencies.

All in all IMHO, it's not a problem with `AutoPrereqs`, `OSPrereqs` or `MakeMaker::Awesome`, but with how `dzil listdeps` works.

Back to your original problem:
>Rex complains Net::SSH2 is not installed when I run it

So I think his happens because rex wasn't installed via either a package manager (which should handle all deps for the given OS), or via `dzil install` (which builds the project, including the `Makefile.PL`, which gets evaluated install-time, and then takes care of the dependencies for the given OS).

That being said, in order to support `Carton` (or even `Carmel`), we _can_ provide a `cpanfile` as an alternative, provided we can describe platform-specific dependencies with it somehow.

ferki:
[sorry, early posting above]
So we can do platform-specific depdendencies with `cpanfile`, and only remaining bit is the autogeneration of it. I guess it would be nice if we won't need to commit it to the git repo, but generate it dynamically (to not accidentally forget including something).

ferki:
And part of the question is do we want slow autogeneration of dependencies, or fast generation, but manual maintenance"

Personally I'm leaning slightly towards the slow autogeneration. I appreciate the fact that I cannot really forget to include a dependency, and I think I need to generate the deps rarely anyways.

kablamo:
Nice debugging of the problem with OSPrereqs.  

I don't `dzil install` because I don't want 1000 dependencies in my $PERL5LIB which may conflict with other projects I'm working on.  Which is why I'm a fan of Carton.  

In terms of helping contributors by providing a cpanfile.  You and I don't need to generate dependencies often, but I think its good to have low barriers for first time contributors.  Obviously lots of people are comfortable with dzil at this point though.

There is a  [CPANFile] plugin to generate a cpanfile.  But it doesn't handle the ospreqs stuff so thats not really helpful.

I don't think there is a solution here which makes everyone happy.  Let's just go with the status quo.  There are bigger issues to solve.

ferki:
We can help (Dist::Zilla) upstream by extending their solutions to handle what we miss.

PR #819, opened 2015-11-01
Colorize and simplify summary

ferki:
This PR makes the summary output more obvious at the end of run by using error level messages if there were failures (making the output red).

Plus it simplifies the summary printing code, and uses a slightly modified wording.

@kablamo, @krimdomu: what do you think?

### DIFF
--- a/lib/Rex/CLI.pm
+++ b/lib/Rex/CLI.pm
@@ -18,6 +18,7 @@ use Cwd qw(getcwd);
 use List::Util qw(max);
 use Text::Wrap;
 use Term::ReadKey;
+use Sort::Naturally;
 
 use Rex;
 use Rex::Args;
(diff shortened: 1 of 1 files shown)


kablamo:
I don't feel strongly about it.  I guess I feel like the old output was slightly easier to read and understand at a glance.  But maybe thats just me.  I feel like the SUMMARY line made it stand out and told the user what it was.  But I see you are going for making it shorter which is a goal I definitely agree with.  

I kind of feel the longer version is nice for tasks with lots of output.  And the shorter one is nicer for quick one offs with very little output.

I disable the Rex::Logger and do all my own logging anyway so I will end up making it look however I want in the end.

ferki:
Using `error` level when printing failures would enable users to use `-qw` on the command line to hide normal out, but have a summary printerd out in case of failures.

kablamo:
Thats a good point.  If you have a good feeling, go for it.  I don't feel strongly about it.

PR #820, opened 2015-11-01
Read STDOUT and STDERR in parallel (fix #756)

### LINKED ISSUE #756: Enough stderr output in a run() call makes it hang forever when set_no_tty=1
I've found an issue that when a script outputs too much stderr, the call hangs forever under OpenSSH.
$ rex -v
(R)?ex 1.3.2

Here is my Rexfile:
```perl
use Rex -feature => '0.42';

desc <<'EOF';
Print X chars to stderr
  Options:
    --count=INTEGER
EOF
task "print_chars_to_stderr" => sub {
    my $opt = shift;
    my $count = $opt->{count} || 1;
    Rex::Logger::info( "connection isa: " . ref(connection) );
    Rex::Config->set_no_tty(1);
    my $cmd = qq{perl -e 'print STDERR "x" for (0..$count);'};
    Rex::Logger::info($cmd);
    my $out = run $cmd, sub {
        my ( $out, $err ) = @_;
        Rex::Logger::info( "stdout length: " . length($out) );
        Rex::Logger::info( "stderr length: " . length($err) );
    }, auto_die => 1;
};
1;
```

Call it like so:
```
$ rex -H localhost print_chars_to_stderr --count=64000
[2015-08-07 12:38:54] INFO - Running task print_chars_to_stderr on localhost
[2015-08-07 12:38:54] INFO - Connecting to localhost:22 (mitch)
[2015-08-07 12:38:54] INFO - Connected and authenticated to localhost.
[2015-08-07 12:38:54] INFO - Successfully authenticated on localhost.
[2015-08-07 12:38:54] INFO - connection isa: Rex::Interface::Connection::OpenSSH
[2015-08-07 12:38:54] INFO - perl -e 'print STDERR "x" for (0..64000);'
```

If it doesn't hang for you, increase the --count parameter and eventually it does. Interestingly, this doesn't happen when printing to STDOUT.

This problem presents itself only when set_no_tty=1. If set_no_tty=0, then it works fine

Can anyone else confirm this issue?

ferki:
Draining the pipes for both STDOUT and STDERR line-by-line whenever any of them has readable data, we can prevent filling up their transfer windows (64k on my machine), which in turn causes the process on the other end of the pipes to wait until the filled buffer gets drained. This should fix #756.

Tested manually by me with OpenSSH and Local connection methods, with both enabled and disabled `Rex::Config->set_no_tty`. I'm not sure how to test it programmatically though.

### DIFF
--- a/lib/Rex/Interface/Exec/Local.pm
+++ b/lib/Rex/Interface/Exec/Local.pm
@@ -16,6 +16,7 @@ use Rex::Commands;
 
 use Symbol 'gensym';
 use IPC::Open3;
+use IO::Select;
 
 # Use 'parent' is recommended, but from Perl 5.10.1 its in core
 use base 'Rex::Interface::Exec::Base';
@@ -94,17 +95,25 @@ sub exec {
   if ( Rex::Config->get_no_tty ) {
     $pid = open3( $writer, $reader, $error, $cmd );
 
-    while ( my $output = <$reader> ) {
-      $out .= $output;
-      $self->execute_line_based_operation( $output, $option )
-        && goto END_OPEN3;
+    my $selector = IO::Select->new();
+    $selector->add($reader);
+    $selector->add($error);
+
+    my $line;
+
+    while ( my @ready = $selector->can_read ) {
+      foreach my $fh (@ready) {
+        $line = <$fh>;
+        goto END_OPEN3 unless defined $line;
+
+        $out .= $line if $fh == $reader;
+        $err .= $line if $fh == $error;
+
+        $self->execute_line_based_operation( $line, $option )
+          && goto END_OPEN3;
+      }
     }
 
-    while ( my $errout = <$error> ) {
-      $err .= $errout;
-      $self->execute_line_based_operation( $errout, $option )
-        && goto END_OPEN3;
-    }
   END_OPEN3:
     waitpid( $pid, 0 ) or die($!);
   }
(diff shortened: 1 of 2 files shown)


krimdomu:
I think, if there is also a problem with `<$fh>` we should use sysread as salva pointed out. Also we should check if the problem also occurs with Net::SSH2 connection and also fix this (if needed)

ferki:
@krimdomu: no, it only happened with OpenSSH and Local. Rex::Interface::Exec::SSH seems to use a different method to get STDOUT and STDERR from the remote (and also does not support `execute_line_based_operation`).

We used <$fh> for ages before we discovered the releated bug, so continueing to use it won't be a regression, but this patch fixes the blocking when more than 64k being written to STDERR by the remote command, so it's already an improvement.

That being said, I fully agree with the `sysread` approach, but that's of "unknown complexitiy" for me, and I'm not sure if it's worth waiting for that, when this patch is already a step forward. I'm glad to work on that on a separate issue (which may also include reading line-by-line for SSH method).

ferki:
Ha! I found a regression caused by this commit: for some reason it breaks the Rex::Test features, more specifically rex can't create test VMs anymore with `Entity: line 2: parser error : Premature end of data in tag domain line 1` or similar errors.

I'll add a sysread based solution.

krimdomu:
I've merged this into 1.4rc branch, because this fixes #884

Issue #832, opened 2015-11-16
alternative storage disk for kvm


Prajithp:
Is there any way to set alternative storage path for kvm?

 right now kvm images are creating under the /root/storage/ and  I would like to change this to /vm/images. 

Here goes my sample Rexfile
```
use Rex -feature => ['0.57'];

use Rex::Commands::Box init_file => "box.yml";

user "root";

group kvm_server => '192.168.1.200';

auth for => "kvm_server" => 
  user => 'root',
  password => 'xxxx';

task "create_vm", group => "kvm_server", sub {
  boxes "init";
};

task setup_vm => sub {
  pkg [ "wget", "nginx" ],
    ensure => "latest";
};
````

**Comment by krimdomu** [comment]
hi @Prajithp ,
i'm sorry but this is currently not possible. The path is currently hardcoded here: https://github.com/RexOps/Rex/blob/374bfd48e62406021b1a158fe67f72df2f1cde65/lib/Rex/Virtualization/LibVirt/import.pm#L39

But this would be easy to change.

**Comment by ferki** [comment]
@Prajithp: until then symlinking might help as a quick and easy workaround- at least that's what I do when I quickly need to do the same.

**Comment by Prajithp** [comment]
@krimdomu  Thank you so much for the clarification.

@ferki Yes, symlinking seems to be the good  idea. But, I have already modified both Box/Base.pm and Virtualization/LibVirt/import.pm to use custom storage path. Please check the below diff.

```
--- ./Virtualization/LibVirt/import.pm  2015-09-04 13:41:22.000000000 +0530
+++ /root/import.pm     2015-11-16 16:23:04.508120837 +0530
@@ -36,9 +36,16 @@
 
   my $dir = dirname $opt{file};
   my ( undef, undef, $suffix ) = fileparse( $opt{file}, qr{\.[^.]*} );
-  my $file = "storage/" . $dom . $suffix;
-  mkdir "./storage";
+  my $storage_path;
+  if ( $opt{storage_path} ) {
+   $storage_path =  $opt{storage_path};
+  }
+  else {
+   $storage_path = "./storage/";
+  }
 
+  my $file = $dom . $suffix;
+  mkdir $storage_path;
   my $format = "qcow2";
 
   my @serial_devices;
@@ -53,7 +60,7 @@
     my @vmdk = grep { m/\.vmdk$/ } i_run "tar -C $dir -vxf $opt{file}";
 
     Rex::Logger::debug(
[...]

PR #835, opened 2015-11-19
#832 Custom storage path for kvm

Prajithp:


krimdomu:
Hi Prajithp,
thanks for the patch. I've just merged it into master branch.

PR #839, opened 2015-11-25
Test kvm improvements

patricklauer:
Wires through memory and cpu setting from Test::Base to Box to Virtualization/libvirt.

Allows setting `$t->cpu( 2 ); $t->memory( 2048 );` in tests to have a VM with larger-than default settings.
Also fixes missing wiring in Rex::Box that didn't pass through values to libvirt, so all machines would end up with default 512MB RAM.

### DIFF
--- a/lib/Rex/Box/Base.pm
+++ b/lib/Rex/Box/Base.pm
@@ -37,7 +37,6 @@ BEGIN {
 
 use Time::HiRes qw(tv_interval gettimeofday);
 use File::Basename qw(basename);
-use Data::Dumper;
 
 sub new {
   my $that  = shift;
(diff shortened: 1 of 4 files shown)


lib/Rex/Box/Base.pm
```diff
@@ -196,7 +196,7 @@ Sets the memory of a VM in megabyte.
 
 sub memory {
   my ( $self, $mem ) = @_;
-  $self->{memory} = $mem;
+  $self->{memory} = $mem * 1024; # libvirt wants kilobytes
```
ferki: What happens if Box is asked to work with VirtualBox backend instead of LibVirt? Is it correct to always multiply the passed memory argument by 1024?

patricklauer:
@ferki thanks for pointing out that vbox uses MB units. I've fixed handling so that Box uses MB externally, and translates for KVM internally.

ferki:
@patricklauer: thanks for this PR! I'm gonna give it some testing, and with your permission, I'll fix a tiny naming consistency issue upon merge (during squashing, etc.).

Issue #794, opened 2015-09-15
The after hook should run even if we fail to connect the the host


mbroadhead:
In the [API documentation](http://www.rexify.org/api/Rex/Commands.pm.html#after-task-sub-), it implies that the "after" hook will execute even if we fail to connect to the server:

```perl
after mytask => sub {
  my ($server, $failed, $cli_args) = @_;
  if($failed) { say "Connection to $server failed."; }

  run "vzctl stop vm$server";
 };
```

However, when I run a simple test, it does not execute the after hook if it fails to connect:
```perl
use Rex -base;

task "foo" => sub {
  say "task foo";
};
after "foo" => sub {
  say "after foo";
};
1;
```

```
$ rex -v
(R)?ex 1.3.3

$ rex -H dkfjdkfj foo
[2015-09-15 11:39:38] INFO - Running task foo on dkfjdkfj
[2015-09-15 11:39:38] INFO - Connecting to dkfjdkfj:22 (mitch)
[2015-09-15 11:39:40] WARN - Can't connect to dkfjdkfj (unable to establish master SSH connection: master process exited unexpectedly)
[2015-09-15 11:39:40] WARN - Error running task/batch: Wrong username/password or wrong key on dkfjdkfj. at /Users/mitch/perl5/perlbrew/perls/perl-5.20.2/lib/site_perl/5.20.2/Rex/TaskList/Base.pm line 309.

[2015-09-15 11:39:40] INFO - Exiting Rex...
[2015-09-15 11:39:40] INFO - Cleaning up...
```

As you can see, the above execution did not output "after foo" as I was expecting.

kablamo:
I agree that it should run the after hook.  This was a problem for me too.  I would love to see a pr for this.  

In the short term, I just discovered there is an undocumented "exit_hook" in Rex::CLI which might help you depending on what you need.  Just be aware that kind of undocumented internal stuff could change or disappear.  To solve this, I wrote my own version of `run()` called `runt()` and handle errors in there.   I needed my own version of `run()` anyway because I need streaming output (like tail -f).  

Both those suggestions suck.  Might be easier to just create a pr to fix this.

mbroadhead:
[...]

PR #843, opened 2015-11-29
The after hook now runs even if the ssh connection fails.  +tests

kablamo:
This is a fix for https://github.com/RexOps/Rex/issues/794.

### DIFF
--- a/t/after.t
+++ b/t/after.t
@@ -1,0 +1,20 @@
+use strict;
+use warnings;
+
+use Test::More tests => 1;
+use File::Temp;
+use Rex::Commands;
+
+#$Rex::Logger::debug = 1;
+$::QUIET = 1;
+
+my $fh = File::Temp->new;
+
+timeout 1;
+
+task  foo => "asdfsadfasdf" => sub { return "yo" };
+after foo => sub { unlink $fh->filename; };
+
+Rex::TaskList->run("foo");
+ok ! -e $fh->filename, "after hook runs even when the ssh connection fails";
+
--- a/lib/Rex/Task.pm
+++ b/lib/Rex/Task.pm
@@ -178,7 +178,8 @@ sub server {
     }
   }
 
-  if ( ref( $self->{server} ) eq "ARRAY" && scalar( @{ $self->{server} } ) > 0 )
+  if ( ref( $self->{server} ) eq "ARRAY"
+    && scalar( @{ $self->{server} } ) > 0 )
   {
     for my $srv ( @{ $self->{server} } ) {
       if ( ref($srv) eq "CODE" ) {
@@ -770,14 +771,14 @@ sub run {
 
     $self->run_hook( \$server, "before" );
 
-    eval {
-      $self->connect($server);
-      1;
-    } or do {
+    eval { $self->connect($server) };
+    if ($@) {
+      my $error = $@;
       $self->{"__was_authenticated"} = 0;
       $self->run_hook( \$server, "after" );
-      die $@;
-    };
+      die $error;
+    }
+
     push @{ Rex::get_current_connection()->{task} }, $self;
 
     if ( Rex::Args->is_opt("c") ) {


t/after.t
```diff
@@ -0,0 +1,20 @@
+use strict;
+use warnings;
+
+use Test::More;
+use Path::Tiny;
+use Rex::Commands;
+
+$::QUIET = 1;
+
+my $tmpfile = Path::Tiny->tempfile;
```
ferki: Do we absolutely need to introduce a new non-core dependency for working with a tempfile? Other tests use `File::Temp` or `Rex::Helper::Path::get_tmp_file()` - can one of them used here too?

t/after.t
```diff
@@ -0,0 +1,20 @@
...
+
+task  foo => "asdfsadfasdf" => sub { return "yo" };
+after foo => sub { $tmpfile->remove };
+
+Rex::TaskList->run("foo");
+ok !$tmpfile->exists, "after hook runs even when the ssh connection fails";
+
+
+done_testing;
```
ferki: I prefer having test plans. That being said, we can merge it like this, but I will clean this up later anyways :)

t/after.t
ferki: I mean it's important because most of these non-core modules are often not available on different OSes, and it makes it hard to impossible for package maintainers to include Rex in different distros. And/or in those cases, _we_ have to package and maintain those dependencies in our own package repositories, which means lots of sustained overhead. And/or those modules are not compatible with perl back to 5.8.9 (which we still have to support, *sigh*).

IMHO all of the above can be avoided by spending some extra minutes on coding instead :)

kablamo: Thanks for the explanation.

kablamo:
- replaced Path::Tiny with File::Temp
  - removed `done_testing`
  - force push
  - ready for review

kablamo:
Ok this got fixed by another commit: https://github.com/RexOps/Rex/commit/57099d11e450b5608f96a437a5631afd388422d8

But my pr has a few super minor things I like.  Mainly a test but also the handling of `$@`.

ferki:
@kablamo: yeah, I'd like to see this merged after the merge conflict has been resolved :+1:

kablamo:
@ferki How do I figure out whats failing here?  Seems like a perltidy failure, but its complaining about line 1 which makes no sense to me...?  Whats wrong with line 1?

ferki:
@kablamo: yes, it does seem like perltidy issue. Just run perltidy on the failing file, using `.perltidyrc` from the root of the repo.

kablamo:
that worked today.  not sure why it wasn't working for me yesterday.  user error apparently.

kablamo:
No it didn't help.  I ran perltidy yesterday and i ran it today and I'm still getting failures.  also the failure is always at line 1.  why is that

ferki:
Because perlcritic only reports whether there is a perltidy problem or not, but it can't tell which line exactly (...). I'll take a look and merge it when passed.

ferki:
It's actually master that has the perltidy problem, introduced on d6fb677133dc96ca6ab797b12a4cda3c38826d0e. Jenkins merges the PR content on top of master before testing, so that's why it reported this commit as broken.

Looks like I need to teach Jenkins about a couple more things to avoid this in the future :)

t/after.t
```diff
@@ -0,0 +1,17 @@
...
+
+use Test::More tests => 1;
+use File::Temp;
+use Rex::Commands;
+
+$::QUIET = 1;
+
+my $fh = File::Temp->new;
+
+task  foo => "asdfsadfasdf" => sub { return "yo" };
```
ferki: Are there a way to make this fail faster? On my machine this test takes 14 seconds, while it basically just waits for the SSH connection to fail.

kablamo: huh it only takes 1.5 seconds on my machine.

kablamo: @ferki I added a timeout of 0 seconds.  let me know if that works for you

ferki:
Ha! The `timeout` setting! Brilliant :+1: I was looking for something like this, but on a much deeper level ^^

On the downside, now it takes 25 seconds for me :confused: Maybe the `timeout` setting is only for Net::SSH2 (I'm using Net::OpenSSH)?

kablamo:
I'm using Net::OpenSSH too.  Timeout works for me.  Uncomment line 8 in `t/after.t` and run the test to see the rex debug logs and let me know what you see there.

krimdomu:
Hi @ferki and @kablamo 
I'm going to merge this pr now. The problem with the timeout is another issue and i will open an issue to track this.

krimdomu:
This pr is now merged. Thank you all and sorry that i commited a same fix without noticing this pr.

PR #848, opened 2015-12-02
Rex::Logger improvements

kablamo:
Some small improvements to the default logger.  

 - renamed `info()` to `logger()`
 - added `warn()`
 - added `error()`
 - added `info()`
 - exported useful methods

```
use Rex;
use Rex::Logger qw/:all/;

debug("doing a thing");
info("boop");
warn("the core is in danger of meltdown");
error("wth are you thinking");
```


lib/Rex/Logger.pm
```diff
@@ -32,11 +32,10 @@ package Rex::Logger;
 
 use strict;
 use warnings;
+use Exporter::Shiny qw/debug info warn error/;
```
ferki: Do we really need `Exporter::Shiny` as a new depdency? :)

kablamo:
- removed Exporter::Shiny dependency even though its awesome :sob: 
  - `warn()` is a perl builtin so all methods now start with `log_`.  Probably makes more sense to have a verb in the func name anyway.
  - force pushed
  - ready for review

lib/Rex/Logger.pm
```diff
@@ -171,6 +174,16 @@ sub info {
...
 }
 
+sub log_error { logger( shift, "error" ) }
+sub log_warn  { logger( shift, "warn"  ) }
+sub log_info  { logger( shift, "info"  ) }
+sub log_debug { debug(@_) }
+
+sub error { logger( shift, "error" ) }
+sub warn  { logger( shift, "warn"  ) }
+sub info  { logger( shift, "info"  ) }
```
ferki: I'm afraid this breaks existing `Rex::Logger::info` calls in the wild, at least when they specified `warn` or `error` levels :/ Am I missing something? Maybe I'm mistaken, didn't test it yet.

kablamo:
Updated to make call to `info()` backwards compatible.  As we agreed on irc this doesn't need to be merged until the next release.  Ready for review though.

Issue #400, opened 2014-08-07
exit_status is not very transparent


aowi:
With exit_status enabled by default since 0.50, Rex has started to return 255 for us every now and then, with no sign of the actual error in the output from Rex.

Internally, there is one or two -1's in @exit_codes. lib/Rex/CLI.pm:549:

``` perl
@exit_codes = Rex::TaskList->create()->get_exit_codes();
```

The feature itself is both good and something we want to use, but just using the first non-zero exit code returned from any of the tasks tells us nothing about what went wrong or where it went wrong. This makes it very hard to figure out why something failed. 

The fact that Rex got an exit code of -1 is also somewhat baffling.

There's also no nice way (as far as I can tell) to disable this feature. For now, we've just set `$Rex::WITH_EXIT_STATUS = 0` in the Rexfile as a workaround.

Things that would be nice to have:
- [x] Configurable exit_status
- [x] Per-task output of exit status, to make post-mortem possible
- [x] No -1's without a good explanation in the output. Whatever caused these should probably also have made Rex yell at me.

Since I don't know the exact source for the last issue, it might not actually be a problem.

I'm more than happy to fix this myself, I'd just like to touch base before I do it.

krimdomu:
For 1.1 we are not at the point to get this better. So i added it to the backlog. I have to take a deeper look at the code how to solve this.

eskaaren:
+1

Also when Rex exits from no Rexfile or no task found the exit code is 0 when it should be 1 or 255 or whatever so that external programs like Jenkins can fail.

kablamo:
I think this pr will go a long way to resolving all the issues mentioned here:
https://github.com/RexOps/Rex/pull/697

kablamo:
Status update on the things people want solved in this issue:

 - [ ] **Configurable exit_status** - If you don't want rex to exit with exit codes, then don't use the 'exit_status' feature?  I guess I don't understand this one.  

 - [x] **Per-task output of exit status, to make post-mortem possible** - This is merged into master in #697 .  See the pr description for details.

 - [x] **No -1's without a good explanation in the output.** - This was also fixed in #697 .  I can't remember the details now, but I believe the exit code 255 and -1 issues are the same or very related..  Here is an explanation: https://github.com/RexOps/Rex/issues/761 and here is the commit that fixed the problem: https://github.com/kablamo/Rex/commit/d0b67aed2480311b51fe9b1a505050cd437cbc22

[...]

PR #854, opened 2015-12-08
Exit with an error message when a bad task name is used.

kablamo:
Rex will now with an error message when a bad task name is used.  Also if the `exit_status` feature is enabled, it will exit 1.  Otherwise it exits 0.

# Testing

```
$ perl bin/rex asdfasdf 
[2015-12-07 21:03:06] ERROR - No task named 'asdfasdf' found. Task names are case sensitive and the module delimiter is a single colon.
$ echo $?
1
```

### DIFF
--- a/lib/Rex/CLI.pm
+++ b/lib/Rex/CLI.pm
@@ -366,24 +366,26 @@ CHECK_OVERWRITE: {
     require "$mod.pm";
   }
 
-  eval {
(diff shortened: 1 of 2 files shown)


ferki:
Thanks, looks good after first try :+1:

ferki:
Pinging @eskaaren based on https://github.com/RexOps/Rex/issues/400#issuecomment-129877518

Issue #851, opened 2015-12-04
Summary shows subtasks (via do_task) unless they are in a transaction


kablamo:
`run_task()` subtasks are not a problem.  Its just `do_task()`.  The easy solution is to add a flag to the task object.  A better solution is to build a task stack trace type mechanism, but I think thats a big project so I'm not going to go that route.

ferki:
I wonder if this still needs to be solved. Currently summary output just states "All tasks were successfull on all hosts" if there was no failure, and if there were any failures, it only lists failed tasks.

In the successful case I'm not sure if it's important to know how many tasks were run exactly (as everything finished fine). And in the failed case, it can be helpful to know which subtask (`do_task`) failed.

Or am I totally missing the point, and the goal here is to make `do_task` accounted for even if it is within transaction?

PR #855, opened 2015-12-08
Don't include do_task() results in the summary

kablamo:
Fix for https://github.com/RexOps/Rex/issues/851

# How to test

 - [ ] `prove t/summary.t`

### DIFF
--- a/lib/Rex/Commands.pm
+++ b/lib/Rex/Commands.pm
@@ -690,15 +690,13 @@ You may also use an arrayRef for $task if you want to call multiple tasks.
 =cut
 
 sub do_task {
-  my $task = shift;
+  my $thing = shift;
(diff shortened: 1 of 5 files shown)


krimdomu:
Currently this crashes on my windows system (win 10, perl 5.22.0) (perl interpreter just dies)
The idea to define a parent task is nice and maybe we can extend this idea to print a tree like summary.

Also i'm not sure if we want to hide `do_task` calls in the summary. The difference between `run_task` and `do_task` is that run_task is used to run one task on one server to get some value in return. So the Rexfile should handle failures with this internally.

`do_task` on the other hand just run the task on all the defined servers so i think it is good to display those results.

krimdomu:
I'm going to add this to 1.5.0 milestone, then we can think about it and how we want to display the results.

PR #856, opened 2015-12-14
Generating SSH Key - create_user

kanishkablack:
Generating SSH Key - create_user

Might be valuable feature to add


ferki:
Thanks @kanishkablack for your contribution, and apologies for the very late feedback.

I believe this functionality would be better offered and maintained as an external rex module than as a core functionality.

If you still need this as a feature and would like to further discuss the topic before (re-)implementing it, please get in touch by opening an issue or messaging us on IRC.

PR #869, opened 2015-12-19
added Rex::Args->get method again. fixes #860

### LINKED ISSUE #860: Rex::Args->get() is missing
Due to some refactoring the `Rex::Args->get` method was removed. We need to readd it and return the current parameters of a task (or, in compatibility mode, return all parameters)

krimdomu:


krimdomu:
I'm going to look at perltidy later

Issue #513, opened 2014-12-22
before_task_start() fails with an ambiguous error when your Rexfile does not return a true value


mbroadhead:
$ rex -v
(R)?ex 0.55.3
$ perl -V:version
version='5.18.1';

This task definition works fine even if your Rexfile does not return a true value:
$ cat Rexfile
task foo => sub {
        say "hello, world!"
};
$ rex -T
Tasks
 foo

If you use "before_task_start" within a Rexfile that does not return a true value, you get an error, but its not clear what the problem is:
$ cat Rexfile
task foo => sub {
        say "hello, world!"
};
before_task_start foo => sub {
        say "before foo";
};
$ rex -T
[2014-12-22 11:47:04] ERROR - There seems to be an error on some of your required files.
[2014-12-22 11:47:05] INFO - Exiting Rex...
[2014-12-22 11:47:05] INFO - Cleaning up...

Compiling the Rexfile does not narrow down the problem:
$ perl -MRex=-base -c Rexfile
Rexfile syntax OK

This works fine though:
$ cat Rexfile
task foo => sub {
        say "hello, world!"
};
before_task_start foo => sub {
        say "before foo";
};
1;
$ rex -T
Tasks
 foo

Running with "rex -dd -T" does not give any other useful information pointing you to the fact that your Rexfile needs to return a true value.

I think a more helpful error message would be what perl normally does for perl modules, something like "Rexfile did not return a true value".

Thanks Rex Team!

krimdomu:
Hi,

this should now be fixed with the above commit. This fix try to evaluate the Rexfile manually and if this works, it just continue the execution. In debug mode it will say something about this.

PR #874, opened 2015-12-20
load a rexfile also if it doesn't return a true value.

krimdomu:
This fixes a regression - RexOps/Rex#513

### DIFF
--- a/lib/Rex/CLI.pm
+++ b/lib/Rex/CLI.pm
@@ -678,10 +678,49 @@ sub load_rexfile {
   }
 
   # load Rexfile
-  eval { require $rexfile };
+  eval {
+    unshift @INC, sub {
+      my $load_file = $_[1];
+      if ( $load_file eq $rexfile ) {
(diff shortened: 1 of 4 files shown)


lib/Rex/CLI.pm
```diff
@@ -678,10 +678,49 @@ sub load_rexfile {
...
+        close($fh);
+        chomp @content;
+
+        push @content, "42;"; # add a true return value at the end of $rexfile.
+          # we need to do this because perl want a "true" value
+          # at the end of a file that is loaded.
+
+        shift @INC; # remove this loader from @INC
+
+        return \join( "\n", @content );
```
kablamo: `\join`?

kablamo:
Thats some crazy shit.  Seems effective though.

lib/Rex/CLI.pm
krimdomu: this is on purpose, this creates a scalar reference which the perl module loader accepts as something like a file handle.

PR #881, opened 2015-12-22
print the error message in the summary when a task die()d

krimdomu:


lib/Rex/CLI.pm
```diff
@@ -612,11 +612,20 @@ sub summarize {
...
+    sort {
+           ncmp( $a->{task}, $b->{task} )
+        || ncmp( $a->{server}, $b->{server} )
+    } @failures
+    )
+  {
+    Rex::Logger::info( "\t$_->{task} failed on $_->{server}", "error" );
+    if ( $_->{error_message} ) {
+      for my $line ( split( $/, $_->{error_message} ) ) {
+        Rex::Logger::info( "\t\t$line", "error" );
```
kablamo: Interesting not sure how this looks.  Might be a good idea.  I think it depends on the how long he error message is.  I was thinking of it like prove where it reported on the # of errors and you scrolled up to see the errors in context.  I assumed the error message would be long and annoying to view inline.  I guess I'm a little dubious that this is going to be readable in all cases.

krimdomu: @kablamo 
Yes, this is has not much style ;-) but without it you won't see any error message.
For example, on one machine i hadn't installed IO::Pty. And when i tried to run rex i didn't saw the error message that it was missing the IO::Pty module. Because it tried to load this module in the fork and so it suppressed that error message.

kablamo: yeah i'm aware of that problem.  something i was thinking of fixing in upcoming prs where I'm refactoring/improving TaskList.  It sounds like we should log the error when its caught?  I'm thinking including it in the summary should be separate decision.   We can polish this in a later release though.

krimdomu: Yes we can polish this later. But it is important that the user is see that error message somewhere, because if it is not there he don't know what it is going on :).

PR #882, opened 2015-12-22
fix error code in $? after a run command which was introduced by 6c8441

krimdomu:


lib/Rex/Interface/Exec/Local.pm
```diff
@@ -119,6 +119,11 @@ sub exec {
 
   }
 
+  # we need to bitshift $? so that $? contains the right (and for all
+  # connection methods the same) exit code after a run()/i_run() call.
+  # this is for the user, so that he can query $? in his task.
+  $? >>= 8;
```
kablamo: It seems like a bad idea to modify $?.  How will users no its not the usual unbitshifted value that Perl normally provides.  Why can't the user bitshift it 8 themselves if they want that value.

krimdomu: Well this has historical reasons. The first connection method was Net::SSH2 which provided already bitshifted return values (afaik, which come directly from libssh2 library). So for all other connection methods i added this.

Also the normal shell behaviour (which admin knows) is to have "normal" return values and no need to bitshift them.

kablamo: Just to clarify -- Net::SSH2 modified $?.

krimdomu: No it doesn't modify it. The Net::SSH2::Channel object has a method `exit_value` (or something similar) which provides the "unbitshifted" return value. The overwirde of $? is done by rex.

PR #886, opened 2015-12-28
Create Rexfile-example2

alx542:
Hi, I was assigned your project in the PR Challenge, and I wasn't quite sure what to do. In trying to get started, I had trouble figuring out how to run some tasks, such as rsync, that I was interested in. Mainly this was because a lot of the code examples in the pods assume you're using the same user and password for different servers. 

So, in case you're interested in using it, I created this Rexfile-example2 as a first step for users  learning about groups with options. 

A helpful person on the support forum noted that there is a resource that explains how to use an INI file.

http://www.rexify.org/docs/rex_book/the_rex_dsl/grouping_servers.html

This is an excellent idea, but IMO, newbies would benefit from a full working example to ease them up the learning curve.

I will probably make use of Rex for some of my projects, so if you have an idea for a different small contribution, I would be open to trying.

### DIFF
--- a/doc/Rexfile-example2
+++ b/doc/Rexfile-example2
@@ -1,0 +1,91 @@
+# enable new Features
+
+use Rex -feature => 0.40;
+
+# This example is meant to give you a quick understanding of how 
+# groups of servers work, and also how options can be associated
+# with each server in a group. Here, each server has 
+# a different user and password. The task "checklogdu_weekly" 
+# also uses a server option=>value to determine if the task should
+# be applied. 
+
+# In practice, a BETTER way to define your groups is with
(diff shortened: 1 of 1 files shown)


ferki:
@alx542: thanks for your participation in the PR challenge! ;) I think your example is showing nicely how one can combine plain perl code and Rex DSL + how to use custom server attributes.

You might also be interested in taking a look at the issues marked as [easy to fix](https://github.com/RexOps/Rex/labels/easy-to-fix) ;)

krimdomu:
Hi @alx542 ,

many thanks for this example.  :+1: 
I've just merged it into development branch.

PR #892, opened 2016-01-01
remove error message, we don't need it for the tests

krimdomu:


ferki:
I'm not sure why dropping the error message is the best course of action here. What do I miss? Is it possible somehow to test for the error message too (since it's a feature now to return the error message)?

It also looks like to be merged despite failing tests, potentially related to #901.

krimdomu:
it is not the best, but (at that time) i don't wanted to search for a test function that allows regex matching with *deeply things.

ferki:
Hm..I'm not sure if with `*deeply` there's such thing as regex matching. But maybe we can check the error message just before it gets deleted. And/or maybe it's enough to simply check if an error message is there or not at first.

PR #902, opened 2016-01-04
Use development versioning

ferki:
- marking release status as a 'testing' release
 - mark version as 'first development release after 1.3.3'

### DIFF
--- a/dist.ini
+++ b/dist.ini
@@ -1,6 +1,6 @@
 name = Rex
-version = 1.4.0
-release_status = stable
+version = 1.3.3_01
+release_status = testing
 author = Jan Gehring <jfried@rexify.org>
 license = Apache_2_0
 copyright_holder = Jan Gehring
--- a/lib/Rex.pm
+++ b/lib/Rex.pm
@@ -78,7 +78,7 @@ use warnings;
 
 # development version if this variable is not set
 if ( !$Rex::VERSION ) {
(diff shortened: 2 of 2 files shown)


krimdomu:
the problem with this is, that with that version scheme we can't test 1.4 feature flag. so we would need some logic in the code that makes 1.4.0 out of 1.3.3_*.

ferki:
Oh.. Would 'if this is a development version, any feature flag is accepted' be enough? Or do we need an exact mapping of which feature flag was added in which (development) version?

krimdomu:
as long as we don't test if rex displays the error message "you try to load a feature that is not available in your version", this would work but i think it is easier to add some logic that just calculate the next real version number out of the dev version number (which would also make it possible to test rexify --use command)

ferki:
Interesting idea! I'm thinking about how to calculate the 'next real version number out of the dev version number', e.g. how to decide programatically if `1.3.4` or `1.4.0` should come after 1.3.3.

ferki:
>13:01 < jfried> FErki: i think it would be okay, if it checks if it is really a developemtn release that is currently running
13:01 < jfried> so for example, in the future: 1.4.2_01  with feature flag 1.5 will be okay. but 1.4.2 with feature flag 1.5 should stop running

Based on our IRC discussion with @krimdomu, we just check if the version string is a development version or not, and if yes, we accept newer version feature flags (from the same major version for now).

PR #908, opened 2016-01-21
Use sysread for local exec interface too (fix #887)

### LINKED ISSUE #887: regression with #820
quote ferki: (see #820)

Ha! I found a regression caused by this commit: for some reason it breaks the Rex::Test features, more specifically rex can't create test VMs anymore with Entity: line 2: parser error : Premature end of data in tag domain line 1 or similar errors.

ferki:


krimdomu:
@ferki 
i think this looks good.
With this patch you don't get the errors you reported in the other ticket?

ferki:
Yes, it solves both the "too much stuff written to STDERR" and "locally run stuff e.g. VM management for testing fails" issues for me. It was also cross-checked by one of my colleagues.

I think I have an idea about how to drain multiple filehandles line-by-line even more correctly, but right now I have no visible issues with the current implementation, so I think I'll just send a separate PR about it later.

PR #915, opened 2016-01-26
Unbundle AWS::Signature4

ferki:
Let's avoid bundling other modules, and just depend on them instead (and package them for OSes that don't provide them by default).


lib/Rex/Cloud/Amazon.pm
```diff
@@ -54,7 +54,7 @@ sub new {
 
 sub signer {
   my ($self) = @_;
-  return Rex::Cloud::Amazon::Sign->new(
+  return AWS::Signature->new(
```
krimdomu: I think this must be AWS::Signature4->new

PR #944, opened 2016-02-18
iptables_clear should clear only loaded tables

kopytov:
Current implementation of iptables_clear forces to load `iptable_(nat|mangle|filter)` kernel modules (and it is not you are really need). In some cases it is not possible to load these modules. For example, if you will try to run iptables_clear in a OpenVZ host node or in a container with default configuration, it will fail because there is no ability to load iptable_nat module ([conntrack is disabled by default](https://openvz.org/Using_NAT_for_container_with_private_IPs#IP_conntracks)).

This patch changes behavior of iptables_clear: it now clears only those tables which is already loaded (and exists in `/proc/net/ip6?_tables_names`).

Please review.

### DIFF
--- a/lib/Rex/Commands/Iptables.pm
+++ b/lib/Rex/Commands/Iptables.pm
@@ -97,6 +97,7 @@ use vars qw(@EXPORT);
 use Rex::Commands::Sysctl;
 use Rex::Commands::Run;
 use Rex::Commands::Gather;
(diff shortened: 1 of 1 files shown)


krimdomu:
Hi @kopytov 

thanks for this patch. I'm going to merge it into 1.4 tree.

PR #945, opened 2016-02-21
Let service tests be silent

ferki:
It makes the output much more readable. Do I need to restore the value of `$::QUIET` to its original value just before finishing? I think it's not needed, but I might be wrong.


krimdomu:
@ferki you have to restore it, because it is a global variable and it won't restore this value after leaving scope.

PR #947, opened 2016-02-23
Reopen STDERR after loading Rexfile

kopytov:
Keeping STDERR closed after loading Rexfile cause problems in modules that opens STDERR on compile time (for example [DDP](https://metacpan.org/pod/DDP)). Patch just reopen STDERR after Rexfile loading.

### DIFF
--- a/lib/Rex/CLI.pm
+++ b/lib/Rex/CLI.pm
@@ -748,10 +748,10 @@ sub load_rexfile {
     # update %INC so that we can later use it to find the rexfile
     $INC{"__Rexfile__.pm"} = $rexfile;
 
-    if ($stderr) {
-      local *STDERR;
-      open STDERR, ">&", $default_stderr;
+    # reopen STDERR
+    open STDERR, ">&", $default_stderr;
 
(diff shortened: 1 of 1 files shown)


ferki:
@krimdomu: I'm tempted to include this in 1.4.0 as a last minute fix - what do you think?

kopytov:
Example of Rexfile to reproduce regression:

```perl
package A;
my $fh = *STDERR;
sub println {
    print {$fh} @_, qq{\n};
}

package main;
use Rex -feature => 1.3;

task test => sub {
    A::println('hello world');
};

1;
```
And then:
```
rex test
```
Before patch it prints nothing. After patch: `hello world`;

ferki:
@kopytov: thanks! I can also confirm your patch solves the problem as expected ;) And I think it looks good :+1:

What we need to ensure that compilation time errors (badly formatted Rexfile, syntax error in an included module, etc) are still reported correctly. I think they are, but let's test thoroughly (I'm also thinking on how we can add tests for it easily).

PR #953, opened 2016-02-25
Fix draining parallel STDOUT and STDERR buffers without newlines

### LINKED ISSUE #951: Rex::Interface::Exec::OpenSSH execute_line_based_operation is no line based operation anymore
this can be seen with the tail command. The tail command sets the readbuffer to 1 byte and this code executes the line based operation for every byte.

ferki:
While keeping line based operations happy too. (Re)introduced on #952. Fixes #951.

ref #756

### DIFF
--- a/lib/Rex/Commands/Tail.pm
+++ b/lib/Rex/Commands/Tail.pm
@@ -79,25 +79,10 @@ sub tail {
 
   Rex::Logger::debug("Tailing: $file");
 
-  my $int =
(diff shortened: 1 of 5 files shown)


ferki:
I think I'm done with this PR, please test (locally/remotely and/or in Windows).

ferki:
Ok, I think I uncovered some improvement possibilities (especially when running on Windows), but for 1.4.0, I'm done.

PR #958, opened 2016-03-04
Normalize EOL characters earlier (fix #957)

### LINKED ISSUE #957: feature flag "tty" doesn't work with Net::OpenSSH anymore
Somewhere in the IOReader.pm module there is a bug removing \r from the output.

This can be seen when running in debug mode. For example:

```
SSH/executing: /sbin/iptables^M-save
```

krimdomu:


krimdomu:
seems to work. i've merged it, because i think the # in the branch name will cause problems with the tests.

PR #964, opened 2016-03-07
Replace [ExtraTests] with [RunExtraTests]

kentfredric:
This means "author" and "release" tests ( which by default are
stored in xt/author/\* and xt/relase/\* ) no longer get copied
to t/author-\* and t/release-\* respectively.

This means that these tests that currently ship as follows:
- t/author-critic.t
- t/author-pod-syntax.t
- t/release-minimum-version.t

Will be instead shipped as:
- xt/author/critic.t
- xt/author/pod-syntax.t
- xt/release/minimum-version.t

And will no longer add 3 extra `exec perl` calls to the end users CPAN
install+test run.

The following things continue to work similarly to how they did before:
- dzil test
- dzil test --release
- dzil test --all
  etc.

It does however mean that this workflow no longer runs author tests from
a CPAN Tarball

$ perl Makefile.PL
$ make
$ AUTHOR_TESTING=1 make test

They can however still be run as follows:

$ perl Makefile.PL
$ make
$ prove -blr t/\* xt/author/*

### DIFF
--- a/dist.ini
+++ b/dist.ini
@@ -8,6 +8,7 @@ copyright_holder = Jan Gehring
 [@Filter]
 -bundle = @Basic
 -remove = MakeMaker
+-remove = ExtraTests
 
 [AutoPrereqs]
 skip = __Rexfile__
@@ -41,6 +42,8 @@ Net::SSH2 = 0
 ; [PodCoverageTests]
 
 [PodSyntaxTests]
+
+[RunExtraTests]
 
 [Prereqs]
 perl = 5.008008


kentfredric:
I've done this against master, because it will be simpler to handle once master is up-to-date.

I can then rebase this on request and pretend there wasn't ever a giant history gap :D

ferki:
Please rebase this PR too.

kentfredric:
This one is technically rebased already, just the PR says "master" as the ref instead of "development".

Currently they're both the same commit, but I can gladly close and reopen a new PR if that makes things simpler.

( Stupid GH doesn't let you simply change the base point branch name ._. )

ferki:
I think we can merge this once we taken care of our remaining perlcritic warnings ^^

ferki:
Better late than never, I'm merging this manually now :) I've rebased it on top of current master, and added some follow-up commits to adapt today's codebase to it.

Thanks @kentfredric for your contribution!

PR #966, opened 2016-03-08
Add META.json to CPAN release

kentfredric:
META.json is a much better communicator of metadata
as META.json by default uses META 2.0 Specification,
which, unlike META.yml + META 1.4, can actually differentiate
between "build" and "test" dependencies.

This means that the dependency graph for people who do

  cpanm --notest Rex

Will be much smaller as a result.

There's naturally a slight amount of risk here if any of your
code inherently relies on dependencies but fails to declare them,
and they just *happened* to be pulled in as children of the "build"
requirements, but become missed as a consequence of being migrated to
children of "test" requirements.

### DIFF
--- a/dist.ini
+++ b/dist.ini
@@ -28,6 +28,8 @@ repository.type = git
 x_twitter       = https://twitter.com/RexOps
 x_IRC           = irc://irc.freenode.net/rex
 
+[MetaJSON]
+
 [OSPrereqs / !~MSWin]
 IO::Pty = 0
 Net::OpenSSH = 0


krimdomu:
Hi @kentfredric 

thanks for the patch. I've tested it and it seems to do the same dependency resolution than without it. By this step i detected that the dependency resolution of Dist::Zilla seems to make some wrong decissions and put some modules into the test section. But this is a different issue.

kentfredric:
> I've tested it and it seems to do the same dependency resolution than without it.

That part depends on where you look.

`Makefile.PL` and `META.yml` don't change at all with this addition. 

Its only the difference of `META.json` which  performs differently under a *cpan client*, because when `META.json` and `META.yml` both exist, a cpan client will prefer the former. ( Assuming the client is new enough ( lol, like since perl 5.12 ) to support it )

PR #968, opened 2016-03-08
Fix #967

### LINKED ISSUE #967: systemctl sometimes outputs to pager and hangs
Based on [systemctl man page](https://www.freedesktop.org/software/systemd/man/systemctl.html), it should be safe to always add `--no-pager` to `systemctl` calls from rex.

stingA0815:
Hey folks,

added the pager option to all systemctl commands.

kind regards
Harm


krimdomu:
Hi @stingA0815 

thanks for this patch.

Issue #976, opened 2016-03-09
Refactor Comands/Kernel.pm to use the new resource structure


krimdomu:
We need to refactor the Command/*.pm modules to use the new resource structure. This is needed to have better reporting, common options (like ensure, on_change) and more things.

krimdomu:
for those who are interessted how such a resource structure looks like, here is an example for the Rex::Commands::Kernel module which i'm am currently working on.

Main module file: https://github.com/krimdomu/Rex/blob/feature_976/lib/Rex/Resource/kernel.pm
Different providers (os support): https://github.com/krimdomu/Rex/tree/feature_976/lib/Rex/Resource/kernel/Provider

PR #996, opened 2016-03-19
Migration of Rex::Commands::Kernel to Resource api

krimdomu:
This is the first migrated Commands module to the new resource api. #976 
There are also some general code changes for resource api, this is why not everything is squashed.

### DIFF
--- a/dist.ini
+++ b/dist.ini
@@ -1,6 +1,6 @@
 name = Rex
-version = 1.4.0
-release_status = stable
+version = 1.4.0_01
+release_status = testing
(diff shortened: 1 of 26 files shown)


ferki:
Cool! I'll go through it fully tomorrow, so I can learn the migration process ^^

dist.ini
```diff
@@ -1,5 +1,5 @@
 name = Rex
-version = 1.4.0
+version = 1.4.0_01
```
ferki: I think setting the `release_status` below to `testing` will fix the test failure we got.

Issue #872, opened 2015-12-20
add possibility for run command to pass parameters as array


krimdomu:
If someone want to execute a command with multiple parameters that contains spaces, quotes and other things it is always complicated for the user to find the right quoting.

So it would be good to have a possibility to pass the arguments as an array to the run command and let rex do the right quoting.

For example something like this:

```perl
run "su", [ "-c", "create user eloor with password 'eloor' ;" ];
```

kentfredric:
Commentary from `#perl` is String::ShellQuote would be useful for this purpose: https://metacpan.org/pod/String::ShellQuote 

And in the view of one day working on windows, Win32::ShellQuote would be useful there: https://metacpan.org/pod/Win32::ShellQuote

krimdomu:
@kentfredric many thanks for the modules, there is also a shellquoting module in Net::OpenSSH which we can use: https://metacpan.org/release/Net-OpenSSH

Lets see which of them fit best

PR #998, opened 2016-03-22
added execution function to have better shell quoting #872

krimdomu:
reference: #872

the second commit refactors all internal run() calls to use i_run(). (i hope i got all)
There needs to be a follow up ticket to refactor all these to use i_exec or i_exec_nohup. But now it is easier to find them with grep...

### DIFF
--- a/lib/Rex/Box/Amazon.pm
+++ b/lib/Rex/Box/Amazon.pm
@@ -82,7 +82,6 @@ use warnings;
 use Data::Dumper;
 use Rex::Box::Base;
 use Rex::Commands -no => [qw/auth/];
-use Rex::Commands::Run;
 use Rex::Commands::Fs;
 use Rex::Commands::Cloud;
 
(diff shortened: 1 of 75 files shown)


ferki:
I think using `i_run` internally instead of `run` lets rex continue on errors when `exec_autodie` is in effect, and it most probably changes rex behaviour in a debatable way (i.e. if somebody relied on "rex will die instantly if something goes wrong", that might not be true anymore).

Maybe we can just simply ask `i_run` to respect `exec_autodie` settings (but then it might break existing user code in a different way, i.e. where users used `i_run` to work around `exec_autodie`).

krimdomu:
every internal command have to use i_run instead of run. because (for example) run gets reported, where i_run doesn't gets reported (which is the right way for internal functions).
Also every call to i_run should check $? if the run was successfull, and if not, it have to stop execution. regardless if `exec_autodie` is set or not.

So the task after this merge is, to make sure that every i_run follows a check to $?. We have two options to do this

* make the check to $? inside i_run/i_exec and throw an exception (which i would prefer)
* manually check $? in every place where i_run/i_exec is used.

ferki:
> every internal command have to use i_run instead of run

Agreed, feels good

> make the check to $? inside i_run/i_exec and throw an exception (which i would prefer)

Even better, thank you! I'll open an issue about it!

PR #1010, opened 2016-04-02
moved task and environment functions into seperate namespace

### LINKED ISSUE #992: Print nicer error message on non-existing task with do_task
If a non-existing task is tried to be executed with `do_task`, rex fails with a `Can't call method "parallelism" on unblessed reference` error. It could check task existence first, and if it's not there, print out a more useful message.

Originally reported by @jolenzy

krimdomu:
moved task and environment functions into seperate namespace. added better error message. ref #992

### DIFF
--- a/lib/Rex.pm
+++ b/lib/Rex.pm
@@ -97,7 +97,9 @@ BEGIN {
 }
 
 our ( @EXPORT, @CONNECTION_STACK, $GLOBAL_SUDO, $MODULE_PATHS,
-  $WITH_EXIT_STATUS );
+  $WITH_EXIT_STATUS, @FEATURE_FLAGS );
+
+@FEATURE_FLAGS = ();
 
 $WITH_EXIT_STATUS = 1; # since 0.50 activated by default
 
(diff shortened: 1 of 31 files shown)


krimdomu:
This is a little bit more than the original task, but when i touch a function in Rex::Commands i will relocate it into a separate namespace. Rex::Commands is too overloaded which makes it hard(er) to use OOP Frameworks like Moo/Moose (or other Modules) because some commands just overwrites commands from these modules.

ferki:
I find this change too big to review, but tests pass at least :) There also seem to be no additional perlcritic warnings.

Slicing Rex::Commands into smaller modules is a good idea :+1: 

I'm gonna install this branch and try to use it as much as possible in order to try to catch any unexpected errors.

ferki:
I get various failures with this patch applied:
- with `-e`: `Undefined subroutine &Rex::Commands::do_task called at /path/to/Rex/CLI.pm line 359, <> line 40.`
- with normal task execution: `Undefined subroutine &Rex::Commands::task called at /path/to/Rex/Commands.pm line 656, <> line 40.`

krimdomu:
@ferki thanks for the test.
I've added 2 tests for this and fixed it.

ferki:
Thanks! Something still fails though according to Travis. I'll take a look later and test again!

Do you think it would be a good idea to keep things on separate commits next time? Like I think the "better error message for do_task" part could have been a small commit, even cherry-pickable separately, while the rest could have been a bigger refactor commit. I think that would help with review.

ferki:
Initial testruns are working for me, thanks! :+1: Let me run it for a couple more days before merging though.

During tests, `t/cli.t` writes a bunch of extra stuff to the terminal:
```
[2016-04-24 13:14:11] WARN - No Rexfile found.
[2016-04-24 13:14:11] WARN - Create a file named 'Rexfile' in this directory,
[2016-04-24 13:14:11] WARN - or specify the file you want to use with:
[2016-04-24 13:14:11] WARN -    rex -f file_to_use task_to_run
[2016-04-24 13:14:11] INFO - Running task eval-line on <local>
```
Adding `$::QUIET = 1;` to it would help keeping the test output clean (I would add it directly, but the branch seems to in your personal repo - maybe I'll open a PR there).

krimdomu:
@ferki 
the Problem with the message their is not a problem with `$::QUIET`. This also gets printed if you use `rex -e`

ferki:
Hmm..still, if I add `$::QUIET = 1;` to `t/cli.t` it doesn't pollute the test output anymore with content of rex. I use `prove --lib --recurse --jobs 4 --timer --shuffle t/` to run the tests locally.

PR #1011, opened 2016-04-02
on windows test for SSH connection type. fix #965

### LINKED ISSUE #965: task.t fails on windows with Net::OpenSSH
Net::OpenSSH can be installed on Windows without forcing it, and if present, it leads to failures in `t/task.t`:
- http://www.cpantesters.org/cpan/report/e97c7360-6c27-1014-a6ef-632b764d66f0
- http://www.cpantesters.org/cpan/report/eb206489-6c27-1014-baf4-db24764d66f0

/cc: reported by @aero via IRC and CPAN Testers (thanks!)

krimdomu:
on windows test for SSH connection type. fix #965

### DIFF
--- a/t/task.t
+++ b/t/task.t
@@ -49,7 +49,7 @@ is( $t1->desc, "Description", "get/set description" );
 
 is(
   $t1->get_connection_type,
-  ( $have_mods{"Net::OpenSSH"} ? "OpenSSH" : "SSH" ),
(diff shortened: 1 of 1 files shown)


ferki:
I didn't test it on windows, but LGTM. We can gather some additional test results by uploading an `1.4.0_01` to CPAN later.

PR #1012, opened 2016-04-02
print right error message if run_task is calling a task that is not available. fix #970

### LINKED ISSUE #970: Unhelpful error message when task name is misspelled
when running `rex Test:run --test=t/foo.t`, when foo.t has a misspelled task:
```
Waiting for SSH to come up on 192.168.122.99:22.
[2016-03-09 14:01:27] <local> DEBUG - Returning existing distribution class of type: Rex::TaskList::Base
[2016-03-09 14:01:27] <local> ERROR - Error running t/foo.t: Can't call method "set_auth" on an undefined value at /home/me/perl5/lib/perl5/Rex/Box/KVM.pm line 193.
```
This is very confusing for the user as it looks like the test VM or the test failed, when it's just a naming mistake.

krimdomu:
print right error message if run_task is calling a task that is not available. fix #970

### DIFF
--- a/lib/Rex/Box/Amazon.pm
+++ b/lib/Rex/Box/Amazon.pm
@@ -199,21 +199,6 @@ sub security_group {
   $self->{security_group} = $sec_group;
 }
 
-sub provision_vm {
-  my ( $self, @tasks ) = @_;
-
-  if ( !@tasks ) {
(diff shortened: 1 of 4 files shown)


ferki:
LGTM :+1: I love the refactor too! :)

PR #1013, opened 2016-04-02
added understandable error message when no virtualization module is given. fix #949

### LINKED ISSUE #949: Improve error message for missing virtualization
While I was trying to run a task for KVM guest creation, I was getting an error message `ERROR - Virtualization Class Rex::Virtualization:: not found.`

The problem was that I've missed to set virtualization: `set virtualization => 'LibVirt'`

The above error message could be more informative and clearly point to a problem.

krimdomu:
added understandable error message when no virtualization module is given. fix #949

### DIFF
--- a/lib/Rex/Virtualization.pm
+++ b/lib/Rex/Virtualization.pm
@@ -47,6 +47,11 @@ sub create {
     $wanted_provider = $wanted_provider->{type} || "LibVirt";
   }
 
+  if ( !$wanted_provider ) {
+    die
+      "No virtualization provider set.\nPlease use `set virtualization => 'YourProvider';` to set one,\nor see `perldoc Rex::Commands::Virtualization` for more options";
+  }
+
   my $klass = "Rex::Virtualization::$wanted_provider";
 
   if ( exists $VM_PROVIDER{$wanted_provider} ) {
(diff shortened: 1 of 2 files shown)


lib/Rex/Virtualization.pm
```diff
@@ -47,6 +47,11 @@ sub create {
     $wanted_provider = $wanted_provider->{type} || "LibVirt";
   }
 
+  if ( !$wanted_provider ) {
+    die
+      "No virtualization provider set.\nPlease use <set virtualization => 'YourProvider';> to set one.\n";
```
ferki: Would it be a good idea to:
- use backticks instead of `<>` for marking code/commands
- also point the user to docs
?

I'm thinkink about something like this:
```
"No virtualization provider set.\nPlease use `set virtualization => 'YourProvider';` to set one,\nor see `perldoc Rex::Commands::Virtualization` for more options";
```

ferki:
I left a slight note about the exact message, but LGTM :+1: Thank you!

ferki:
Works well, will fix the mentioned comments after merge.

krimdomu:
I added the proposed error message

PR #1014, opened 2016-04-02
removed hard coded qemu-img create format and added use of driver_type option. fix #948

### LINKED ISSUE #948: qcow2 image creation problem
I was trying to create a KVM guest with `vm create`. No matter that I've specified `driver_type => 'qcow2'`, Rex still creates a raw image.

krimdomu:
removed hard coded qemu-img create format and added use of driver_type option. fix #948

### DIFF
--- a/lib/Rex/Virtualization/LibVirt/create.pm
+++ b/lib/Rex/Virtualization/LibVirt/create.pm
@@ -83,7 +83,7 @@ sub execute {
       my $size = $_->{'size'};
       if ( !is_file( $_->{"file"} ) ) {
         Rex::Logger::debug("creating storage disk: \"$_->{file}\"");
-        i_run "$QEMU_IMG create -f raw $_->{'file'} $size";
+        i_run "$QEMU_IMG create -f $_->{driver_type} $_->{'file'} $size";
         if ( $? != 0 ) {
           die("Error creating storage disk: $_->{'file'}");
         }
@@ -97,7 +97,8 @@ sub execute {
         "building domain: \"$opts->{'name'}\" from template: \"$_->{'template'}\""
       );
       Rex::Logger::info("Please wait ...");
-      i_run "$QEMU_IMG convert -f raw $_->{'template'} -O raw $_->{'file'}";
+      i_run
+        "$QEMU_IMG convert -f raw $_->{'template'} -O $_->{driver_type} $_->{'file'}";
       if ( $? != 0 ) {
         die(
           "Error building domain: \"$opts->{'name'}\" from template: \"$_->{'template'}\"\n
--- a/t/issue/948.t
+++ b/t/issue/948.t
@@ -1,0 +1,87 @@
+use strict;
+use warnings;
+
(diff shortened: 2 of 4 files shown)


t/issue/948.t
```diff
@@ -0,0 +1,81 @@
...
+    like $exec, qr/^qemu\-img create \-f $fmt/, "qemu-img created a raw image.";
+    $count_exec++;
+  }
+
+  return "";
+}
+
+my $v = Rex::Virtualization->create("LibVirt");
+
+$v->execute(
```
ferki: Those commands emit normal rex logs mixed together with the TAP output. Can you make the tests quiet, by adding e.g. `$::QUIET = 1;`, please?

t/issue/948.t
```diff
@@ -0,0 +1,81 @@
...
+use Rex::Commands::Run;
+
+my $image_format = "raw";
+
+sub get_image_format { return $image_format; }
+
+my $count_file = 0;
+my $count_exec = 0;
+
+no warnings 'redefine';
```
ferki: 1. Should this `warnings` override be in effect to the end of the file? Can we put it inside a block, or even better, put a short `use warnings;` later where the override should end?

t/issue/948.t
```diff
@@ -0,0 +1,81 @@
...
+
+sub Rex::Commands::File::file {
```
ferki: I'm not sure if I feel fully comfortable with the subroutine redefines in the tests. Though I agree that sooner or later we would need some mocking in the test suite, and this can be a precursor to it.

Can you maybe add a `#TODO: implement mocking and/or test this without subroutine redefining`, or similar, so at least it's marked as a task for the future?

ferki:
I left some minor notes, but overall it's :+1:

Just to make sure I fully understand the situation, and don't miss some details, can you share some reasoning about the subroutine redefines in the test, please?

lib/Rex/Virtualization/LibVirt/create.pm
```diff
@@ -83,7 +83,7 @@ sub execute {
       my $size = $_->{'size'};
       if ( !is_file( $_->{"file"} ) ) {
         Rex::Logger::debug("creating storage disk: \"$_->{file}\"");
-        i_run "$QEMU_IMG create -f raw $_->{'file'} $size";
+        i_run "$QEMU_IMG create -f $_->{driver_type} $_->{'file'} $size";
```
ferki: Can we add a note about this to the documentation, please? Something like `With LibVirt, the same driver types are supported as formats qemu-img supports.`

ps.: looks like `driver_type` is not included at all in the current examples in `Rex::Commands::Virtualization`

ferki:
Works well in my tests, I'll address the above points after merge.

krimdomu:
yes, the redefines make it possible to do the testing without executing the whole thing on  a real server.
and yes, we need to use a mocking library for this, but if i add a mocking library this won't be a change for 1.4.1.

PR #1017, opened 2016-04-10
fix hardcoded git tag command

qsimpleq:
fix hardcoded "git checkout -b $tag $tag" to $CHECKOUT_TAG_COMMAND


ferki:
@qsimpleq: good catch, thank you!

PR #1026, opened 2016-04-24
Throw an exception if i_run encounters problems (fix #1016)

### LINKED ISSUE #1016: Check return values inside i_run
ref https://github.com/RexOps/Rex/pull/998#issuecomment-206804456

> make the check to $? inside i_run/i_exec and throw an exception (which i would prefer)

ferki:
This is the right thing to do, but this most probably will break code where `i_run` was used to mask execution failures while using `exec_autodie`. I believe that is a misuse, and proper error checking should be implemented in those cases instead.

This might expose some corner cases we missed earlier, but at least now we can catch the exception and handle them properly. And more importantly, some execution glitches (like around service handling in some cases) can't go undetected.
I already fixed some of those problematic points, like in `can_run()`, dmidecode, and for systemctl detection (see commits). I quickly went through all `i_run` usage, but didn't found anything else obvious at first glance. Plus, most of the internal `i_run` usage were run up until the previous release, and it didn't really died unexpectedly even with `exec_autodie` in effect.

That being said, feel free to report any breakage, so we can include a fix for that.

### DIFF
--- a/lib/Rex/Helper/Run.pm
+++ b/lib/Rex/Helper/Run.pm
@@ -19,6 +19,7 @@ use Net::OpenSSH::ShellQuoter;
 use Rex::Interface::File;
 use Rex::Interface::Fs;
 use Rex::Helper::Path;
+use Carp;
 require Rex::Commands;
 require Rex::Config;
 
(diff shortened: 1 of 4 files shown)


ferki:
Let's convert all cases that are checking for `$? != 0` after an `i_run` (they need to catch the exception instead).

krimdomu:
i just cherry-picked all the commits from this pr to development-1.x branch

krimdomu:
as i said in the comment above, i cherry picked the changes, so i close this pr.

krimdomu:
this absolutely doesn't work. when we throw an exception inside i_run, we have to refactor all code that use i_run.

PR #1029, opened 2016-04-24
Fix compatibility with recent Net::SSH2 (fix #1023)

### LINKED ISSUE #1023: Net::SSH2 0.59 compatibility changes
As salva on irc mentioned, with Net::SSH2 0.59 there will be some changes. We need to find out what code parts we need to change so that Rex works with the new version and also with old ones.

- [x] passing options to connect is deprecated at Rex::Interface::Connection::SSH.pm line 79 (and timeout uses wrong unit, should be passed in milliseconds) -> fixed in `net-ssh2-compat-#1023` branch
- [x] ~~`key_auth` is not working with Net::SSH2-0.59_10~~ false report
- [x] `rex -d` reports ` Found Net::OpenSSH and Net::SFTP::Foreign - using it as default` even if `set connection => 'SSH'` is set (and it is really using Net::SSH2)
- [x] rex hangs at shell detection on executing `version` (for idrac shell detection) as a non-existing command with Net::SSH2-0.59_10: printing to remote STDERR causes the hang -> possible regression in Net::SSH2: rkitover/net-ssh2#32 (or rex retrieves STDOUT/STDERR in a problematic way)

ferki:
- passing options to connect is deprecated.
- timeout value should be passed in milliseconds

### DIFF
--- a/lib/Rex/Interface/Connection/SSH.pm
+++ b/lib/Rex/Interface/Connection/SSH.pm
@@ -67,6 +67,7 @@ sub connect {
 CON_SSH:
   $port ||= Rex::Config->get_port( server => $server ) || 22;
   $timeout ||= Rex::Config->get_timeout( server => $server ) || 3;
+  $self->{ssh}->timeout($timeout);
 
   $server =
     Rex::Config->get_ssh_config_hostname( server => $server ) || $server;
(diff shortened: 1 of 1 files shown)


ferki:
@krimdomu: can you test it on Windows with both 0.58 and 0.59_11 version of Net::SSH2, please?

PR #1030, opened 2016-04-24
Report all packages, even with `bulk_install` (fix #838)

### LINKED ISSUE #838: bulk install is broken
b03eecd0b18516c34b141d406aa1239149f66151 forces `pkg` to be called one-by-one on all passed packages in an arrayref, which makes the package manager to be called as many times as many elements are in the list.

This is basically avoiding `bulk_install` to be called, and adds considerable amount of overhead, slowing down the whole process.

It would be nice if we could do the per-package reporting in a different form, and/or at least tieing this behaviour to a feature flag.

ferki:
This PR reports every change (or non-change) for all packages that were passed to `pkg`.
It even works with `bulk_install`, which is now re-enabled.

I have a variation that creates only a single entry for all packages, with comma separated package names, and then reports all change (or non-change) inside that single report entry. I believe the variation presented here is more useful, but I'm up to fine tune.

On top of that, there are a bunch of refactoring commits to make it simpler to understand.

### DIFF
--- a/lib/Rex/Commands/Pkg.pm
+++ b/lib/Rex/Commands/Pkg.pm
@@ -122,6 +122,9 @@ sub pkg {
 
   foreach my $candidate ( reverse sort @package_list ) {
     if ( my ($change) = grep { $candidate eq $_->{name} } @modifications ) {
+      if ( exists $option{on_change} && ref $option{on_change} eq "CODE" ) {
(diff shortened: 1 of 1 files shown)


krimdomu:
lgtm :+1:

ferki:
@krimdomu: thanks for the review!

Issue #971, opened 2016-03-09
Dist::Zilla creates some wrong dependencies


krimdomu:
For example it puts Parallel::ForkManager, Xml::LibXML and String::Escape into the "test" section.

ferki:
I guess that's because those are optional dependencies, but we do have tests for them, so it has no better way than put it into the test section.

Maybe we can mark optional dependencies as recommended deps, or try to use something like [Dist::Zilla::Plugin::OptionalFeature](https://metacpan.org/pod/Dist::Zilla::Plugin::OptionalFeature).

krimdomu:
I think the dependencies are now right. So i'm closing this issue.

Issue #404, opened 2014-08-10
Rex::Group::Lookup::XML introduced a heavy dependcy to Rex


fanyeren:
XML::LibXML is not shipped by ditros default, and is difficult to installed in many older systems, in our dept, we have 20k machines with RHEL 4.6, i spent several hours, but still failed to install. 

so i doubt is there anybody who really needed this feature in production?

PR #1040, opened 2016-05-05
make some optional dependencies to hard dependencies

krimdomu:
This is a fix for #971 making some optional dependencies to hard dependencies.

This is also to minimize the confusions for users which get error message for missing perl modules.

### DIFF
--- a/lib/Rex/Group/Lookup/XML.pm
+++ b/lib/Rex/Group/Lookup/XML.pm
@@ -33,7 +33,7 @@ use Rex -base;
 require Exporter;
 use base qw(Exporter);
 use vars qw(@EXPORT);
-XML::LibXML->require;
+use XML::LibXML;
 
 @EXPORT = qw(groups_xml);
 
--- a/lib/Rex/Cloud/Amazon.pm
+++ b/lib/Rex/Cloud/Amazon.pm
@@ -19,16 +19,15 @@ use Rex::Logger;
 use Rex::Cloud::Base;
 use AWS::Signature4;
 use HTTP::Request::Common;
-
+use Digest::HMAC_SHA1;
 use base qw(Rex::Cloud::Base);
+use LWP::UserAgent;
+use XML::Simple;
 
 BEGIN {
   use Rex::Require;
-  LWP::UserAgent->use;
-  Digest::HMAC_SHA1->use;
   HTTP::Date->use(qw(time2isoz));
   MIME::Base64->use(qw(encode_base64 decode_base64));
-  XML::Simple->require;
 }
 
 use Data::Dumper;
--- a/lib/Rex/Helper/INI.pm
+++ b/lib/Rex/Helper/INI.pm
@@ -8,10 +8,9 @@ package Rex::Helper::INI;
 
 use strict;
 use warnings;
+use String::Escape qw/string2hash/;
 
 # VERSION
-
-BEGIN { String::Escape->use('string2hash'); }
 
 sub parse {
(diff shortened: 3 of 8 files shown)


lib/Rex/Group/Lookup/XML.pm
```diff
@@ -33,7 +33,7 @@ use Rex -base;
 require Exporter;
 use base qw(Exporter);
 use vars qw(@EXPORT);
-XML::LibXML->require;
+use XML::LibXML;
```
ferki: This may conflict with #404.

ferki:
In general, I like to idea of making as much dependencies hard instead of optional as possible in order to provide a more complete default user experience :+1: This would reduce support overhead as well, while we still can make some dependencies optional again in a bit nicer way (as standard Recommended/Suggested/etc. deps via CPAN).

Also, we might be able to remove some currently skipped modules in `t/base.t`. I guess at least
- `Rex::Group::Lookup::INI`
- `Rex::Helper::INI`
- ` Rex::Group::Lookup::XML`
- `Rex::TaskList::Parallel_ForkManager`

On top of those, I propose making some more deps of `Rex::Cloud::Amazon` hard, as people seem to have difficulties regularly with the optional dependencies of that module.

lib/Rex/Group/Lookup/XML.pm
krimdomu: This is right, but we can't take care for these old distributions. CentOS 4 support ended 2012 (https://de.wikipedia.org/wiki/CentOS#Versionen) and RHEL 4 support ended 2015 (https://de.wikipedia.org/wiki/Red_Hat_Enterprise_Linux#Versionen)

krimdomu: What we can do is, that we provide a package where rex is bundled with perl, so that you can just run it on every glibc compatible linux system.

ferki: No, I'm especially fine with dropping support for unsupported distros. I just read that [RHEL4 has extended support till 2017-03-31](https://en.wikipedia.org/wiki/Red_Hat_Enterprise_Linux#Product_life_cycle). Let's find some official info.

ferki: [Official EOL of RHEL4 extended support is 2017-03-31](https://access.redhat.com/support/policy/updates/errata/#Life_Cycle_Dates). I'm fine with a Rex "policy" of only supporting up to normal production lifetime (i.e. excluding extended life-cycle).

On one part, I believe as a software project/community we should endorse best practices such as updates and upgrades done in a timely fashion regularly. On the other hand, taking on extended lifecycle support for RHEL5 would mean being stuck with lots of ballast till end of 2020. Which sounds scary.

krimdomu: I don't think that we want to support until the end of extended lifecycle which is often only available with payment and so we also can't test on those distributions. Also (in regard of the mentioned ticket) CentOS 4 isn't really on the supported os list. On the other hand, people can also continue using an old version of rex (we don't delete the source and rpm/deb files from rex.linux-files.org) and so the people can also build a version on their own (or with support by us via irc) which can remove those dependencies they don't need.

krimdomu:
Dependencies on R::C::Amazon are now also hard

lib/Rex/Group/Lookup/XML.pm
ferki: Exactly, agreed. Let's support till normal EOL :+1:

Issue #437, opened 2014-09-30
Split remote execution commands and configuration management commands 


krimdomu:
We need to split remote execution commands (functions that doesn't describe a state [ex.: is_file(), free(), ...]) and configuration management commands (functions that describe a state [ex.: file(), service(), pkg(), ...]) into different namespaces.
- Execution functions: Rex::Commands::\* 
- Config-Management functions: Rex::Resource::*

This has the advantage that it is clear which functions are idempotent and which are not. And we can define meta parameters like on_change, subscribe_to, only_if, ... to all resources.

This step need special attention, because we have to maintain backward compatibility.

Since 0.53 there is a new function calles resource(). With this function it is possible to create new resources. All the resources that are created with this function will have the mentioned meta parameters and can trigger events like "created", "changed" or "removed".

All the resources (rex basic ones, and custom ones) will be also visible in the reporting. And example resource can be found here: http://bit.ly/ZoGe5N)

PR #1042, opened 2016-05-08
show how to assign a user with a non-default shell

andrewsolomon:
Please note - I only tested this using the 'account' method.

A question: am I right in thinking that the only difference between account and create_user is the 'ensure => present/absent'?

### DIFF
--- a/lib/Rex/Commands/User.pm
+++ b/lib/Rex/Commands/User.pm
@@ -26,6 +26,7 @@ With this module you can manage user and groups.
      password    => 'blahblah',
      system      => 1,
      create_home => TRUE,
+     shell       => '/bin/bash',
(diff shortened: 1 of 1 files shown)


ferki:
@andrewsolomon: Good catch, thank you!

The difference is `account` is a "resource", and `create_user` is a "command". One tell-tale of resources is the `ensure` option, but the main feature of resources is idempotency, and on top of that, resources get reported via the Report feature (which is kinda an audit log of what rex did (or not) on a specific host during a specific execution).

#437 (and some other issues when searched for resource) is about separating them more obviously, and is an on-going effort.

ferki:
ps.: I added you to the list of contributors ;)

Issue #1054, opened 2016-06-01
copypasto in Server.pm?


shattered:
https://github.com/RexOps/Rex/blob/master/lib/Rex/Group/Entry/Server.pm#L215

public_key looks out of place here.

stiron:
@krimdomu 
I can create a pull request during the weekend with the fix if you think.

stiron:
@krimdomu 
It's done now.

PR #1055, opened 2016-06-03
Fix for #1054 - get_private_key() to check the private key file

stiron:
I just created the fix.


krimdomu:
Hi @stiron ,

thank you for this patch. I'm going to merge it.

PR #1059, opened 2016-06-22
Report warnings from package commands

abeverley:
This patch adds a framework to test for and report warnings and errors from the package commands. Currently only the Debian module includes tests, but others can be easily added.


krimdomu:
@abeverley thank you very much for this patch. I'm going to merge it into master branch.

abeverley:
Brilliant, thanks guys.

PR #1063, opened 2016-07-01
Don't add duplicates env variable in crontab

rapenne-s:
Actually if you add an env variable to a crontab it will be added even if it's already in the file.

With this commit, it checks if it exists and then
exists with same value as the new one => we do nothing
exists with a different value => we remove it using its index and we add the new one

### DIFF
--- a/lib/Rex/Commands/Sysctl.pm
+++ b/lib/Rex/Commands/Sysctl.pm
@@ -38,6 +38,7 @@ use warnings;
 
 use Rex::Logger;
 use Rex::Helper::Run;
+use Rex::Commands::File;
 
 require Rex::Exporter;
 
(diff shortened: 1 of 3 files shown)


krimdomu:
Hi @rapenne-s ,

thank you for this patch. I've just merged it into development branch and also into the 1.4 branch. Which is the branch for the next 1.4.1 bugfix release where we are currently working to make it stable.

ferki:
@rapenne-s: great, thanks for this patch!

hint: you can quickly see the results of the automated test suite, as indicated by the red X (or green checkmark) next to the commits, and by the test summary at the bottom of the PR :wink:  For this PR, it all seem to be cleaned up thanks to @krimdomu :)

PR #1067, opened 2016-07-09
Using OpenBSD rcctl tool for managing services

rapenne-s:
OpenBSD as a tool to manage services since 5.8, this commit let use rcctl for managing services from Rex now.

### DIFF
--- a/lib/Rex/Service/OpenBSD.pm
+++ b/lib/Rex/Service/OpenBSD.pm
@@ -24,34 +24,17 @@ sub new {
   bless( $self, $proto );
 
   $self->{commands} = {
-    start   => '/etc/rc.d/%s start',
-    restart => '/etc/rc.d/%s restart',
-    stop    => '/etc/rc.d/%s stop',
-    reload  => '/etc/rc.d/%s reload',
-    status  => '/etc/rc.d/%s status',
-    action  => '/etc/rc.d/%s %s',
+    start        => '/usr/sbin/rcctl start %s',
+    restart      => '/usr/sbin/rcctl restart %s',
+    stop         => '/usr/sbin/rcctl stop %s',
+    reload       => '/usr/sbin/rcctl reload %s',
+    status       => '/usr/sbin/rcctl check %s',
+    ensure_start => '/usr/sbin/rcctl enable %s',
+    ensure_stop  => '/usr/sbin/rcctl disable %s',
+    action       => '/usr/sbin/rcctl action %s',
   };
 
   return $self;
-}
-
-sub ensure {
-  my ( $self, $service, $options ) = @_;
-
-  my $what = $options->{ensure};
-
-  if ( $what =~ /^stop/ ) {
-    $self->stop( $service, $options );
-    delete_lines_matching "/etc/rc.conf",
-      matching => qr/rc_scripts="\${rc_scripts} ${service}"/;
-  }
-  elsif ( $what =~ /^start/ || $what =~ m/^run/ ) {
-    $self->start( $service, $options );
-    append_if_no_such_line "/etc/rc.conf",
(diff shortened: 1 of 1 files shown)


ferki:
@rapenne-s: thanks for this patch!

We try to aim to support OSes with Rex that are supported by their respective upstream. According to my info OpenBSD 5.7 has been released on 2015-05-01, so I guess its already EOL. Can you confirm that my info is correct, please (possibly by linking some official source about that)?

lib/Rex/Service/OpenBSD.pm
```diff
@@ -24,34 +24,17 @@ sub new {
...
-    restart => '/etc/rc.d/%s restart',
-    stop    => '/etc/rc.d/%s stop',
-    reload  => '/etc/rc.d/%s reload',
-    status  => '/etc/rc.d/%s status',
-    action  => '/etc/rc.d/%s %s',
+    start        => '/usr/sbin/rcctl start %s',
+    restart      => '/usr/sbin/rcctl restart %s',
+    stop         => '/usr/sbin/rcctl stop %s',
+    reload       => '/usr/sbin/rcctl reload %s',
+    status       => '/usr/sbin/rcctl check %s',
```
ferki: Well done preserving backwards compatibility! Do you think it would be a good idea to also add an explicit `check => ...` line for this?

Since I'm not an active OpenBSD user, I wonder if that would be handy for OpenBSD users or not. I mean to be able to use the same "words" as they would use on the command line.

ferki:
I've left a small question on the diff, but overall it looks good and simplifies OpenBSD management :+1:

lib/Rex/Service/OpenBSD.pm
rapenne-s: I don't think that would be useful to map a check to rcctl check because it would break the compatibility of a Rexfile that we would use on another system.

rapenne-s:
I can confirm you that OpenBSD 5.7 is EOL. http://www.openbsd.org/faq/faq5.html
" When very important fixes are made to -current, they are backported into the two supported -stable branches. "

Actually the latest release is 5.9 so 5.8 is supported and 5.7 support as been dropped.
OpenBSD discourage people to stick with an out-of-date version.

lib/Rex/Service/OpenBSD.pm
rapenne-s: What could be very useful is a flags => "-A" for a service because it's a way to give parameters for the service. The syntax is `rcctl set apmd flags -A ` to set parameter -A for apmd service. 

Actually I use `run` to do it.

ferki: > I don't think that would be useful to map a check to rcctl check because it would break the compatibility of a Rexfile that we would use on another system.

Yeah, I meant keeping `status` as is (for backwards compatibility), but also adding `check` (to be inline with openbsd terms). If you think that wouldn't be that useful, it's OK like this (as we can add `check` later anytime when needed).

> What could be very useful is a flags => "-A"

I see, and agree :+1: If it's easy to add similar to `action`, then feel free to push one more commit to this branch. Otherwise I'd say let's keep this PR small and easy, and add `flags` separately.

rapenne-s: no urge for `action` now, maybe I will add it later but it's easy to use `append_or_amend` or `run "rcctl..."`

ferki:
@rapenne-s: please rebase this on top of current development branch, and push again (even forced, if needed). Unfortunately the development branch had a test failure at the time when you forked your feature branch off, and that's why tests fail now here.

Otherwise I think it looks good, so let's see those tests passing :+1:

krimdomu:
Hi @rapenne-s ,

thank you very much for this patch. I've just merged it into development and development-1.x branch.

PR #1069, opened 2016-07-09
Sysctl saving

rapenne-s:
This commit permit the following :

sysctl "net.ipv4.ip_forward" => 1, persistent => "true";

so net.ipv4.ip_forward will be set to 1 and net.ipv4.ip_forward=1 will be added/updated in /etc/sysctl.conf

### DIFF
--- a/lib/Rex/Commands/Sysctl.pm
+++ b/lib/Rex/Commands/Sysctl.pm
@@ -94,11 +94,9 @@ sub sysctl {
       Rex::Logger::debug("$key has already value $val");
     }
 
-    if ( defined %options ) {
-      if ( exists $options{persistent} && $options{persistent} eq "true" ) {
-        Rex::Logger::debug("Writing $key=$val to sysctl.conf");
-        sysctl_save $key, $val;
-      }
+    if ( exists $options{persistent} && $options{persistent} eq "true" ) {
+      Rex::Logger::debug("Writing $key=$val to sysctl.conf");
+      sysctl_save $key, $val;
     }
 
   }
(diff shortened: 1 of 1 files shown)


lib/Rex/Commands/Sysctl.pm
```diff
@@ -79,6 +94,11 @@ sub sysctl {
       Rex::Logger::debug("$key has already value $val");
     }
 
+    if ( exists $options{persistent} && $options{persistent} eq "true" ) {
```
krimdomu: @rapenne-s maybe we can just do `if ( exists $options{persistent} && $options{persistent} ) {` there, so people can also use `persistent => 1` to activate this.

rapenne-s: I was thinking about this but I think I should add a way to remove the sysctl from /etc/sysctl.conf.

That could be changed to 

```
if ( exists $options{persistent} )
  if ( $options{persistent} )
    sysctl_save;
  else
    sysctl_remove;
}
```
Better ?

ferki: I think `sysctl_remove` should only be called if `persistent` is FALSE.

As a somewhat related topic, I would find `ensure => 'present'` and `ensure => 'absent'` more clear about what it means and does, plus more in-line with the usual rex syntax for other things.

ferki:
@rapenne-s: please rebase this on top of current development branch, and push again (even forced, if needed). Unfortunately the development branch had a test failure at the time when you forked your feature branch off, and that's why tests fail now here.

ps.: what do you think about having `ensure => 'present'` and `ensure => 'absent'` instead of `persistent`?

krimdomu:
Hi @rapenne-s 

thank you also for this patch :+1: 

I've just merged it into development and development-1.x branch.
I modified it a bit.

I changed 
```perl
if ( exists $options{persistent} && $options{persistent} eq "true" ) {
```
to
```perl
if ( $options{persistent} ) {
```

PR #1071, opened 2016-07-13
Adds 64bit module path to @INC for linux

mbroadhead:
I'm using rex 1.4.0 on Fedora 24.

If you use `rexify --resolve-deps` to install 64-bit modules into your project's tree, rex will not load them because that path does not exist in @INC.

For example, on Fedora 24, adding **DateTime** as a dependency and running `rexify --resolve-deps` downloads DateTime.pm to:
lib/perl/lib/perl5/x86_64-linux/DateTime.pm

This patch adds  "lib/perl/lib/perl5/x86_64-linux/" to @INC for linux hosts.


krimdomu:
Hi @mbroadhead ,

thanks for this patch. I'm going to include it also in 1.4.1

PR #1074, opened 2016-07-18
file: do md5sum only when necessary.

slashbeast:


ferki:
@slashbeast : please rebase this on top of current development branch, and push again (even forced, if needed). Unfortunately the development branch had a test failure at the time when you forked your feature branch off, and that's why tests fail now here.

Otherwise it looks good, so let's see it passing tests too :+1:

slashbeast:
Done!

lib/Rex/Resource/file/Provider/base.pm
```diff
@@ -76,7 +76,10 @@ sub test {
   my $is_dir    = $self->is_dir;
   my $is_absent = any { $self->config->{ensure} eq $_ } qw/absent/;
 
-  if ( $is_file && $self->fs->is_file($file) ) {
+  if ( $is_file
+    && $self->fs->is_file($file)
+    && ( $self->config->{source} || $self->config->{content} ) )
```
ferki: Looks like the new extra condition also skips file stat comparisons, not just checksum comparison: https://github.com/slashbeast/Rex/blob/f09a95d8b0cced1c4a662464caa0d0398a304f9d/lib/Rex/Resource/file/Provider/base.pm#L82-L102

Can you make sure it only skips checksum calculation, please?

If you can, please add a test for this case first (no `content` or `source`, but at least one of `owner`/`group`/`mode` should be changed), but let me know if you would like me to add it.

ferki:
Great, thanks!

lib/Rex/Resource/file/Provider/base.pm
ferki: I can't comment directly on the line below, but I think `$self->_set_upload(0);` still should be called even if the checksum calculation is skipped, to signal "there's no need to change the _content_ of the file" (but other changes might still be necessary).

slashbeast:
Hi, Updated as per comments.

krimdomu:
Hi @slashbeast ,

thanks for this patch. The tidy problem is not your fault. I'm going to merge this pr.

PR #1077, opened 2016-07-21
Escape backquotes when run command via sudo in the remote host, fix 1076

nfetisov:
Adding regex to escape ` like " , $ and \ symbols when preparing command 
line to run in the inner shell.


krimdomu:
Hi @nfetisov ,

thanks for this patch :+1: .
I've merged it into development and development-1.x branch.

PR #1081, opened 2016-07-28
try to add same update_system behaviour for as many systems as possile

krimdomu:
This is for RexOps/Rex#546


ferki:
I didn't give it a try yet, but looks fine to me at first glance :+1: (and I still need to enable travis tests for `development-1.x`)

PR #1084, opened 2016-07-30
[WIP] resources and functions are now similar

krimdomu:
- Added Controller class
- Refactored resources so that resources and functions work similar
- resources now also can have multiple code and type checking


krimdomu:
* removed controller class
* added multisub class to handle functions with same name distinguished by signature.
* migrate resource and functions to multisub

ferki:
@krimdomu: the test suite currently fails due to untidy files in the `bin` directory.

krimdomu:
@ferki yes, i know. the test suite would also fail if these file are tidy ;-)

krimdomu:
this pull request is obsolete: see #1212

PR #1086, opened 2016-08-05
sysctl supports remove and keyword changed to ensure

rapenne-s:
I changed the keyword persistent => TRUE to the usual ensure => "present" and now you can also delete a line from /etc/sysctl.conf with ensure => "absent"

### DIFF
--- a/lib/Rex/Commands/Sysctl.pm
+++ b/lib/Rex/Commands/Sysctl.pm
@@ -16,8 +16,6 @@ Version <= 1.0: All these functions will not be reported.
 
 All these functions are not idempotent.
 
-This function doesn't persist the entries in /etc/sysctl.conf.
-
 =head1 SYNOPSIS
 
  use Rex::Commands::Sysctl;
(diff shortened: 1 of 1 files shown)


lib/Rex/Commands/Sysctl.pm
```diff
@@ -95,10 +99,14 @@ sub sysctl {
       Rex::Logger::debug("$key has already value $val");
     }
 
-    if ( $options{persistent} ) {
+    if ( $options{ensure} eq "present" ) {
```
ferki: Does it work if no options are passed at all (i.e. calling `sysctl` command from existing code)? Wouldn't this complain about `Use of uninitialized value in string eq` or similar due to being undefined?

rapenne-s: Yes it was throwing an error if not using ensure. I committed a change to check if ensure is used, and also I added a check to be give an error if it's not absent/present, so if you type "presetn" you will know something is wrong.

ferki:
Thank you! :+1:

ferki:
I also merged it into the development branch.

PR #1087, opened 2016-08-18
Fix Rsync.pm paths and ports.

necrophcodr:
This fixes the following issues with Rsync.pm:

Not correctly translating paths to usable ones.
Unable to use non-standard ports for transfers.

Additionally, this commit also does away with defaulting to transfers with -p enabled, since it is unlikely the local uid and gid will be available anywhere else, and unlikely that this is even what the user wants.


krimdomu:
Hi @necrophcodr 
thanks for this patch. I've just merged it into development-1.x branch.

Issue #1051, opened 2016-05-28
Logger.pm - uninitialized value $host


stiron:
This is Rex version: 1.4.0

The logger class cannot put the target hostname in particular situations. I did not have enough time to deep-analyse the whole issue, but I assume the `$host` variable is empty in the time of the substitution

I see such messages:
`Use of uninitialized value $host in substitution (s///) at /home/tmolnar/.perl5/perlbrew/perls/perl-5.22.0/lib/site_perl/5.22.0/Rex/Logger.pm line 270.`

I use this log format in my Rexfile:
`logformat "[%D] %l %h - %s";`

As a temporary solution I patched my system this way:
(In Rex/Logger.pm)
```
- $line =~ s/\%h/$host/gms;
+ $line =~ s/\%h/$host/gms if $host;
```

stiron:
@krimdomu Do you have any news regarding this one? As I see in several cases the $host cannot be resolved.

krimdomu:
@stiron sorry, i haven't looked into this yet. i think that this happens when there is no connection to any server yet.

stiron:
@krimdomu yeah, that's the case. My quickfix was to use a statement modifier, and do not substitute the $host if it is not present. Alternatively there can be a default <local> in it as in other logs.

If you think I can send a PR with my fix above.

ferki:
@stiron: sounds fine, feel free to send a PR with the `<local>` variation (for consistency).

PR #1088, opened 2016-08-25
Fix for  #1051 - Logger.pm - uninitialized value

stiron:


ferki:
There seems to be a [test failure](https://travis-ci.org/RexOps/Rex/builds/154982941#L297-L311) for this change. Can you update the PR, please?

Since it seems to be perltidy-only, feel free to just amend the commit and force push here again, and github should pick it up automatically (in other words, IMHO there's no need to add a second tidy-only commit).

stiron:
ATM I am not at a computer where I can do that, but I will update the commit.

stiron:
Fix'd!

krimdomu:
@stiron thanks for the cleanup. I'm going to merge it.

PR #1094, opened 2016-10-06
Trying to fix #791

### LINKED ISSUE #791: Rex fails to guess server name for some .ssh/config
Having a config like this on .ssh/config

```
Host foo bar web*
    Hostname %h.mydomain.tdl
    User username
    IdentityFile ~/.ssh/my_special_key
```

Rex fails to connect to any host without '*' (foo and bar in the example). It throws:

```
[2015-09-10 12:34:07] INFO - Connecting to %h.mydomain.tdl:22 (root)
[2015-09-10 12:34:09] WARN - Can't connect to %h.mydomain.tdl (unable to establish master SSH connection: master process exited unexpectedly)
```
The ones with the wildcard work as expected, as any other Host entry defined in the config file.

pacoesteban:
#791 
Now get_ssh_config_hostname sub returns a good hosntame.

If one has a config on ssh_config like:

<pre>
Host server1 web0* db*
    Hostname %h.mydomain.com
    User foo
    IdentityFile ~/.ssh/my_key
</pre>

`%SSH_CONFIG_FOR` on `Config.pm` has something like this:

<pre>
      {
          'server1' => {
                           'identityfile' => '~/.ssh/my_key',
                           'user' => 'foo',
                           'hostname' => '%h.mydomain.com'
                         },
          'web*' => {
                           'identityfile' => '~/.ssh/my_key',
                           'user' => 'foo',
                           'hostname' => '%h.mydomain.com'
                         },
          'db*' => {
                           'identityfile' => '~/.ssh/my_key',
                           'user' => 'foo',
                           'hostname' => '%h.mydomain.com'
                         }
        };
</pre>
That means that when you call for `web01` for instance, on `lib/Rex/Interface/Connection/SSH.pm` (line 69) or `lib/Rex/Interface/Connection/OpenSSH.pm` (line 65) you'll get `web01` because there's no matching entry on `%SSH_CONFIG_FOR` but you'll get `%h.mydomain.com` for `server1` because there's a matching entry.

I try to fix this with minimal changes (maybe is not the right approach).

Tested with Net::OpenSSH and Net:SSH2. The later has some cases that don't work but are not related with the patch (they don't work either without the patch applied).

### DIFF
--- a/lib/Rex/Config.pm
+++ b/lib/Rex/Config.pm
@@ -679,7 +679,12 @@ sub get_ssh_config_hostname {
     && exists $SSH_CONFIG_FOR{ $param->{server} }
     && exists $SSH_CONFIG_FOR{ $param->{server} }->{hostname} )
   {
-    return $SSH_CONFIG_FOR{ $param->{server} }->{hostname};
+    if ( $SSH_CONFIG_FOR{ $param->{server} }->{hostname} =~ m/^\%h\.(.*)/ ) {
+      return $param->{server} . "." . $1;
+    }
+    else {
+      return $SSH_CONFIG_FOR{ $param->{server} }->{hostname};
+    }
   }
 
   return 0;


krimdomu:
Hi @pacoesteban ,

thank you much for the patch. I've just merged it.

PR #1096, opened 2016-10-07
Give the kernel time to get informed of partition changes

### LINKED ISSUE #1095: partprobe should retry if it fails after partition creation
Right now after partition creation, `rex` immediately calls `partprobe`. Sometime the raid controller is not able to refresh the information fast enough and it fails. If we call it later, it can refresh the partition table. We should wait and retry if it fails.

roa:
Fix #1095 

For example hardware raid controllers may need some time to commit
changes. If we manage to inform the kernel immediately, behavior is
the same. If it fails, it retries a few times after a sleep.


krimdomu:
Hi @roa ,

thanks for this patch. i've just merged it into development-1.x branch.

Issue #1102, opened 2016-10-12
rex needs at least version 0.25 of Data::Validate::IP


ferki:
According to its [changelog](https://metacpan.org/changes/distribution/Data-Validate-IP), `is_ip()` has only been added to Data::Validate::IP in 0.25. Since Rex doesn't require a specific version, it leads to test failures e.g. on Mac OS X, like `Error:  "is_ip" is not exported by the Data::Validate::IP module`

Example failures:
- http://www.cpantesters.org/cpan/report/dc35753c-903c-11e6-ab25-f642c9d0f03a
- http://www.cpantesters.org/cpan/report/9c7d9a34-8fef-11e6-836b-c0b7c8d0f03a

PR #1110, opened 2016-10-30
don't use is_ip function

krimdomu:
don't use is_ip function, so that it also works with older Data::Validate::IP module


krimdomu:
this fixes #1102

PR #1118, opened 2016-12-03
Fix uninitialized value warnings when using Augeas module

abeverley:
When using the Augeua module remove or modify commands without an on_change parameter, uninitialized value warnings are displayed.


krimdomu:
Hi @abeverley 
thanks for this patch. I've just merged it into development and development-1.x branch.

PR #1119, opened 2016-12-14
Rex::Pkg::Gentoo: Smarter installed check to reinstall on USE updates

alip:
Rex does not know about package USE flags on Gentoo, thus ensuring a
package's presence does not necessarily mean the package is installed
with the most recently updated USE flags.

This commit addresses the issue by invoking emerge --pretend --changed-use in
is_installed for installed packages to check for possible changes in the
installed package. The option --quiet is to skip unnecessary output and
--verbose is to have the potential USE flag change logged in Rex debug
output.

Here is a simple scenario:

    my $pkg = 'virtual/ssh';
    my $use = 'minimal';

    # Enable $use flag for $pkg in portage configuration.
    pkg $pkg, ensure => 'present';

    # Disable $use flag for $pkg in portage configuration.
    pkg $pkg, ensure => 'present';

Before this change the second `pkg` invocation would not reinstall
the package with the USE flag disabled. The logic is the same for
target (e.g: ruby, python) changes.

### DIFF
--- a/lib/Rex/Pkg/Gentoo.pm
+++ b/lib/Rex/Pkg/Gentoo.pm
@@ -25,8 +25,9 @@ sub new {
   bless( $self, $proto );
 
   $self->{commands} = {
-    install         => 'emerge --update %s',
-    install_version => 'emerge --update =%s-%s',
+    install         => 'emerge --update --changed-use %s',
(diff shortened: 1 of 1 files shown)


krimdomu:
Hi @alip 
thank you also for this patch. I'm going to merge it into development-1.x branch.

PR #1129, opened 2017-01-03
Rex::Pkg::Gentoo: Add support for slotted package specs

alip:
Rex does not have a way to specify package SLOT using pkg command on
Gentoo. Thus it is not possible to ensure a specific slot of a package is
installed.

This commit addresses the issue by adding support to is_installed function for
slot specifications. The slot may be specified either in the package name or
the version (but not both, which throws an error).

    Here is a simple scenario:

	# Ensure latest version with specified slot is installed
	pkg 'www-client/links:2', ensure => 'latest';

        # Ensure specified version with specified slot is installed
        pkg 'www-client/links, ensure => '2.14:2';

	# This is invalid and will throw an illegal package spec error.
	pkg 'www-client/links:2', ensure => '2.14:2'

### DIFF
--- a/lib/Rex/Pkg/Gentoo.pm
+++ b/lib/Rex/Pkg/Gentoo.pm
@@ -52,11 +52,26 @@ sub bulk_install {
 sub is_installed {
 
   my ( $self, $pkg, $option ) = @_;
+  my $slot;
   my $version = $option->{version};
+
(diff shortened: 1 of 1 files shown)


krimdomu:
Hi,

thanks much for this patch. I've just merged it and it will be released with 1.5.0

PR #1136, opened 2017-03-14
set_auth used before creating the task object

Prajithp:
Can't locate object method "set_auth" via package "install" (perhaps you forgot to load "install"?) at /usr/local/share/perl5/Rex/CLI.pm line 231.

 I was getting the above error message, while using -P argument.


Prajithp:
Any update on this?

krimdomu:
Hi @Prajithp ,

thanks for this patch and sorry for the delay. This will be included in the next release.

PR #1144, opened 2017-05-25
Use fail_ok in cron

ddico:
If cron list is empty ret_val is 1, triggering an error in i_run
Probably this was missed in #1123


krimdomu:
Hi @ddico ,

many thanks for the patch. I'm going to merge it into 1.5.1.

PR #1145, opened 2017-05-26
Get PkgConf working with Debian clones like Ubuntu

jbearden0:
With this change PkgConf works with Ubuntu.


krimdomu:
Hi @jbearden0 ,

thanks for this patch. This will be included in the next release.

Issue #1115, opened 2016-11-29
Rex::Hardware::Network::Linux brctl command


LeMerP:
Function _get_bridge_devices_ should check if brctl command is available on the system before calling.

krimdomu:
Hi @LeMerP 

this is now merged into 1.6 line.

PR #1147, opened 2017-05-30
Check brctl command is available (resolves GH#1115)

perlancar:


krimdomu:
Hi @perlancar 

thanks for the patch, but i think this would break execution. If `brctl` is not available, this would stop execution instead of just having no information about the bridges.

So i think, to solve #1115 it would be better to just return an empyt hashref.

krimdomu:
Hi @perlancar 

i've just merged this patch to 1.6 line and modified it a bit.

PR #1148, opened 2017-06-01
mkfs for lvm volumes optional

samuelet:
Don't format lvm volumes if not explicitly requested using the fstype option.


krimdomu:
Hi @samuelet 

many thanks for the patch. I'm going to merge it into 1.5.1.

PR #1154, opened 2017-07-07
Fix Pkgconf for Debian

abeverley:
For reasons I do not understand, PkgConf for Debian adds 2 lines for each question, which causes errors with debconf-set-selections and is not as per the manual. It was me that wrote that line in originally, and I don't know why I did it or why it didn't fail before.

Anyway, this patch removes it, and allows it to function correctly.

### DIFF
--- a/lib/Rex/PkgConf/Debian.pm
+++ b/lib/Rex/PkgConf/Debian.pm
@@ -86,7 +86,6 @@ sub set_options {
       next;
     }
 
-    push @updated, $question;
     Rex::Logger::debug("Will set option $question: $value (type $type)");
     push @updated, "$pkg $question $type $value";
   }


krimdomu:
Hi @abeverley ,

thanks for the fix. I'm going to merge it into 1.5.1.

ferki:
FTR: it was originally merged into the `master` branch instead of the `development-1.x` branch, so we reverted it on the former, and cherry-picked it to the latter. Don't worry about this though, we would like to work on our branching model and contributing guide, it's just here for the sake of transparency about the changes :)

abeverley:
Great, thanks for letting me know. I seem to recall I was confused about the merge last time I stumbled on this problem, so that explains it!

PR #1159, opened 2017-07-17
Allow connection config to be altered in before hook

abeverley:
The before hook is called too late to be used for things like altering the connection configuration. This patch moves it earlier in the task.


krimdomu:
Hi @abeverley ,

thanks for the patch. I think this is a good idea. I'm going to merge it.

PR #1160, opened 2017-07-21
indent code to be properly formatted

yanick:


krimdomu:
Hi @yanick 

thanks for this patch.

PR #1164, opened 2017-07-28
fixed path bug in git scm checkout

jwbargsten:
The SCM package to handle git repos created repos in the wrong dir, if
confronted with path attribute consisting of multiple dirs. E.g. the
call in the  Rexfile:

  checkout 'my-repo', path => 'app/my-repo'

would result in a git repo that is checked out into

  app/app/my-repo

instead of

  app/my-repo

on the remote machine.


krimdomu:
Hi @jwbargsten 

thanks for your patches. I've just merged them into 1.6 line. This also merged #1173 because both pull requests contain same commits :)

Also many thanks for the test improvement.

Issue #943, opened 2016-02-18
delete_user should not return error if user does not exist


LeMerP:
Trying to remove a user that does not exist returns an error.

I would suggest to ether only return a warning, or to expand the function "delete_user" with an option, i.e. "ignore_error => 1".

Otherwise you cannot run a task more than once on a system.

PR #1166, opened 2017-08-04
Warn instead of dying when delete_user called on non existing user

laouji:
Original issue here: https://github.com/RexOps/Rex/issues/943

I thought it would also be useful to know the reason the user deletion failed (in exceptional cases), so added it to the original dying message.

First time contributor, so let me know if there is anything I should do differently 😄

### DIFF
--- a/lib/Rex/User/FreeBSD.pm
+++ b/lib/Rex/User/FreeBSD.pm
@@ -165,9 +165,11 @@ sub rm_user {
     $cmd .= " -r ";
   }
 
-  i_run $cmd . " -n " . $user;
-  if ( $? != 0 ) {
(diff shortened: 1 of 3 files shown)


krimdomu:
Hi @laouji 

many thanks for this patch and welcome :+1: . I'm going to merge it into 1.6.0 release.

ferki:
FTR: it was originally merged into the `master` branch instead of the `development-1.x` branch, so we reverted it on the former, and cherry-picked it to the latter (with resolving some merge conflicts). Don't worry about this though, we would like to work on our branching model and contributing guide, it's just here for the sake of transparency about the changes :)

PR #1172, opened 2017-09-28
Added support for Devuan as clone of Debian.

cwolfhugel:
Very simple change to add Devuan distribution as clone of Debian.


krimdomu:
Hi @cwolfhugel 

thanks for this patch. I've just merged it to 1.6 line.

PR #1178, opened 2017-11-03
only source ~/.profile if it really exists

jwbargsten:
Usually it is no problem to activate profile sourcing for users that don't have ~/.profile, because one can create ~/.profile. But for servers with sudo => TRUE and /root/.profile not present, all commands that should create or copy this file will fail. This is slightly annoying.

### DIFF
--- a/lib/Rex/Interface/Shell/Bash.pm
+++ b/lib/Rex/Interface/Shell/Bash.pm
@@ -85,7 +85,7 @@ sub exec {
   }
 
   if ( $self->{source_profile} ) {
-    $complete_cmd = ". ~/.profile >/dev/null 2>&1 ; $complete_cmd";
+    $complete_cmd = "[ \\! -r ~/.profile ] || . ~/.profile >/dev/null 2>&1 ; $complete_cmd";
   }
 
   if ( $self->{source_global_profile} ) {


ferki:
@jwbargsten: thanks for this PR too!

Are there any specific reason why a negative test of `[ \\! -r ~/.profile ]` was used instead of `[ -r ~/.profile ]`? Or perhaps instead even a Rex-specific postfix condition of `if is_file('~/.profile');`? I'm asking because I'm afraid I might be missing some point behind this one.

jwbargsten:
If `.profile` cannot be read `-r .profile`, the first statement before the `||` will be true (due to the negation) and the evaluation stops. If it can be read, the `[ ... ]` will evaluate to false and the stuff after `||` needs to be evaluated, which means that `.profile` will be sourced. This is exploiting the lazy evaluation of if clauses in bash.

ferki:
Sorry, it seems I was not specific enough. I meant using `[ -r ~/.profile ] &&` vs. `[ \\! -r ~/.profile ] ||`. But I think you still answered my question :)

I believe we can say that using the `is_readable()` Rex built-in would have higher overhead than letting the remote shell figure the same out, so I can drop that idea.

That being said, I personally find the `[ -r ~/.profile ] && . ~/profile` format easier to read/understand/maintain, so I might change it to that upon merge, if you don't mind.

Please feel free to let me know if you have objections or if I miss something :)

jwbargsten:
OK, this has been some time ago since I touched this, but even if I don't know exactly anymore what the background was, I concluded at that time: the whole `. ~/.profile` part caused an error in my Rex script and if I made this part return `0` as exit status (because non-existent `~/.profile` is a valid "condition" and should not exit as error), it worked. Here an example to clarify what I mean:

```bash
foo@bar:~ $ [ -r ~/.fake_profile ] && . ~/.fake_profile; echo $?
1
foo@bar:~ $ echo 'echo "sourced fake profile"' >~/.fake_profile
foo@bar:~ $ [ -r ~/.fake_profile ] && . ~/.fake_profile; echo $?
sourced fake profile
0
foo@bar:~ $ rm ~/.fake_profile
removed '/home/foo/.fake_profile'
foo@bar:~ $ [ -r ~/.fake_profile ] && . ~/.fake_profile; echo $?
1
foo@bar:~ $ [ ! -r ~/.fake_profile ] || . ~/.fake_profile; echo $?
0
foo@bar:~ $ echo 'echo "sourced fake profile"' >~/.fake_profile
foo@bar:~ $ [ ! -r ~/.fake_profile ] || . ~/.fake_profile; echo $?
sourced fake profile
0
```

ferki:
@jwbargsten: thanks for this contribution too! I now merged it manually into master branch and applied minor changes (perltidy + `[ -r ... ] &&` style testing).

PR #1179, opened 2017-11-03
Show all tasks

jwbargsten:
Currently it is not possible to list hidden tasks on the commandline. Now it is :)

### DIFF
--- a/bin/rex
+++ b/bin/rex
@@ -83,6 +83,7 @@ Rex can be used to:
     -Q     Really quiet: output nothing
   
     -T     List tasks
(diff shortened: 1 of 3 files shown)


ferki:
@jwbargsten: thanks again, it sounds useful and looks good! It might also make perltidy to complain, but don't worry about that too much as I can fix all such issues in one run after merging (and I really should set up automated tests for that :)).

krimdomu:
Hi @jwbargsten 
thanks also for thie one. I've merged it into master and it will be in the next release.

PR #1181, opened 2017-11-10
create_host: drop useless get_host invocation

alip:
The get_host call earlier sets `cur_host` so there's no point in
recalling `get_host` as the contents of the the hosts file has not
changed in the meantime.

As a side affect this seems to fix readding host entries already present
in the hosts file with different data (ip, aliases)

### DIFF
--- a/lib/Rex/Commands/Host.pm
+++ b/lib/Rex/Commands/Host.pm
@@ -137,10 +137,9 @@ sub create_host {
     $fh->close;
   }
   else {
-    my @host = get_host( $host, { file => $data->{file} } );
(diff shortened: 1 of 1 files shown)


ferki:
@alip: thanks, and sorry for the delay! It looks good to be included in the next release.

Could you help me understand the side effect part with an example, please? Or even better, through some additional tests in `t/hosts.t` perhaps?

ferki:
Thanks @alip again for the contribution! I'm still unsure about the side effect you mentioned, but the PR contents make sense, so I now merged it manually on top of current master.

PR #1186, opened 2018-01-02
Check if an iptables chain exists

frankiejol:
New method to check if a chain exists in a given iptables filter

### DIFF
--- a/lib/Rex/Commands/Iptables.pm
+++ b/lib/Rex/Commands/Iptables.pm
@@ -104,7 +104,7 @@ use Rex::Logger;
 
 @EXPORT = qw(iptables is_nat_gateway iptables_list iptables_clear
   open_port close_port redirect_port
-  default_state_rule);
+  default_state_rule chain_exists);
 
 sub iptables;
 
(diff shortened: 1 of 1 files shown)


lib/Rex/Commands/Iptables.pm
```diff
@@ -485,6 +485,49 @@ sub _iptables_list {
...
+
+
+=cut
+
+sub chain_exists {
+  my ( $table, $chain, @params ) = @_;
+  my $iptables = _get_executable( \@params );
+  my @lines    = run "$iptables-save";
+
+  return _chain_exists( $table, $chain, @lines );
```
ferki: I think it might be simpler and faster to check for the return code of `iptables [--table $table] --list $chain > /dev/null` command. If the chain exists it will be true, but otherwise false.

Perhaps even the built-in `iptables()` sub can be used for that too (but it probably needs to be wrapped in a try-catch block as it `die()`s in case of non-OK return codes).

ferki:
@frankiejol: thanks for the followup! It seems like there are some failing tests. Could you take a look at those, please?

If you add followup commits, please make sure the commit history remains readable (I guess two commits, one for the test and one for the new functionality would suffice here). Feel free to rebase on top of current master and force push your feature branch for that.

PR #1189, opened 2018-01-19
Fix `needs` with nested namespaces (fix #1188)

### LINKED ISSUE #1188: Task dependencies via `needs` is broken when modules have namespace separators
As [reported via a comment](https://github.com/RexOps/Rex/issues/1066#issuecomment-357228734) by @alip, calling `needs` seem to be broken at least for some use cases, and it at least happens since merging 7cf0ca4 (to fix #1066 on #1157).

I'm opening this issue in order to describe some details based on my own debugging, and also to collect more feedback before applying a fix.

I've used the following code while trying to replicate the reported behavior:

```bash
$ cat Rexfile
```
```perl
use Rex;

use Module;
use Nested::Module;

task 'test', sub {
  say q(This is a test);

  needs 'local_needs';
  needs Module 'module_needs';
  needs Nested::Module 'nested_needs';

  say q(end test);
};

task 'local_needs', sub {
  say q(I'm needed locally);
};
```
```bash
$ cat lib/Module.pm
```

```perl
package Module;

use Rex -base;

task 'module_needs', sub {
  say q(I'm needed in a module);
};

1;
```

```bash
$ cat lib/Nested/Module.pm
```

```perl
package Nested::Module;

use Rex -base;

task 'nested_needs', sub {
  say q(I'm needed in a nested module);
  needs main 'local_needs';
};

1;
```

This tests needing tasks from the same namespace, from a "simple" namespace, and from a "nested" namespace, as well as needing a task in the main namespace from one of the `use`d modules.

Trying to execute the `test` task shows the following (`-qw` is only used for reduced output before copy-pasting):

```
$ rex -qw test
This is a test
I'm needed locally
I'm needed in a module
end test
```

Which indicates that the originally mentioned use cases of 

> `needs Pkg 'task'` and `needs 'task'`

are working correctly, but it fails to handle cases when the module contains name separator(s).

Applying the following hotfix converts Perl's module namespace separators to Rex task namespace separators, and seems to solve the above for me:

```diff
diff --git a/lib/Rex/Commands.pm b/lib/Rex/Commands.pm
index 256b32f0..9a619fb7 100644
--- a/lib/Rex/Commands.pm
+++ b/lib/Rex/Commands.pm
@@ -1035,6 +1035,8 @@ sub needs {
 
   Rex::Logger::debug("need to call tasks from $self");
 
+  $self =~ s/::/:/g;
+
   my @tasks_to_run;
   if($self) {
     @tasks_to_run = $tl->get_all_tasks(qr{^\Q$self\E:[A-Za-z0-9_\-]+$});
```

ferki:
This PR fixes #1188 by converting any Perl module namespace separators into Rex task namespace separators.

@alip: could you try this and see if it solves your use case, please? I'm considering to include this in 1.6.0 as a last minute patch, so I thought some extra feedback would be nice :)

### DIFF
--- a/lib/Rex/Commands.pm
+++ b/lib/Rex/Commands.pm
@@ -1035,6 +1035,8 @@ sub needs {
 
   Rex::Logger::debug("need to call tasks from $self");
 
+  $self =~ s/::/:/g;
+
   my @tasks_to_run;
   if($self) {
     @tasks_to_run = $tl->get_all_tasks(qr{^\Q$self\E:[A-Za-z0-9_\-]+$});
--- a/t/needs.t
+++ b/t/needs.t
@@ -23,6 +23,20 @@ use Rex::Commands;
   1;
 }
 
+{
+  package Nested::Module;
+
+  use strict;
+  use warnings;
+
+  use Rex::Commands;
+
+  task "test", sub {
+    open( my $fh, ">", "test.txt" );
+    close($fh);
+  };
+}
+
 task "test", sub {
   needs MyTest;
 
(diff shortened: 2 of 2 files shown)


alip:
@ferki: Thanks! I'll test as soon as I can and report back.

PR #1198, opened 2018-06-11
Adds SLE 15 / Leap 15.0 support

LeMerP:
Replaced "SUSE Linux Enterprise Server" with "SUSE Linux Enterprise". It does not matter, if it is a server or desktop installation. It just does matter, that it is an enterprise installation.
Preferred to use "/etc/os-release" file (available since SLE 11.4) instead of deprecated "/etc/SuSE-release", that is not available anymore.

### DIFF
--- a/lib/Rex/Hardware/Host.pm
+++ b/lib/Rex/Hardware/Host.pm
@@ -139,7 +139,7 @@ sub get_operating_system {
     return "Debian";
   }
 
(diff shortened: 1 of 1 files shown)


krimdomu:
Hi @LeMerP 

thanks for your patch. I'm going to merge it into master branch, so it will be available in the next version.

Issue #1195, opened 2018-03-14
append_or_ammend_line overwrites symlink with real file when using sudo


mbroadhead:
```
$ rex -v
(R)?ex 1.6.0
```

Rexfile for the test:
```
use Rex;
use Rex::Helper::Path;

task symlink_tests => sub {
    require Test::More;
    Test::More->import('no_plan');
    my $tmp_dir = get_tmp_file();
    file $tmp_dir, ensure => 'directory';
    ok( is_dir($tmp_dir), 'create a temporary directory' );

    my $sub_dir1 = "$tmp_dir/sub1";
    my $sub_dir2 = "$tmp_dir/sub2";

    file $sub_dir1, ensure => 'directory';
    file $sub_dir2, ensure => 'directory';

    my $source_file = "$sub_dir1/source.txt";
    file $source_file, content => 'foo=bar';
    ok( is_file($source_file), 'file exists' );

    my $content = cat($source_file);
    is( $content, "foo=bar\n", 'file contents' );

    my $symlink = "$sub_dir2/symlink";
    symlink '../sub1/source.txt', $symlink;
    ok( is_symlink($symlink), 'create symlink' );

    my $changed;
    append_or_amend_line $symlink,
      line      => 'foo=new',
      regexp    => qr{^foo=},
      on_change => sub {
        $changed++;
      };

    is( $changed,      1,           'append_or_amend_line on_changed' );
    is( cat($symlink), "foo=new\n", 'append_or_amend_line line' );
    ok( is_symlink($symlink), 'is_symlink' );
};
```

Running locally without sudo (pass):
```
$ rex symlink_tests
18312 - <local> - 2018-03-14 13:29:30 - INFO => Running task symlink_tests on <local>
ok 1 - create a temporary directory
ok 2 - file exists
ok 3 - file contents
ok 4 - create symlink
ok 5 - append_or_amend_line on_changed
ok 6 - append_or_amend_line line
ok 7 - is_symlink
1..7
18307 - <local> - 2018-03-14 13:29:31 - INFO => All tasks successful on all hosts
```

Running locally with sudo (fail):
```
rex -s symlink_tests                                                                                                                                                                                          0
18424 - <local> - 2018-03-14 13:30:26 - INFO => Running task symlink_tests on <local>
ok 1 - create a temporary directory
ok 2 - file exists
18424 - <local> - 2018-03-14 13:30:28 - ERROR => Error executing task:
[...]

PR #1199, opened 2018-06-15
first check if file is a symlink and then use real file

krimdomu:
this is a fix for #1195


krimdomu:
Test for it can be found in https://github.com/RexOps/rex-build/commit/ac739ebe8ff6e8cd2038a49f13034f954af2fb18

PR #1204, opened 2018-09-06
delete_host: fix regex so it does not match on prefixes

alip:
Previously `delete_host` which is usually called by `host_entry` in case
the host entry is already present in the hosts file could end up
deleting more host entries than expected since the regular expression to
filter out the line to update in the hosts file matches on a prefix of
the hostname rather than the full hostname, e.g: with foo-11, foo-12,
foo-13 in hosts delete_host("foo-1") deletes all three host entries.

Change makes the regular expression stricter to match on a boundary and
applies escaping to the host variable.

### DIFF
--- a/lib/Rex/Commands/Host.pm
+++ b/lib/Rex/Commands/Host.pm
@@ -173,7 +173,7 @@ sub delete_host {
     my @content = $fh->read_all;
     $fh->close;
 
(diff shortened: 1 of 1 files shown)


krimdomu:
Hi @alip,
many thanks for this patch. And sorry for the long delay.
It will be included in the next release.

PR #1207, opened 2018-10-10
Add checksum testing to Rex::Test

### LINKED ISSUE #485: Support file checksum checking with Rex::Test

kiwiroy:
## Aim

Fix #485 

Only supporting three hash algorithms, currently *md5*, *sha1* and *sha256*, the former using existing code (`md5` from `Rex::Commands::MD5`), the latter two using `Digest::SHA`.

### DIFF
--- a/lib/Rex/Test/Base/has_checksum.pm
+++ b/lib/Rex/Test/Base/has_checksum.pm
@@ -1,0 +1,54 @@
+#
+# (c) Jan Gehring <jan.gehring@gmail.com>
+#
+# vim: set ts=2 sw=2 tw=0:
+# vim: set expandtab:
+
+package Rex::Test::Base::has_checksum;
+
+use strict;
+use warnings;
+
+# VERSION
+
+use Rex -base;
+use Digest::SHA;
+use Rex::Commands::MD5;
+use base qw(Rex::Test::Base);
+
+sub new {
+  my $that  = shift;
+  my $proto = ref($that) || $that;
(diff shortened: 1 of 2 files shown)


lib/Rex/Test/Base/has_checksum.pm
```diff
@@ -0,0 +1,54 @@
+#
+# (c) Jan Gehring <jan.gehring@gmail.com>
```
ferki: IANAL, but I'm pretty sure in this case the copyright is being held by the author of this file (and not by Jan :)).

lib/Rex/Test/Base/has_checksum.pm
```diff
@@ -0,0 +1,54 @@
...
+
+  bless( $self, $proto );
+
+  my ( $pkg, $file ) = caller(0);
+
+  return $self;
+}
+
+sub run_test {
+  my ( $self, $file, $checksum, $algo, $computed ) = (shift, shift, shift, shift);
```
ferki: I'm not entirely sure about using positional arguments here. Perhaps it might be better with a hash?

Also, there are 5 variables on the left hand side, but only 4 `shift`s on the right hand side. Perhaps `$computed` can be dropped since doesn't seem like used as an input argument?

t/file-tests.t
```diff
@@ -0,0 +1,54 @@
+use strict;
```
ferki: Bonus points for the tests! :100:

PR #1213, opened 2019-02-04
When the user doesn't exist yet, the $user_info{"uid"} variable

symacx:
Hi,

I'm sorry but I introduced a small «non initialised variable use» problem in the last pull request.  Could you please merge this fix ?
«
When the user doesn't exist yet, the $user_info{"uid"} variable doesn't exist and triggers an error. Add this test clause for this case.
»

Thank you,
Best
Olivier

### DIFF
--- a/lib/Rex/User/OpenBSD.pm
+++ b/lib/Rex/User/OpenBSD.pm
@@ -72,12 +72,18 @@ sub create_user {
     $cmd = "usermod ";
   }
 
-  if ( exists $data->{uid} ) {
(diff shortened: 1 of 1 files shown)


symacx:
@krimdomu 
Could you please commit this small change ?
Thank you,
Best

ferki:
@symacx: thanks for your contribution and sorry for the delay. I have now cherry-picked your commit into the master branch and applied some slight follow-up modifications (reword commit message, perltidy).

PR #1214, opened 2019-02-25
Fix placeholder matching for %h in ssh config

maxeaubrey:
Minor change to how hostnames with placeholders are matched in ssh config. Instead of relying on the assumption that this will always be in the format of `%h.` allows a plain `%h` (which is a valid entry, of course) by including the period in the regex.

Previously `Hostname %h` would be matched as a literal hostname.


ehuelsmann:
Thanks for your contribution

PR #1217, opened 2019-06-26
Added the Option 'continuous_read' to the documentation

elisdg:


ferki:
@elisdg: thanks for your contribution! I'm going to manually merge your commit as it is, and I'll add the rewording and the realigning of the block as separate followup commits.

ferki:
Merged!

PR #1219, opened 2019-07-26
* Replace dynamic loading through 'eval()' with Module::Runtime

ehuelsmann:
Module::Runtime does error checking and input sanitizing,
which be better make use of than try to do ourselves.

Besides, the pattern using eval() warns on critic tests,
which we're resolving through this commit too.

### DIFF
--- a/dist.ini
+++ b/dist.ini
@@ -48,6 +48,7 @@ Net::SSH2 = 0
 [Prereqs]
 perl = 5.008008
 YAML = != 1.25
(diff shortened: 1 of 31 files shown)


ehuelsmann:
Needs to be merged after #1221 ; when rebasing, also need to remove the 'eval return value needs to be checked' policy being excluded from testing.

ehuelsmann:
Work from this branch will be cherry-picked onto the perlcritic brach.

PR #1220, opened 2019-07-26
* Fix critic 'select-as-sleep' failure

ehuelsmann:
In an effort to remove Critic complaints (and to clean up the code base),
use modules readily available for the specific purposes.

### DIFF
--- a/dist.ini
+++ b/dist.ini
@@ -47,6 +47,7 @@ Net::SSH2 = 0
 
 [Prereqs]
 perl = 5.008008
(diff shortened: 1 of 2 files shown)


ehuelsmann:
Needs to be merged after #1221 ; when rebasing, also need to remove the 'select as sleep' policy being excluded from testing.

ehuelsmann:
This work ends up on the perlcritic branch through cherry-picking.

PR #1223, opened 2019-08-18
Test for code tidyness

ferki:
As a maintainer I would like to ensure consistent formatting of code in order
to be able to offload related work (formatting, validating, feedback) to
machines, iltimately making maintanance easier.

Test::PerlTidy doesn't check files under bin because they don't have an
extension, so perlcritic is used as a workaround, as early as possible (i.e.
don't even try to build the testing environment if we already know there's a
violation).

For Test::PerlTidy to work, it needs the `.perltidyrc` in the build.
Unfortunately `.../.perltidyrc` doesn't work to discover the "nearest" config
file, but anyway, it feels like a good idea to make sure the same config is
used in various environments (instead of incremental lookup accidentally
discovering some other config). So for that, GatherDir, PruneCruft, and
MANIFEST.SKIP had to be modified.

Autogenerated files and directories are excluded from tidyness checks.

Then, as a last step, the latest perltidy version was used to format all code. This fixed existing formatting violations, as well as applied some formatting rules that was made default recently.

Edit: this is also salvaging some more code from #1005, and builds up on previous work by @ehuelsmann on #1222.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,8 +8,10 @@ cache:
     - ${PERLBREW_ROOT}/perls/${PERLBREW_PERL}/lib
     - ${PERLBREW_ROOT}/perls/${PERLBREW_PERL}/bin
 before_install:
+  - cpanm Perl::Critic
+  - perlcritic --quiet --single-policy=RequireTidyCode bin
   - dzil authordeps --missing | cpanm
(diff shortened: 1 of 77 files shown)


ferki:
@yko: thanks for your insights earlier on IRC! I now updated tidyness checks to be author tests.

PR #1225, opened 2019-08-26
Perlcritic tests

ferki:
This is still heavily work in progress, but overall I believe there are not so complicated fixes for all the violations at the current severity level, so I try to include them in this PR. If things get complicated or we decide the changes are too risky, we can still just mark the known violations as known ones for documentation purposes.

I preferred writing `## no critic ...` style comments to make sure we don't silently introduce further violations of a given policy (which is possible if we would just disable the policy completely in `.perlcriticrc`).

Depending on the decisions when the full picture is already known, I'll probably rewrite the commit messages before merging to make the history easier to understand.

### DIFF
--- a/lib/Rex/Commands.pm
+++ b/lib/Rex/Commands.pm
@@ -317,7 +317,7 @@ sub task {
     push( @_, "" );
   }
 
-  no strict 'refs';
+  no strict 'refs'; ## no critic ProhibitNoStrict
   no warnings;
   push( @{"${class}::tasks"}, { name => $task_name_save, code => $_[-2] } );
   use strict;
@@ -329,7 +329,7 @@ sub task {
   if (!$class->can($task_name_save)
     && $task_name_save =~ m/^[a-zA-Z_][a-zA-Z0-9_]+$/ )
   {
-    no strict 'refs';
+    no strict 'refs'; ## no critic ProhibitNoStrict
     Rex::Logger::debug("Registering task: $task_name");
     my $code = $_[-2];
     *{"${class}::$task_name_save"} = sub {
@@ -1087,7 +1087,7 @@ sub needs {
   }
 
   if ( $caller_pkg && ( $caller_pkg eq "Rex::CLI" || $caller_pkg eq "main" ) ) {
-    no strict 'refs';
+    no strict 'refs'; ## no critic ProhibitNoStrict
     *{"main::needs"} = \&needs;
     use strict;
   }
--- a/lib/Rex/Commands.pm
+++ b/lib/Rex/Commands.pm
@@ -1059,7 +1059,7 @@ sub needs {
   if ($self) {
     my $suffix = $self;
     $suffix =~ s/::/:/g;
-    @args = map { $_ = "$suffix:$_" } @args;
+    @args = map { "$suffix:$_" } @args;
   }
 
   for my $task (@tasks_to_run) {
--- a/lib/Rex/Commands.pm
+++ b/lib/Rex/Commands.pm
@@ -1202,7 +1202,7 @@ With the LOCAL function you can do local commands within a task that is defined 
 
 =cut
 
-sub LOCAL (&) {
+sub LOCAL (&) { ## no critic ProhibitSubroutinePrototypes
   my $cur_conn      = Rex::get_current_connection();
   my $local_connect = Rex::Interface::Connection->create("Local");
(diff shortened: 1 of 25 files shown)


ferki:
Since it's taking too long, I've decided to limit the scope of this PR to just simply enable the perlcritic tests, and mark known violations as accepted. A followup PR is in progress with all the fixes (or that can be done in smaller chunks too, one PR per violated policies).

lib/Rex/Commands.pm
```diff
@@ -1059,7 +1059,7 @@ sub needs {
   if ($self) {
     my $suffix = $self;
     $suffix =~ s/::/:/g;
-    @args = map { $_ = "$suffix:$_" } @args;
+    @args = map { $_ = "$suffix:$_" } @args; ## no critic ProhibitMutatingListFunctions
```
ehuelsmann: The code as it is there now, is equivalent to:

```perl
@args = map { "$suffix:$_" } @args;
```

due to the fact that the return value of `map` is assigned to the variable being iterated over. I think it's better, even for the scope of this PR not to suppress but to fix this instance.

lib/Rex/Commands/Run.pm
```diff
@@ -255,7 +255,7 @@ sub run {
 
     if ( $args && ref($args) eq "ARRAY" ) {
       my $quoter = Net::OpenSSH::ShellQuoter->quoter( $exec->shell->name );
-      $cmd = "$cmd " . join( " ", map { $_ = $quoter->quote($_) } @{$args} );
+      $cmd = "$cmd " . join( " ", map { $_ = $quoter->quote($_) } @{$args} ); ## no critic ProhibitMutatingListFunctions
```
ehuelsmann: `@args` isn't used in the remainder of the function, which leads to a similar argument as above:

The code as it is there now, is equivalent to:

```perl
@args = map { "$suffix:$_" } @args;
```

due to the fact that the mutated value of `@args` isn't used. I think it's better, even for the scope of this PR not to suppress but to fix this instance.

lib/Rex/Helper/Run.pm
```diff
@@ -146,7 +146,7 @@ sub i_exec {
   my $exec   = Rex::Interface::Exec->create;
   my $quoter = Net::OpenSSH::ShellQuoter->quoter( $exec->shell->name );
 
-  my $_cmd_str = "$cmd " . join( " ", map { $_ = $quoter->quote($_) } @args );
+  my $_cmd_str = "$cmd " . join( " ", map { $_ = $quoter->quote($_) } @args ); ## no critic ProhibitMutatingListFunctions
```
ehuelsmann: `@args` isn't used in the remainder of the function, which leads to a similar argument as above:

The code as it is there now, is equivalent to:

```perl
@args = map { "$suffix:$_" } @args;
```

due to the fact that the mutated value of `@args` isn't used. I think it's better, even for the scope of this PR not to suppress but to fix this instance.

lib/Rex/Helper/Run.pm
```diff
@@ -157,7 +157,7 @@ sub i_exec_nohup {
...
```
ehuelsmann: `@args` isn't used in the remainder of the function, which leads to a similar argument as above:

The code as it is there now, is equivalent to:

```perl
@args = map { "$suffix:$_" } @args;
```

due to the fact that the mutated value of `@args` isn't used. I think it's better, even for the scope of this PR not to suppress but to fix this instance.

ferki:
Most of the fixes are already done and ready to be merged after this one from the [perlcritic-fix](https://github.com/RexOps/Rex/commits/perlcritic-fix) branch, including for the ones you highlighted.

If we really do want to mix marking and fixing violations here, perhaps it's best to draw the line at the "obvious" fixes, and include these three right away:
 - 823fad300873a2cb60ed03874a2997c92cb77170 Remove unnecessary subroutine prototype
 - da3f5537277204a18f03502ec2cd91a4d43cc71c Fix critic for ProhibitMutatingListFunctions
 - 91bc3336821bf28037d0cc6a58020d57ca715a06 Fix critic for ProhibitSleepViaSelect

@ehuelsmann: if you agree with those three commits, I can update this PR with them (fix instead of mark), leaving the rest for another PR(s).

ehuelsmann:
Completely agreed with the three commits!

PR #1237, opened 2019-10-20
Remove redundant argument to rsync command

sdondley:
The sprintf function used to populate the rsync command was getting passed two place-holder values when only one is needed.


ferki:
Good catch @sdondley, thanks for contributing!

PR #1242, opened 2019-10-29
Fix tests on Dragonfly BSD

ferki:
Based on [CPAN Testers results for Rex-1.6.0 on Dragonfly BSD](http://www.cpantesters.org/distro/R/Rex.html?grade=1&perlmat=2&patches=2&oncpan=2&distmat=2&perlver=ALL&osname=dragonfly&version=1.6.0), the exit codes need to be adjusted.


ferki:
Hmm, actually, I'll try an alternative, more generic approach first.

ferki:
Closing unmerged in favor of #1243.

PR #1243, opened 2019-10-30
Fix known failures from CPAN Testers

ferki:
This PR fixes all known, reproducible and real failures reported for Rex-1.6.0 via CPAN Testers (thanks!).

By "real", I mean failures that is not caused by the perl-5.8.8 vs perl-5.8.9 [issue](https://github.com/RexOps/Rex/issues/578).

This includes the following:
  
- Fix summary testing failures ([example report](http://www.cpantesters.org/cpan/report/c3e6fae4-f059-11e8-afbf-aec3b07506ca))

    This is one of the most common failure. When testing the run command in a task, that is expected to fail, the exit code was either `1` or `2` depending on OS, based on various implementations of `ls`. By using `type` and `cat` instead, I expect the exit code will be always `1`. This should fix the test suite on DragonFly BSD in particular (replacing #1242), and should be valid for all non-Linux systems too, allowing a simplification of the testing code.
  
- Use perl to query environment ([example report](http://www.cpantesters.org/cpan/report/c3e6fae4-f059-11e8-afbf-aec3b07506ca), same as above)

    Some environments does not seem to have a `printenv` command, but `perl` should be available, since we're already running the test suite. That might also fix environment tests on Windows, but I'd prefer dealing with Windows cases separately, and use this PR to only focus on the already reported failures.
  
- Require recent Test::UseAllModules ([example report](http://www.cpantesters.org/cpan/report/1baadcfa-f04b-11e8-9f72-eeb84b3d96e1))

    Another quite common failure, but was only temporary. It won't hurt to require a minimum version here, while "handling dependencies" can be a separate topic. See Test-More/test-more#685 for details.

If this passes CI tests, I plan to merge and release 1.6.0_02 aka 1.7.0-rc2, to gather more feedback from CPAN Testers.

### DIFF
--- a/dist.ini
+++ b/dist.ini
@@ -63,6 +63,7 @@ YAML = != 1.25
 
 [Prereqs / BuildRequires]
 Test::Pod = 0
+Test::UseAllModules = 0.15
 
 [Prereqs / DevelopRequires]
 Test::PerlTidy = 0
--- a/t/env.t
+++ b/t/env.t
@@ -7,6 +7,6 @@ $::QUIET = 1;
 SKIP: {
   skip 'Do not run tests on Windows', 1 if $^O =~ m/^MSWin/;
 
-  my $s = run( "printenv REX", env => { 'REX' => 'XER' } );
+  my $s = run( q(perl -e 'print $ENV{REX}'), env => { 'REX' => 'XER' } );
   like( $s, qr/XER/, "run with env" );
 }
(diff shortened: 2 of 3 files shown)


ehuelsmann:
@ferki why not simply deprecate any 5.8 version and require 5.10 at least?

ferki:
@ehuelsmann: we already did that on c4e35d62ed416e2a5f58713bcef9c10ef544e72b as part of #1222 ;) That's why I say perl-5.8.8/5.8.9 failures aren't "real" failures, CPAN Testers will start ignoring them now, and simply report results as `NA`.

PR #1256, opened 2019-12-06
Configurable sleep time for fork managers

ferki:
I'm opening this PR as draft to share the idea early and to get testing feedback early too.

Based on profiling results, `t/summary.t` is slow mostly on the tests utilizing `Parallel::ForkManager`, and in turn that is caused by relying on the default value for `waitpid_blocking_sleep` which is 1 second. Setting it to a lower value speeds the test up considerably, and real-worls task executions should see a performance boost too.

A comparable property of the built-in `Rex::Fork::Manager` is the sleep time in `wait_for`, which is hard-coded to be 0.1 second.

My initial idea was to make those timings the same for consistency. 0.1 and 0.001 second seems to work fine for both, but it's also possible to set it to 0 for `Parallel::ForkManager` which is even more efficient but may have caveats (see [BLOCKING CALLS](https://metacpan.org/pod/Parallel::ForkManager#BLOCKING-CALLS) for more details).

So currently I think it's best to make it into a configurable option which has a default value either globally or per forking backend, but it's also overridable by the users on demand.

### DIFF
--- a/lib/Rex/Config.pm
+++ b/lib/Rex/Config.pm
@@ -61,7 +61,7 @@ our (
   $set_no_append,            $use_net_openssh_if_present,
   $use_template_ng,          $use_rex_kvm_agent,
   $autodie,                  $task_chaining_cmdline_args,
-
+  $waitpid_blocking_sleep_time,
 );
 
 # some defaults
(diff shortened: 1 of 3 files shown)


krimdomu:
Hi ferki,
I think i had to use this sleep because on slow network it had problems without it. so making it configurable should be a good choise.

ferki:
@krimdomu: thanks for the extra details, and for validating the approach! :+1:

ferki:
On my laptop this results in almost 75% reduction of runtime for `t/summary.t` (9765ms vs. 2562ms).

So far I mostly done local sequential testing, so I intend to run it in more cases over various networks (slow/fast), and also with various amount of parallelity.

@krimdomu: if you can review it, I'd be interested in making sure I got the "make it configurable" part right, at least for an initial version. It works with `set waitpid_blocking_sleep_time => 0.001` format at least.

ferki:
Thanks for the feedback, @krimdomu!

PR #1259, opened 2019-12-09
Use Symbol instead of `no strict refs`

ferki:
Turns out in many, if not all cases, using the core `Symbol` module makes it possible to avoid using `no strict refs`.

This PR is a work in progress to clean all cases up. During the WIP phase I convert cases one-by-one on each commit to make it easier to test and review. Before merging I plan to squash them all into one, or at least group them together somehow, and write proper useful commit messages.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -17,6 +17,7 @@ Revision history for Rex
 
  [REVISION]
  - Use author tests to check tidiness of bin files
+ - Use Symbol to manipulate Perl symbols
 
 1.11.0 2020-06-05 Ferenc Erki <erkiferenc@gmail.com>
  [BUG FIXES]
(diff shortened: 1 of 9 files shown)


ferki:
This is a collection of work done back in 2019 between 26 Aug and 9 Dec, collapsed into a single commit as I see no value in keeping the essentially same code conversion scattered across 14 commits.

This removes most (but unfortunately not all) `no strict refs` remarks with cleaner code containing less magic. I haven't had a reliable way to test the remaining ones (HAL, SMBios, etc.), so those should be fixed in another go.

I was running rex with this patchset for a good while now, and I'd like to take a bash at merging at least the already completed bits. It probably would be a good idea to do an early development release containing this to get more feedback via CPAN Testers.

PR #1260, opened 2019-12-09
Support for Void Linux

### LINKED ISSUE #1258: Add Void Linux support
Void Linux ships Rex, but unfortunately it's not yet possible to use Rex to manage Void Linux systems. We'd be happy to assist contributors to change this :)

Getting it done would basically consist of the following things:
 - teach rex about how to detect Void Linux systems (hint: `Rex::Hardware::Host::get_operating_system()` and `Rex::Commands::Gather:is_voidlinux()`)
 - if Void Linux is detected, choose the proper package and service management modules (`Rex::Service` and `Rex::Pkg`)
 - Void Linux uses `runit` for service management and `xbps` for package management; support for those needs to be implemented in corresponding new modules, e.g. `Rex::Service::Void`/`Rex::Service::Runit` and `Rex::Pkg::Void`/`Rex::Pkg::Xbps`.

Depending on maintainability/sustainability factors, I can imagine this both as a separate extension module for Rex or as a built-in module in the core distribution.

leahneukirchen:
The first beginnings of Void Linux support in Rex.

Mapping runit to Rex::Service is not completely straight-forward, as services need to be enabled before they are started, so we merge both operations currenty. (This does not support taking a service down using a `down` file in runit.)

Closes #1258.

### DIFF
--- a/lib/Rex/Commands/Gather.pm
+++ b/lib/Rex/Commands/Gather.pm
@@ -45,7 +45,7 @@ use vars qw(@EXPORT);
 
 @EXPORT = qw(operating_system_is network_interfaces memory
   get_operating_system operating_system operating_system_version operating_system_release
-  is_freebsd is_netbsd is_openbsd is_redhat is_linux is_bsd is_solaris is_suse is_debian is_mageia is_windows is_alt is_openwrt is_gentoo is_fedora is_arch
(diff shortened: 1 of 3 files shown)


leahneukirchen:
I didn't realize Github draft PR had to be announced on creation, unlike Gitlab.

Anyway, fixed the perltidy stuff and removed the NYI things; adding custom Void repos is rather special, so this works well without.

ferki:
Thanks @leahneukirchen for the updated PR, LGTM! :+1:

PR #1269, opened 2020-02-02
Document embedded templates

ferki:
This PR adds initial documentation for embedded templates.

[Reference](https://www.rexify.org/docs/rex_book/the_rex_dsl/using_templates.html)

@gonzalo-radio: is this already helpful for your use case?


gonzalo-radio:
@ferki Thank you very much for improve this section of the documentation.

Issue #1274, opened 2020-03-06
account() fails with 'comment' including non-ASCII character when `use utf8;` is in effect


dboehmer:
My last name *Böhmer* contains a German umlaut *ö*. Setting the `comment` of my user account via SSH fails.

Example:

```perl
#!/usr/bin/env perl

use strict;
use warnings;

use utf8;
use Rex;

user 'root';

task test => sub {
    account 'daniel' => (
        comment     => "Daniel Böhmer",
        create_home => FALSE,
    );
};
```

Ran it like:

    $ rex -d -H $HOST test

ferki:
@dboehmer: could you also attach your immediate error message to the opening post as well, please? The one that complains about the JSON (de)serialization.

My initial triaging suggests that the behavior might be slightly different depending on which JSON module is loaded by `JSON::MaybeXS`, but I still couldn't reproduce your case.

Following that lead, please also post which version of which JSON module gets loaded on the system where you are running rex. If all's well, then I would be able to reproduce the same results based on that info.

dboehmer:
I hope you had a pleasant journey back home:-)

Seems like the error persists although the relevant modules are up to date:

```
$ rex -H $HOST test
[2020-03-10 17:21:11] INFO - Running task test on XXX_HOST_XXX
[2020-03-10 17:21:13] ERROR - Error executing task:
[2020-03-10 17:21:13] ERROR - malformed UTF-8 character in JSON string, at character offset 136 (before "\x{fffd}r","/home/da...") at /home/daniel/perl5/perlbrew/perls/perl-5.30.0-threads-buster/lib/site_perl/5.30.0/Rex/User/Linux.pm line 353, <> line 45.

[2020-03-10 17:21:13] ERROR - 1 out of 1 task(s) failed:
[2020-03-10 17:21:13] ERROR -   test failed on XXX_HOST_XXX
[2020-03-10 17:21:13] ERROR -           malformed UTF-8 character in JSON string, at character offset 136 (before "\x{fffd}r","/home/da...") at /home/daniel/perl5/perlbrew/perls/perl-5.30.0-threads-buster/lib/site_perl/5.30.0/Rex/User/Linux.pm line 353, <> line 45.
$ 
$ cpanm JSON JSON::XS JSON::MaybeXS Rex
JSON is up to date. (4.02)
JSON::XS is up to date. (4.02)
JSON::MaybeXS is up to date. (1.004000)
Rex is up to date. (1.8.2)
```

[...]

PR #1275, opened 2020-03-28
Add option to control encoding of files written by Rex

ferki:
Rex writes files to the managed system via `Rex::Interface::File` modules. This mechanism is used both for `file $file, content => '';` and `Rex::FS::File` as well.

Regarding the encoding of the written file, Rex behaves identically to Perl currently ("it's just Perl"). However depending on the encoding of the input, this may lead to unexpected side effects, e.g. when Rex carries out its tasks by using temporary files on the managed systems (see for example #1274).

This PR adds an option to control the encoding of files written by Rex as a first step to give more control to the users in those cases while still preserving the current behaviour for backwards compatibility.

With the option in place, it will be possible to establish a default value for this setting later, and/or to control the behaviour via a feature flag.

Example code:
```
use Rex;

set file_write_encoding => 'UTF-8';
```

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -8,6 +8,7 @@ Revision history for Rex
  [DOCUMENTATION]
 
  [ENHANCEMENTS]
+ - Add option to control encoding of files written by Rex
 
  [MAJOR]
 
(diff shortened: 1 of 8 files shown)


ferki:
Based on feedback over IRC from @krimdomu and @ehuelsmann, I close this PR in favor of #1276.

The main problem with this approach is that the setting has an effect globally, and that breaks pre-existing modules which already do their own encoding (in other words, it causes double encoding).

PR #1276, opened 2020-03-29
Add write_utf8_files feature flag (close #1274)

### LINKED ISSUE #1274: account() fails with 'comment' including non-ASCII character when `use utf8;` is in effect

ferki:
This PR is a reiteration of #1275 with a different approach.

The idea here is to check the utf8 flag of the string to be written to a file, and encode it as needed.

This "check and encode" step is controlled by the `(no_)write_utf8_files` feature flag to allow complete opt-in/opt-out choice over it.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -8,6 +8,7 @@ Revision history for Rex
  [DOCUMENTATION]
 
  [ENHANCEMENTS]
+ - Add write_utf8_files feature flag to control encoding of files written by Rex
 
  [MAJOR]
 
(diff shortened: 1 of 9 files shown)


ferki:
@krimdomu: many thanks for the review! :+1: 

I plan to merge this soon and do a development release just to start gathering CPAN Testers feedback from various platforms and perl versions. That should also make community testing easier.

PR #1281, opened 2020-04-04
Document configuration functions

ferki:
This PR aims to document all configuration options supported by Rex as codified in `Rex::Config`.

I'm publishing this ongoing work early as a draft PR to allow others to proofread my descriptions and possibly to get help from others in this effort.

Most of the POD has been autogenerated in a way that should be easier to fill out as a human on one hand, and makes the POD coverage tests of `Devel::Cover` happy on the other.

In the short time I already worked on this, I already found this to be a good exercise to check for validity of those options and to spot potential bugs and improvement possibilities.

### DIFF
--- a/lib/Rex/Config.pm
+++ b/lib/Rex/Config.pm
@@ -269,14 +269,6 @@ sub get_use_cache {
   return $use_cache;
 }
 
-sub get_sudo_without_locales {
-  return $sudo_without_locales;
-}
-
-sub get_sudo_without_sh {
-  return $sudo_without_sh;
-}
-
 sub set_openssh_opt {
   my ( $class, %opt ) = @_;
 
@@ -300,9 +292,17 @@ sub set_sudo_without_locales {
   $sudo_without_locales = shift;
 }
 
+sub get_sudo_without_locales {
+  return $sudo_without_locales;
+}
+
 sub set_sudo_without_sh {
   my $class = shift;
   $sudo_without_sh = shift;
+}
+
+sub get_sudo_without_sh {
+  return $sudo_without_sh;
 }
 
 sub set_executor_for {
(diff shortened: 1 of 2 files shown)


lib/Rex/Config.pm
```diff
@@ -126,6 +184,16 @@ sub get_check_service_exists {
   return $check_service_exists;
 }
 
+=head2 set_register_cmdb_template
+
+=head2 get_register_cmdb_template
+
+Sets and gets the value of the C<$register_cmdb_template> configuration variable.
+
+This controls whether Rex should make L<CMDB|Rex::CMDB> data available to be uses in templates as variables.
```
ehuelsmann, line 193: s/to be uses/to be used/

lib/Rex/Config.pm
```diff
@@ -99,6 +127,16 @@ sub get_use_rex_kvm_agent {
   return $use_rex_kvm_agent;
 }
 
+=head2 set_use_template_ng
+
+=head2 get_use_template_ng
+
+Sets and gets the value of the C<$use_template_ng> configuration variable.
+
+This controls whether Rex should use L<Rex::Template::NG> to render templates.
```
ehuelsmann, line 136: As `Rex::Template::NG` seems to be undocumented, what does this mean?

ferki: `Rex::Template::NG` should get its own docs for the true meaning of the effects.

Strictly speaking, the config option here is about whether it's on or off. So in the context of `Rex::Config`, I just tried to stick to document that behavior (but already linked to `Rex::Template::NG` despite having no docs there yet).

lib/Rex/Config.pm
ferki: Well spotted, thanks!

Issue #1299, opened 2020-04-16
Rex hangs when trying to modify file without proper permission (macos)


sdondley:
This task
```
task edit_file => sub {
  file "/Users/me/rex/test2/file",
    content => 'change me';
};
```
will hang if the local user does not have permission to edit the file.

sdondley:
debug output:
```
[2020-04-15 23:50:04] DEBUG - This is Rex version: 1.8.2
[2020-04-15 23:50:04] DEBUG - Command Line Parameters
[2020-04-15 23:50:04] DEBUG - 	d = 1
[2020-04-15 23:50:04] DEBUG - Creating lock-file (Rexfile.lock)
[2020-04-15 23:50:04] DEBUG - Loading Rexfile
[2020-04-15 23:50:04] DEBUG - Enabling task_chaining_cmdline_args feature
[2020-04-15 23:50:04] DEBUG - Activating new template engine.
[2020-04-15 23:50:04] DEBUG - Disabling usage of a tty
[2020-04-15 23:50:04] DEBUG - Activating autodie.
[2020-04-15 23:50:04] DEBUG - Using Net::OpenSSH if present.
[2020-04-15 23:50:04] DEBUG - Add service check.
[2020-04-15 23:50:04] DEBUG - Setting set() to not append data.
[2020-04-15 23:50:04] DEBUG - Registering CMDB as template variables.
[2020-04-15 23:50:04] DEBUG - activating featureset >= 0.51
[2020-04-15 23:50:04] DEBUG - activating featureset >= 0.40
[2020-04-15 23:50:04] DEBUG - activating featureset >= 0.35
[2020-04-15 23:50:04] DEBUG - activating featureset >= 0.31
[2020-04-15 23:50:04] DEBUG - Creating new distribution class of type: Base
[2020-04-15 23:50:04] DEBUG - new distribution class of type Rex::TaskList::Base created.
[2020-04-15 23:50:04] DEBUG - Returning existing distribution class of type: Rex::TaskList::Base
[2020-04-15 23:50:04] DEBUG - Creating task: Blah:Boo:example
[2020-04-15 23:50:04] DEBUG - Found Net::OpenSSH and Net::SFTP::Foreign - using it as default
[2020-04-15 23:50:04] DEBUG - Registering task: Blah:Boo:example
[2020-04-15 23:50:04] DEBUG - Returning existing distribution class of type: Rex::TaskList::Base
[...]

PR #1300, opened 2020-04-16
prevent hanging when renaming file

sdondley:
For https://github.com/RexOps/Rex/issues/1299


ferki:
I'm afraid a full fix would be more involved than forcing the operation unconditionally on a single FS interface. See my [comment on the issue](https://github.com/RexOps/Rex/issues/1299#issuecomment-614416463).

Let's discuss the desired behavior of Rex and implementation details first over there.

PR #1301, opened 2020-04-17
Add in pkg globbing support and add it for FreeBSD.

VVelox:
While apt will do this automatically, pkg does not. Requires appending -g.

This adds in support for checking if * is passed and then if it is it checks that OS's module to see if there is a specific glob add/remove command and uses that.

This does appear to require the OS support module also include the methods remove and is_installed exist with glob support as well, at least in the case of FreeBSD.

The use case for this is basically installing like all of php72-* or the like when installing a new web environment instead of having to list each one you want to install.

### DIFF
--- a/lib/Rex/Pkg/Base.pm
+++ b/lib/Rex/Pkg/Base.pm
@@ -10,6 +10,7 @@ use strict;
 use warnings;
 use Rex::Helper::Run;
 use Rex::Interface::Exec;
+use Net::OpenSSH::ShellQuoter;
 
 # VERSION
 
@@ -66,11 +67,25 @@ sub update {
   my $env     = $option->{'env'}     || ();
 
   Rex::Logger::debug( "Installing $pkg" . ( $version ? "-$version" : "" ) );
-  my $cmd = sprintf $self->{commands}->{install}, $pkg;
+  my $cmd;
+  if ( ( $pkg =~ /\*/ )
+    && defined( $self->{commands}->{install_glob} ) )
+  {
 
-  if ( exists $option->{version} ) {
-    $cmd = sprintf $self->{commands}->{install_version}, $pkg,
-      $option->{version};
+    # quote the pkg name so it won't error when ran
+    my $exec   = Rex::Interface::Exec->create;
+    my $quoter = Net::OpenSSH::ShellQuoter->quoter( $exec->shell->name );
+    $pkg = $quoter->quote($pkg);
+    $cmd = sprintf $self->{commands}->{install_glob}, $pkg;
+  }
+  else {
+    $cmd = sprintf $self->{commands}->{install}, $pkg;
+
+    # not compatible with globs, so skip over this
+    if ( exists $option->{version} ) {
+      $cmd = sprintf $self->{commands}->{install_version}, $pkg,
+        $option->{version};
+    }
   }
 
   my $f = i_run $cmd, fail_ok => 1, env => $env;
@@ -135,7 +150,20 @@ sub remove {
   my ( $self, $pkg ) = @_;
 
   Rex::Logger::debug("Removing $pkg");
-  my $cmd = sprintf $self->{commands}->{remove}, $pkg;
+  my $cmd;
+  if ( ( $pkg =~ /\*/ )
+    && defined( $self->{commands}->{remove_glob} ) )
+  {
+
+    # quote the pkg name so it won't error when ran
+    my $exec   = Rex::Interface::Exec->create;
+    my $quoter = Net::OpenSSH::ShellQuoter->quoter( $exec->shell->name );
+    $pkg = $quoter->quote($pkg);
+    $cmd = sprintf $self->{commands}->{remove_glob}, $pkg;
+  }
+  else {
+    $cmd = sprintf $self->{commands}->{remove}, $pkg;
+  }
 
   my $f = i_run $cmd, fail_ok => 1;
 
--- a/lib/Rex/Pkg/FreeBSD.pm
+++ b/lib/Rex/Pkg/FreeBSD.pm
@@ -15,6 +15,7 @@ use Rex::Helper::Run;
 use Rex::Commands::File;
 use Rex::Pkg::Base;
 use base qw(Rex::Pkg::Base);
+use Net::OpenSSH::ShellQuoter;
 
 sub new {
   my $that  = shift;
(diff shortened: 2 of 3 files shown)


lib/Rex/Pkg/FreeBSD.pm
```diff
@@ -56,10 +58,14 @@ sub bulk_install {
...
-  my $pkg_version = $pkg_found->{"version"};
+  my $pkg_version = '';
 
-  return $self->SUPER::remove("$pkg-$pkg_version");
+  if ( $pkg !~ /\*/ ) {
+    my ($pkg_found) = grep { $_->{"name"} eq "$pkg" } $self->get_installed();
+    $pkg_version = '-' . $pkg_found->{"version"};
+  }
+
+  return $self->SUPER::remove("$pkg$pkg_version");
```
ferki, line 68: So if it's not a package glob, prefix the version string with `-`, otherwise leave it as it is.

Just for my education, could you provide an example about how package and package glob versions look like on FreeBSD, please?

lib/Rex/Pkg/FreeBSD.pm
```diff
@@ -69,9 +75,21 @@ sub is_installed {
   Rex::Logger::debug(
     "Checking if $pkg" . ( $version ? "-$version" : "" ) . " is installed" );
 
+  # only need -g if pkg is a glob
+  my $glob_arg = '';
+  if ( $pkg =~ /\*/ ) {
+
+    # back quote asterisks so the command can actually be used
+    $pkg =~ s/\*/\\\*/g;
```
ferki, line 83: Referencing the quoting method note from above.

lib/Rex/Pkg/FreeBSD.pm
```diff
@@ -69,9 +75,21 @@ sub is_installed {
...
+    $glob_arg = ' -g ';
```
ferki, line 84: Are both leading and trailing spaces are needed for this to work?

Nit: I believe the trailing one would cause a double space below when continued with ` -e $pkg`, so it might be removed.

lib/Rex/Pkg/FreeBSD.pm
```diff
@@ -69,9 +75,21 @@ sub is_installed {
...
+  }
+
   # pkg info -e allow get quick answer about is pkg installed or not.
   my $command =
-    $self->{commands}->{query} . " -e $pkg" . ( $version ? "-$version" : "" );
+      $self->{commands}->{query}
+    . $glob_arg
```
ferki, line 90: For the sake of consistency, what do you think about using a similar logic for `$glob_arg` as for `$version` below? That is, only adding it to the command if needed/defined.

That might simplify the code, and we could even rename it to a more generic `$extra_args` if needed.

VVelox: Done!

dist.ini
```diff
@@ -50,6 +50,7 @@ contributor = Ilya Evseev <ilya.evseev@gmail.com>
 contributor = Niklas Larsson <pnlarsson@gmail.com>
 contributor = Qiao Liu <qiao.liu@yottaa.com>
 contributor = Renato CRON <rentocron@cpan.org>
+contributor = Zane C. Bowers-Hadley <vvelox@vvelox.net>
```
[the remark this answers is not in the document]
VVelox: Done!

lib/Rex/Pkg/Base.pm
```diff
@@ -66,11 +66,22 @@ sub update {
   my $env     = $option->{'env'}     || ();
 
   Rex::Logger::debug( "Installing $pkg" . ( $version ? "-$version" : "" ) );
-  my $cmd = sprintf $self->{commands}->{install}, $pkg;
+  my $cmd;
+  if ( ( $pkg =~ /\*/ )
+    && defined( $self->{commands}->{install_glob} ) )
+  {
+    # back quote asterisks so the command can actually be used
+    $pkg = ~s/\*/\\\*/g;
```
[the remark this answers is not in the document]
VVelox: Done!

lib/Rex/Pkg/Base.pm
```diff
@@ -135,7 +146,17 @@ sub remove {
   my ( $self, $pkg ) = @_;
 
   Rex::Logger::debug("Removing $pkg");
-  my $cmd = sprintf $self->{commands}->{remove}, $pkg;
+  my $cmd;
+  if ( ( $pkg =~ /\*/ )
+    && defined( $self->{commands}->{remove_glob} ) )
+  {
+    # back quote asterisks so the command can actually be used
+    $pkg =~ s/\*/\\\*/g;
```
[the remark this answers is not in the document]
VVelox: Done!

lib/Rex/Pkg/FreeBSD.pm
VVelox: This in testing resulted in trying to remove 'foo*-'.  The - at the end would mean it will never match anything.

lib/Rex/Pkg/FreeBSD.pm
VVelox: Done!

lib/Rex/Pkg/FreeBSD.pm
VVelox: Not really, but out of habit I always do this when putting together args stuff like to be overly careful.

lib/Rex/Pkg/Base.pm
```diff
@@ -66,11 +66,22 @@ sub update {
   my $env     = $option->{'env'}     || ();
 
   Rex::Logger::debug( "Installing $pkg" . ( $version ? "-$version" : "" ) );
-  my $cmd = sprintf $self->{commands}->{install}, $pkg;
+  my $cmd;
+  if ( ( $pkg =~ /\*/ )
+    && defined( $self->{commands}->{install_glob} ) )
```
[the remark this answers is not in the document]
VVelox: Yup!

I am wondering if we want to check for * by default and then processing packages. For example apt does not mind globs, but * also needs escaped for that.

ferki:
@VVelox: as discussed on IRC, I've clean up the git history here and pushed the result back to your branch. Now all your changes are in two commits:
  - the first one adds the bits needed for initial package glob support to the base class
  - the second one adds the bits specific to FreeBSD package globs

I'll wait for test results here, and also reply on any remaining open items.

lib/Rex/Pkg/Base.pm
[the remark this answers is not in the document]
ferki: If I understand you correctly, yes, it might be possible and useful to always escape the contents of `$pkg`, and not just in the cases when we think it would need escaping.

If you would like to work on that on this PR, please mark it as draft, hack on it, and mark it "ready for review" again when you are done, and tests pass.

Alternatively, it might be just as good to do that on a separate follow-up PR (or to open an issue about it as a reminder).

PR #1304, opened 2020-04-22
Add tools in contrib/ to strip dist.ini in-place for Gentoo

kentfredric:
The implementation of this is slightly more complicated than strictly
necessary, but it adds a lot of value in output transparency and the
ability to error when things that were expected to be removed, failed
to remove, for whatever reason.

The idea is that in gentoo's live ebuilds, instead of this logic
being embedded in the ebuild, we can invoke this during `src_prepare`,
and subsequently, give the git repository far more control of keeping
this problem under wraps, as when plugins are added that pose a
problem, they can be removed again by this script in order to get the
gentoo live ebuilds working again for your users, without needing to
push more changes through gentoo.

I've tried to make the various extraction tools as vendor agnostic
as possible, and that way, if need be, other vendors can have their own
script in this dir which nukes different things.

Also, these contributed files are currently configured in `[GatherDir]`
such that they are trimmed from the published release, as they really
don't have much value there, because as it is, many projects will *not*
work when running `dzil build` on the published sources, as vast
quantities of assumptions it makes about "things not being changed yet"
fail to be true.

### DIFF
--- a/contrib/vendor/gentoo.sh
+++ b/contrib/vendor/gentoo.sh
@@ -1,0 +1,43 @@
+# This script patches the tree, and dist.ini to remove plugins
+# and soforth that are either useless, or harmful, when building
+# rex from git sources for Gentoo
+#
+# Many of these are useful for authorship to ensure the code meets
+# various standards, but this is not useful at a point of consumption
+# as defects in things like documentation or spelling, aren't things
+# that should be considered "blockers" for an end user, and can
+# substantially inflate the dependency tree without adding any user
+# value
+
+# usage:
+#   cd ${git_root}
+#   bash contrib/vendor/gentoo.sh
+
+info() {
+  printf "contrib/vendor/gentoo.sh: %s\n" "$@" >&1
+}
+err() {
+  printf "contrib/vendor/gentoo.sh: ERROR: %s\n" "$@" >&1
+  exit 1
+}
+
+blacklist=(
+  PodSyntaxTests
+  Test::MinimumVersion
+  Test::Perl::Critic
+  Test::Kwalitee
+  Test::CPAN::Changes
+)
+for section in "${blacklist[@]}"; do
+  perl contrib/vendor/remove-section.pl "${section}" dist.ini ||
+    err "Can't remove [$section]"
+done
+
+devdep_blacklist=(
+  Test::Kwalitee
+  Test::PerlTidy
+  Test::Pod
+)
+
+perl contrib/vendor/remove-dev-requires.pl "${devdep_blacklist[@]}" dist.ini ||
+  err "Can't strip DevelopRequires prereqs"
--- a/contrib/vendor/remove-dev-requires.pl
+++ b/contrib/vendor/remove-dev-requires.pl
@@ -1,0 +1,119 @@
+#!perl
+use strict;
+use warnings;
+
+# This script is to remove various entries from dist.ini
+# for gentoo purposes.
+#
+# Primarily, it removes entries from the DevelopRequires Prereqs
+# section, and if only blacklisted entries are found, removes the whole section
+#
+# usage: perl contrib/vendor/gentoo-dev-requires.pl Test::Kwalitee Test::PerlTidy dist.ini
+
+use File::Spec::Functions qw( splitpath catpath );
+
+exit usage() if grep /\A-(h|H|-help|\?)/, @ARGV;
+
+my $section = '[Prereqs / DevelopRequires]';
+my $file =
+  ( @ARGV > 0 ? pop @ARGV : die usage_err("Missing argument for 'FILE'") );
+my $out = do {
+  my (@path) = splitpath($file);
+  my (@out)  = @path;
+  $out[-1] = "._tmp_" . $path[-1] . "." . $$;
+  catpath(@out);
+};
+my (@blacklisted) =
+  ( @ARGV > 0 ? @ARGV : die usage_err("Missing argument for 'DEP'") );
+my (%removed);
+
+open my $src,  '<', $file or die "Can't open $file for read, $!";
+open my $dest, '>', $out  or die "Can't open $out for write, $!";
+
+my $seen = 0;
+
+main_loop: while ( my $line = <$src> ) {
+  if ( $line !~ /^\Q$section\E/ ) {
+    $dest->print($line);
+    next main_loop;
+  }
+  $seen++;
+
+  # Stash plugin heading for later
+  my $heading       = $line;
+  my $needs_section = 0;
+plugin_loop: while ( my $section_line = <$src> ) {
+    for my $blacklisted (@blacklisted) {
+      $removed{$blacklisted} = 0 unless exists $removed{$blacklisted};
+      if ( $section_line =~ /^\Q$blacklisted\E\s*=/ ) {
+        STDERR->print("Removing $section -> $section_line");
+        $removed{$blacklisted}++;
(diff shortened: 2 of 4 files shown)


ferki:
I had an uncomfortable feeling about merging this into the Rex main repo, but I couldn't put a finger on it for a good while.

If I understand/remember all the bits correctly, this is a way for package vendors to strip out parts of dist.ini when they are trying to build Rex from source with their package manager, for example the live ebuild in Gentoo's portage tree.

While related to Rex, this topic feels more like a generic problem of "how to build Perl packages using Dist::Zilla from source with a different dist.ini than what upstream considers necessary". I agree that there are cases when certain parts can be thrown out from the end-user's perspective, and that's probably a good enough simplification of the build process for the given purpose.

This sounds like something that is quite opinionated, and probably differs from vendor to vendor (or perhaps even from maintainer to maintainer), and I don't think I want to push Rex to be the source which dictates and/or maintains these downstream decisions. That's part of the reasons why we decided to stop packaging Rex on our side, and only release it to CPAN.

All in all, I feel like this is a highly interesting idea that would be better off having its own repo/project.

@kentfredric: if you would prefer keeping "close to Rex" for now, I can offer another repo under the `RexOps` namespace. Do you have a good name for this tool in mind?

kentfredric:
A lot of this is very specific to the "build sources from live git" work flow, which I should point out, is generally not done with Perl, nor encouraged in Perl.

Its not so much oriented at "package vendors" so much, but to provide you with an "out", to work around various problems that may crop up between releases while the code is still in -9999 versions.

Vendors don't need any help doing this, but doing it this way is intended to give *you* more control over the deployment process when people build from git sources, without necessitating you send PR's to fix the ebuild (and this avoids the proliferation of git-specific patch logic leaking into the non-git ebuilds)

So it sits *right* up against the vendor/upstream split, just sitting slightly on your side of the split, instead of the vendor side of the split.

Putting this code anywhere else would be a net negative, because its coupled specifically to the Rex build process, and it would also be a premature generalization, ( as nothing else needs this at present ).

And it is by very nature "hackish", in that, the approaches used are fine because we know exactly what code its working with, where a generalized approach would need to accommodate a broader rage of variability, and thus, increase its complexity.

ferki:
> A lot of this is very specific to the "build sources from live git" work flow, which I should point out, is generally not done with Perl, nor encouraged in Perl.

Then why are we putting so much effort into doing it? :) Sadly, there seems to be no existing projects/modules that would provide functionality for dynamic or conditional configuration for Dist::Zilla.
 
> Its not so much oriented at "package vendors" so much, but to provide you with an "out", to work around various problems that may crop up between releases while the code is still in -9999 versions.
> 
> Vendors don't need any help doing this, but doing it this way is intended to give _you_ more control over the deployment process when people build from git sources, without necessitating you send PR's to fix the ebuild (and this avoids the proliferation of git-specific patch logic leaking into the non-git ebuilds)

If a vendor (e.g. Gentoo), would like to build Rex differently than upstream, that is great, but I think it still primarily is a decision and concern of the vendor. Why take up the burden of that downstream decision on upstream side? What we patch here doesn't look like a bug or mistake, but more like an opinion.

> So it sits _right_ up against the vendor/upstream split, just sitting slightly on your side of the split, instead of the vendor side of the split.

Yes, I think part of the issue is that I'm not comfortable bringing back vendor-specific packaging details into the project scope, but let's suppose we decide to do it at a later point in time.

Wouldn't it be more straightforward then to directly provide a Gentoo-specific dist.ini as `contrib/vendor/gentoo/dist.ini`, instead of ~250 new lines of a custom solution which patches out 11 non-empty lines from the vanilla dist.ini? Or perhaps even just the patch that the Gentoo ebuild may apply?

If we start to have alternative dist.ini files (or generator scripts for them), we'd also need to find a way to build and test the changes to these dist.ini files (and/or to the generator scripts) before merging these changes to the master branch. I guess that would require having a Gentoo build machine available, and also having some kind of hint for the ebuild to use some other source than the master branch under some circumstances. That feels complex, but maybe there's a simple way to go about this aspect?

kentfredric:
> Then why are we putting so much effort into doing it? :) Sadly, there seems to be no existing projects/modules that would provide functionality for dynamic or conditional configuration for Dist::Zilla.

While saying that its not *generally* done, Gentoo is special here. What's *also* not *generally* done by vendors is to even provide a way to deploy entirely unchecked code directly from git as part of your build chain ;)

Most vendors who would even anticipate such a thing only have the *option* of doing it as a snapshot, where they'd run the deploy code against a specific commit, and then ship a hard built copy of that commit with a patch-revision to their users.

> Wouldn't it be more straightforward then to directly provide a Gentoo-specific dist.ini as contrib/vendor/gentoo/dist.ini, instead of ~250 new lines of a custom solution which patches out 11 non-empty lines from the vanilla dist.ini? Or perhaps even just the patch that the Gentoo ebuild may apply?

Mostly the problems there are:
- The ini file gets out-of-sync, and so you double effort to keep gentoo simply working when you add new changes that are otherwise unproblematic
- The general nature of patches aren't well suited to the sorts of mangling this code does, and the patches have a high risk of simply failing to apply ( and you'll get bitten perhaps by the specific mechanism gentoo uses to apply patches if you don't format the patch differently when you update it )

That's all this code really papers over, it implements "patch" in a somewhat-syntax-aware and platform-agnostic way, so the patcher only has to indicate the blocks of syntax they need to get rid of.

And that way, if gentoo needs different mechanisms than is already in git, its easier to patch the dist.ini patching code, than to ship a patch for dist.ini itself. ( Patches being "add/remove X line" are more resilient than patches that "augment X line" )

I honestly don't expect to see other vendors using any of this, but it should make their lives easier if they do. (eg: they just add their own patch script that utilizes the existing provided tool libraries, instead of reworking dist.ini in a more fragile way)

ferki:
What I'm considering as an alternative approach is to let Travis CI publish the build results of each successfully tested commits of the master branch to somewhere, e.g. in a `build` branch or in another `rex-builds` repository.

Tagged releases may be able to use GitHub Releases as well, and each successful commit would have its associated build results saved in a location that may be used by vendors who want to supply their users with bleeding edge versions.

This would mean there would be official git builds from the Rex project for each "green" commits which would be built in the same way as intended by upstream (so it's easier to commit to maintaining these results), while vendors who want to build Rex in a non-official way are still free to do so on their end if they insist.

@kentfredric: what do you think? Would it be an acceptable middle ground between all the various factors involved?

kentfredric:
> What I'm considering as an alternative approach is to let Travis CI publish the build results of each successfully tested commits of the master branch to somewhere, e.g. in a `build` branch or in another `rex-builds` repository.

If you go that way, I'd encourage you to find a way of doing that outside of dzil (ie: some bash glue that calls `dzil build` and stashes the result itself somewhere), otherwise you'll grossly complicate the workflow for general contribution/development.

( I say this having used such a thing myself, and grown unfond of how it does this )

ferki:
> If you go that way, I'd encourage you to find a way of doing that outside of dzil (ie: some bash glue that calls `dzil build` and stashes the result itself somewhere), otherwise you'll grossly complicate the workflow for general contribution/development.

Exactly. Current Travis CI setup already calls `dzil build`, we just don't have a deploy step configured upon success.

ferki:
Based on the discussion here, I'm closing this one in favor of #1390 (which would probably also be a better place to discuss ideas and implementation details).

Thanks for the food for thought, @kentfredric!

Issue #1310, opened 2020-04-30
Rex chokes when "external_master" openssh option is set


sdondley:
See https://github.com/RexOps/Rex/issues/1309 for reference. external_master option is explained here: https://metacpan.org/pod/Net::OpenSSH#external_master

Fix needed at `Rex/Interface/Connection/OpenSSH.pm` at line 97 to:

```
  push @connection_props, master_opts      => \@ssh_opts_line if @ssh_opts_line
    && !($net_openssh_constructor_options{external_master});
```

Proper test and patch forthcoming.

ferki:
Good catch, and without looking at the code context, the proposed patch seems about right :+1:

PR #1312, opened 2020-05-01
Allow exernal_master OpenSSH initializer option to be set

sdondley:
For #1310 

The test can probably be improved. It just tests the proper error message thrown because no real connection is set.

### DIFF
--- a/t/issue/1310.t
+++ b/t/issue/1310.t
@@ -1,0 +1,36 @@
+use strict;
+use warnings;
+use Rex -base;
+use Rex::Task;
+use Rex::Config;
+use Test::More;
+use Test::Warn;
+use Test::Exception;
+
+my %have_mods = (
+  'Net::OpenSSH' => 1,
+);
+
+for my $m ( keys %have_mods ) {
+  my $have_mod = 1;
+  eval "use $m;";
+  if ($@) {
+    $have_mods{$m} = 0;
+  }
+}
+
+unless ( $have_mods{'Net::OpenSSH'} ) {
+  plan skip_all =>
+    'Net::OpenSSH not installed.';
+}
+else {
+  plan tests => 1;
+}
+
+my %hash = (initialize_options => {'external_master' => 1, 'ctl_path' => '/home/me/.ssh/socket_name'});
+Rex::Config->set_openssh_opt(%hash);
+#ok (Rex::Config->set_openssh_opt(%hash), 'can set external_master');
+my $t1 = Rex::Task->new( name => "foo" );
+$t1->set_server("192.168.1.1");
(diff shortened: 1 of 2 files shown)


t/issue/1310.t
```diff
@@ -0,0 +1,36 @@
...
+  plan skip_all =>
+    'Net::OpenSSH not installed.';
+}
+else {
+  plan tests => 1;
+}
+
+my %hash = (initialize_options => {'external_master' => 1, 'ctl_path' => '/home/me/.ssh/socket_name'});
+Rex::Config->set_openssh_opt(%hash);
+#ok (Rex::Config->set_openssh_opt(%hash), 'can set external_master');
```
ferki, line 32: Are there any reason to include commented out code in this test?

t/issue/1310.t
```diff
@@ -0,0 +1,36 @@
...
+my $t1 = Rex::Task->new( name => "foo" );
+$t1->set_server("192.168.1.1");
+throws_ok ( sub { $t1->run('192.168.1.1') }, qr/authenticate against/, 'authenticate against');
+#warnings_like { $t1->run('192.168.1.1') } qr/WARN.*does not point to a socket/, 'throws socket warning';
```
ferki, line 36: Same question as the above once about including commented out code.

t/issue/1310.t
```diff
@@ -0,0 +1,36 @@
...
+else {
+  plan tests => 1;
+}
+
+my %hash = (initialize_options => {'external_master' => 1, 'ctl_path' => '/home/me/.ssh/socket_name'});
+Rex::Config->set_openssh_opt(%hash);
+#ok (Rex::Config->set_openssh_opt(%hash), 'can set external_master');
+my $t1 = Rex::Task->new( name => "foo" );
+$t1->set_server("192.168.1.1");
+throws_ok ( sub { $t1->run('192.168.1.1') }, qr/authenticate against/, 'authenticate against');
```
ferki, line 35: If I read this correctly, this would attempt connecting to `192.168.1.1` during the test. Please make sure no network connections are attempted during test time.

ferki:
Alright, I'm trying to cherry-pick the actual bugfix commit from this PR, and then clean that up (commit message has typo, code is untidy, probably can be refactored for increased readability, and at that point a rebase on top of current master won't hurt either). 

The test is not going to make it this time (doesn't pass, not clean, and attempts network connection). At this point I'm not exactly sure what would be the best testing strategy, though.

Manual testing indicates it works as expected for the main use case (setting `external_master` to a true value, and passing a `ctl_path` parameter with it), and it feels like an acceptable risk to take (which sounds like famous last words...worst case we'll revert the change).

Issue #1311, opened 2020-04-30
Documentation missing for setting init options for openssh object construction


sdondley:
The `Rex::Config->set_openssh_opt(%hash)` option lacks documentation for setting init options for an ssh object.

See https://github.com/RexOps/Rex/issues/1309 for reference.

ferki:
Right, this is exactly what I just [commented](https://github.com/RexOps/Rex/issues/1309#issuecomment-626679960) over on #1309.

I hate to duplicate the same thing, but I think the best option is to copy-paste relevant parts of my comment here, and close the other issue as duplicate in favor of this one.

>The related user story format would probably look something like:
>
>**As:** a rex user,
**I would like to:** know how to set constructor options for the underlying Net::OpenSSH object,
**so:** I could fine-tune connection details according to my use case.
>
>This topic possibly could go directly into `Rex::Interface::Connection::OpenSSH`[...]
>
>For the list of available options, probably the best we could do is to point readers to the corresponding external documentation: Net::OpenSSH->new().

PR #1313, opened 2020-05-01
document how to set OpenSSH initializer options

sdondley:
For #1311

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -11,6 +11,7 @@ Revision history for Rex
  - Fix task hooks specified in Rexfile for tasks defined in modules
 
  [DOCUMENTATION]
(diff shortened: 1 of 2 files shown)


ferki:
I've spent quite some time writing a full review for this one, but then I realized it would be much faster to just push follow-up changes directly on the PR.

I've cleaned up an off-topic change (nice catch about the extra trailing space, though!), restored formatting consistency and reworded the changes for additional clarity.

This PR still needs to be rebased on top of current master (to fix the ChangeLog entry), and many commits can be simply squashed away. The commits are kept separate for easier review.

I'll do the remaining steps later, just wanted to push the modifications ASAP.

PR #1319, opened 2020-05-12
Clarify auth docs

### LINKED ISSUE #1314: improve documentation regarding "auth for" operation wrt groups
Documentation should make it explicit that "auth for" command works only for named groups and not for individuals hosts in the group. In other words, if a task is issued to a host from the command line with the -H option, the credentials that apply for the group will not be applied to the individual server even if it is part of a group with an "auth for" command applied to it.

ferki:
This PR fixes #1314 by clarifying the docs of the `auth` and `group` commands.

@sdondley: what do you think, does it cover your original concerns?

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -7,6 +7,7 @@ Revision history for Rex
 
  [DOCUMENTATION]
  - Clarify auth documentation
+ - Clarify setting per-host authentication details within a group
 
  [ENHANCEMENTS]
 
--- a/ChangeLog
+++ b/ChangeLog
@@ -6,6 +6,7 @@ Revision history for Rex
  [BUG FIXES]
 
  [DOCUMENTATION]
+ - Clarify auth documentation
 
  [ENHANCEMENTS]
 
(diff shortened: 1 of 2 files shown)


sdondley:
That's better. I would still prefer to see something like this added in there:

```
C<auth for> settings for groups are applied only when tasks are run in
a group context. In other words, a server doesn't inherit the group's authorization
parameters when a task is run on an individual server.
```

And then you could have your line: `If you are looking for setting them per-host within a group, please take a look at the L<group|https://metacpan.org/pod/Rex::Commands#group> command.`

It's not that big of a deal, though.

ferki:
@sdondley: I added a note about "setting auth info for a host group" is currently considered by Rex as a different operation than "setting auth info for the members of the host group".

sdondley:
Yeah, that's good. I think it's enough to even stop a bonehead like me from reading things into the documentation that aren't there.

ferki:
Thanks for the feedback, I'm glad it covers your concerns! I'm squashing the work commits here, then I'll merge this PR.

PR #1348, opened 2020-06-19
Discontinue support for some EOL Windows versions

### LINKED ISSUE #1347: Test suite failures on Windows 7
## Describe the bug

According to results from CPAN Testers (thanks! :pray:), some of the tests started to fail on Windows 7 for some reason. Some failing test reports:

- http://www.cpantesters.org/cpan/report/ec6ede96-7155-1014-8fdb-d49bd618ba9a
- http://www.cpantesters.org/cpan/report/77e16c22-7156-1014-ae68-f898d618ba9a

Interestingly enough, the failure seems to be caused by the system in question claims there's a command available with the litreal name of `I'm pretty sure this command doesn't exist anywhere`.

The [reports analysis for Rex-1.11.0](http://analysis.cpantesters.org/solved?distv=Rex-1.11.0) (even more thanks! :heart: ) also confirms high correlation between test failures and Windows 7 as OS. It also seems to correlate with the CPU model, but both information should be taken into account with a grain of salt, since it seems to be a single system producing all the similar failures.

Unfortunately the reports seems to be generated by "NONAME" so I don't see any way to contact the actual person for further clarification.

Either way, discontinuing Windows 7 support still seems timely, given that "Mainstream support ended on January 13, 2015" and "Extended support ended on January 14, 2020" according to the [Windows 7 page of Wikipedia](https://en.wikipedia.org/wiki/Windows_7).

## How to reproduce it

Lacking a Windows 7 system, I'm not entirely sure, but perhaps run the test suite on Windows 7?

## Expected behavior

Rex should bail out early from the build process on unsupported operating systems.

## Circumstances

 - Rex version:  1.11.0
 - Perl version: various
 - OS running rex: Windows 7
 - OS managed by rex: local

## Debug log

-

ferki:
This PR closes #1347 by disabling support for the following Windows versions, which are all already EOL:

- Windows Vista
- Windows 7
- Windows Server 2008
- Windows Server 2008 R2

Due to the slightly increased complexity of the check, the logic has been moved to a file that is included into `Makefile.PL` by `MakeMaker::Awesome`.

## Checklist

- [ ] tests pass on Travis CI 
- [x] git history is clean    
- [x] git commit messages are [well-written](https://chris.beams.io/posts/git-commit)

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -4,6 +4,7 @@ Revision history for Rex
  [API CHANGES]
 
  [BUG FIXES]
+ - Discontinue support for Windows 7, Windows Server 2018 R2, and older
 
  [DOCUMENTATION]
 
--- a/dist.ini
+++ b/dist.ini
@@ -22,7 +22,7 @@ skip = __Rexfile__
 
 [MakeMaker::Awesome]
 eumm_version = 7.1101
-header = die 'OS unsupported' if ( $^O eq 'MSWin32' && scalar((Win32::GetOSVersion())[1]) < 6 );
+header_file = misc/check_supported_OS.pl
 
 [ManifestSkip]
 
--- a/misc/check_supported_OS.pl
+++ b/misc/check_supported_OS.pl
@@ -1,0 +1,1 @@
+die 'OS unsupported' if ( $^O eq 'MSWin32' && scalar((Win32::GetOSVersion())[1]) < 6 );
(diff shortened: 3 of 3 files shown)


rai-gaurav:
Just an observation- 
Change log on https://metacpan.org/changes/release/FERKI/Rex-1.11.0.1-TRIAL mentions -

1.11.0.1 2020-06-20 Ferenc Erki <erkiferenc@gmail.com>
 [BUG FIXES]
   - Discontinue support for Windows 7, Windows Server 2018 R2, and older

There is typo in change log - 'Windows Server 2018' instead of 'Windows Server 2008' which is  misleading.

ferki:
@rai-gaurav: thanks! I'll fix that for the stable release :+1:

Issue #1289, opened 2020-04-10
Make rex bash completion script official


ferki:
We have the `misc/rex-tab-completion.bash` script with us for a good while, and it has been proven useful and reliable. I'd say it's time to make it official and include it in the next release.

I'd expect things like this to do:
  - move the script into a more standard `share` directory
  - make sure Dist::Zilla/Manifest is configured correctly to include that directory in the releases
  - document usage (perhaps in `lib/Rex.pm`? see also RexOps/rexify-website#146)

mestia:
I have a couple of suggestions regarding the bash tab-completion.

1. currently, bash tab completion doesn't support multiple Rexfiles in different dirs. After completing hosts or groups for one Refxile and navigating to a directory with a different Rexfile, completion will show wrong "old" hosts or groups. This can be fixed by calling `rex -Ty 2>/dev/null` in every completion call.

2. host completion. I'd like to complete not only hosts defined in a Rexfile, but also all hosts available via `_known_hosts_real` bash-completion function, which can extract hosts from my ~/.ssh/config for example. One can chain rex hosts and system hosts this way: `COMPREPLY=( $( compgen -W '${hosts[@]}' -- "$cur" ) ) ||  _known_hosts_real -a "$cur"` - by default complete rex hosts, if a different host is specified, complete it from `_known_hosts_real`

See: https://github.com/mestia/Rex/blob/master/misc/rex-tab-completion.bash
I'll open a PR if the stuff above makes sense.

PR #1352, opened 2020-06-20
Update tab completion for bash

mestia:
This PR fixes the following issues:
 - remove shebang, since completions aren't executables
 - complete available -* options
 - also complete hosts provided by _hosts function, for example from `~/.ssh/config`
 - call `rex -Ty` every time to make completion work between different `Rexfiles`

## Checklist
- [x] git commit messages are [well-written](https://chris.beams.io/posts/git-commit) (hopefully :))

### DIFF
--- a/.mailmap
+++ b/.mailmap
@@ -1,3 +1,4 @@
+Alex Mestiashvili <amestia@rsh2.donotuse.de> <mailatgoogl@gmail.com>
 Alexander Romanenko <alex1line@gmail.com>
 Anders Ossowicki <and@vmn.dk> <aowi@novozymes.com>
 Andrej Zverev <andrej.zverev@gmail.com> <az@sunner.semmy.ru>
--- a/ChangeLog
+++ b/ChangeLog
@@ -5,16 +5,19 @@ Revision history for Rex
 
  [BUG FIXES]
  - Discontinue support for Windows 7, Windows Server 2018 R2, and older
+ - Avoid caching of Bash completion options to support multiple Rexfiles
 
  [DOCUMENTATION]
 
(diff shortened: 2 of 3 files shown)


ferki:
Oh, I've left out some housekeeping: ref #1289 ([comment](https://github.com/RexOps/Rex/issues/1289#issuecomment-643139128))

mestia:
> It works great based on my local tests, thanks! +1

Great!

> 
> I could see a few nitpicks about whitespace and commit organization, but nothing critical, so it's good to merge.

I am always open for critic, so feel free to tell me what could be improved.

ferki:
> I am always open for critic, so feel free to tell me what could be improved.

Well, if you insist :) Dumping my thoughts:

Re: whitespace: there's an unnecessary extra empty line, and a double-space - I'll add a follow-up commit to this PR before merge, no need to change anything.

Re: commit organization: the commit contains 4 distinct changes, which makes it harder for the reviewer to tell which part of the diff is related to which change (especially when more than one change affects the same line), and might be harder for the maintainer to see where a specific change was introduced (or roll back only one of the changes). The commit message and the PR description lists those, which is good, but it also could be 4 commits (one commit for each change) next time.

Again, these are more like nitpicks, and all-in-all the PR is good enough to accept :+1:

mestia:
Ok, in general I agree and also like to commit often, but since in this case the changes are kind of trivial, I though it would be easier to have it as one change. Thanks for accepting the PR!

ferki:
Alright, I also added changelog and mailmap entries as a follow-up, so now all important elements should be self-contained in the PR ^^

Issue #1106, opened 2016-10-19
problem with rsync command and wildcards


krimdomu:
There is a problem reported via IRC that wildcards and the rsync command doesn't work.

I think this is a problem with the function that try to calculate the path.

Example: https://people.debian.org/~christoph/POC.txz

ferki:
FTR: #1075 seems at least related, if not duplicate.

mestia:
See the same problem with rex 1.6.
It also complains about 

> Redundant argument in sprintf at /usr/share/perl5/Rex/Commands/Rsync.pm line 238

ferki:
> See the same problem with rex 1.6.
> It also complains about
> 
> > Redundant argument in sprintf at /usr/share/perl5/Rex/Commands/Rsync.pm line 238

@mestia: seems like that was a different problem, and it was resolved on #1297 (and a similar one on #1237).

ferki:
Closing as duplicate of #1075.

Issue #1315, opened 2020-05-09
cannot sync to server over non-standard ssh port


sdondley:
See this line in the code: https://metacpan.org/release/Rex/source/lib/Rex/Commands/Rsync.pm#L118

That line makes a call to this function: https://metacpan.org/source/FERKI/Rex-1.10.0/lib/Rex/Helper/IP.pm#L25

It looks like the function expects the port to be attached to the server name and separated by colon like `server:22`

Issue #1075, opened 2016-07-20
Possible buggy variable evaluation with rsync


ferki:
I'm just converting a [thread](https://groups.google.com/forum/#!msg/rex-users/E5OBofIqDUk/6sUDEJPZAQAJ) on the mailing list into an issue, where `path_map` is used with `sync`. When the source directory contains a `*`:

``` perl
set "path_map", {
    "PATH/" => [ "filesys/{environment}/{hostname}.{domain}/",
                "filesys/{environment}/",
                "filesys/default/" ]
};
sync 'PATH/dir/*', $conf->{'src_dir'};
```

then it fails with the `PATH/dir/*` part becoming `loader/0x12fa460/PATH/dir/*`.

Without `*` and/or absolute directories it works fine:

```
sync 'PATH/dir/', $conf->{'src_dir'};
```

ferki:
This was duplicated by #1106, attaching POC code here to have the information in one place: [POC.tar.gz](https://github.com/RexOps/Rex/files/4832449/POC.tar.gz)

So far it seems, the main problem is that the source and destination values are having single quotes around them (I presume to handle pathes with space characters in them):

```
rsync -rl --verbose --stats '/home/ferki/projects/rex/t/sync' '/tmp/Nt136isqQl'
```

which is fine in general, but breaks as soon as a wildcard is added:

```
rsync -rl --verbose --stats  '/home/ferki/projects/rex/t/sync/*' '/tmp/Nt136isqQl'
[...]
rsync: link_stat "/home/ferki/projects/rex/t/sync/*" failed: No such file or directory (2)
[...]
```

ferki:
Actually, it might be simpler than that:

```perl
task 'test_get_file_path', sub {
    say Rex::Helper::Path::get_file_path('/tmp');
    say Rex::Helper::Path::get_file_path('/tmp/*');
};
```

It prints:

```
[2020-06-25 19:54:04][<local>]  INFO Running task test_get_file_path on <local>
/tmp
/loader/0x5568477db850//tmp/*
[2020-06-25 19:54:04][<local>]  INFO All tasks successful on all hosts
```

`Rex::Helper::Path::get_file_path` was not really designed to take wildcards into account, so either it needs a fix for that, or another method should be introduced that supports wildcards.

ferki:
So it turns out to be a combination of things:

[...]

PR #1353, opened 2020-06-23
Fix rsync operations for source directories containing a relative pat…

sdondley:
This seems to do the trick for #1106.

I also ran a quick local test with a source destination that did not have a wildcard and it still worked.

### DIFF
--- a/lib/Rex/Commands/Rsync.pm
+++ b/lib/Rex/Commands/Rsync.pm
@@ -122,7 +122,7 @@ sub sync {
   if ( !exists $opt->{download} && $source !~ m/^\// ) {
 
     # relative path, calculate from module root
-    $source = Rex::Helper::Path::get_file_path( $source, caller() );
+    $source = Rex::Helper::Path::get_file_path($source);
   }
 
   Rex::Logger::debug("Syncing $source -> $dest with rsync.");


sdondley:
`caller()` was removed. `get_file_path()` calls `caller()` it if there is no second argument. See https://github.com/RexOps/Rex/blob/master/lib/Rex/Helper/Path.pm#L60

ferki:
> `caller()` was removed. `get_file_path()` calls `caller()` it if there is no second argument. See https://github.com/RexOps/Rex/blob/master/lib/Rex/Helper/Path.pm#L60

Yep, I just updated my comment once I realized I read the diff backwards :) Thanks for the explanation!

sdondley:
Oh, one more thing I noticed. So with this patch, the full path to the dir is now found. In other words, if the source dir is `some_dir/*`, the source dir will become absolute with something like `/home/me/some_dir/*`.

However, if you don't have a wildcard character and just have `some_dir`, the source dir is not fully qualified, even without this patch.

I think, ideally, directories without a wildcard should get fully qualified as well. To achieve, this, you can comment out this line: https://github.com/RexOps/Rex/blob/master/lib/Rex/Helper/Path.pm#L108. With that line commented out, the `get_file_path` will then return a full qualified path to the file (or dir).

But things still seem work without being fully qualified so maybe it's better not to mess with it.

At any rate, it feels like whoever wrote Rsync to use `get_file_path` used it as a quick hack. Someone should probably take a really close look at how Rsync resolves relative paths.

ferki:
I don't think we can fully untangle the various rsync questions without having tests that describe the expected behavior. I started write some tests locally (and so far, I had to add support for local rsync runs to be able to test).

I also closed #1106 as duplicate of #1075. This should take care of cross-linking the related issues/PRs.

> I think, ideally, directories without a wildcard should get fully qualified as well.

That might work, yes.

> To achieve, this, you can comment out this line: https://github.com/RexOps/Rex/blob/master/lib/Rex/Helper/Path.pm#L108. With that line commented out, the `get_file_path` will then return a full qualified path to the file (or dir).

Without reading the code, it feels a bit risky to change the path resolution behavior for all consumers both in the core and in every other existing code outside of it. Perhaps some kind of "please give me the absolute path flag" has less disrupting potential.

> But things still seem to work without the source dir being fully qualified so maybe it's better not to mess with it.

Yes, ideally it would be preferred to address the bug directly, instead of redesigning the overall behavior.

> At any rate, it feels like whoever wrote Rsync to use `get_file_path` used it as a quick hack. Someone should probably take a really close look at how Rsync resolves relative paths.

I'd say the motivation was more like "let the Rsync module use the exact same path resolution logic like any other function" (something like look for candidates in the current dir, the module path, or in any of path_map locations, also look for files with extension matching the environment name (prod, dev, etc.), and so on). In that sense, it would be quite desirable to keep the behavior the same.

sdondley:
Right, it definitely needs tests. There should be one set of tests just to determine if it is generating the desired command. That way, you don't need to have rsync installed to actually run the tests.

To test the rest of the command, you'd have to set up a fake server that sent back different response strings. I don't think that's really necessary for now as that doesn't seem to be broken.

I would like to figure out how to handle running `rsync sudo` per https://github.com/RexOps/Rex/pull/1298. There is also https://github.com/RexOps/Rex/issues/1315

Also, the rsync command automatically inserts the `-d` and `-l` options. I haven't filed a bug report on this but it really probably shouldn't be doing that. Or if it is, that should at least be documented.

Overall, I think the modules needs a little mini overhaul because it has some fairly big problems.

ferki:
> Right, it definitely needs tests. There should be one set of tests just to determine if it is generating the desired command. 
That way, you don't need to have rsync installed to actually run the tests.

Could be done, though so far I'm more interested in testing the behavior ("does it actually sync the files as expected"), instead of "can it generate an expected rsync command". That way, I expect many of the tests would be easier to reuse for `Rex::Commands::Sync` as well.

> 
> To test the rest of the command, you'd have to set up a fake server that sent back different response strings. I don't think that's really necessary for now as that doesn't seem to be broken.

I feel mocking an overkill, while it's definitely possible. Unless there's an existing solution, I'm not sure it's worth to invent and maintain our own rsync mocking server inside rex core.

Currently, I chose to add a `t/sync` dir with a bunch of files and a nested dir under it, and ask the test to sync it to a temp directory, then compare the contents of source and the target.

> 
> I would like to figure out how to handle running `rsync sudo` per #1298. There is also #1315

Yep, they are rsync related, but offtopic here.

> 
> Also, the rsync command automatically inserts the `-d` and `-l` options. I haven't filed a bug report on this but it really probably shouldn't be doing that. Or if it is, that should at least be documented.

Yeah, not sure if bug or feature request, but definitely a different issue :)

> 
> Overall, I think the modules needs a little mini overhaul because it has some fairly big problems.

Rsync is also the only one module that is not running on the managed host. In the long run, it might worth cutting the whole module out into a separate distribution.

With all that, we drifted quite far away of the original topic of this PR though and I'd strongly prefer not mixing too many topics in a single thread.

How about opening an feature request about writing the (r)sync tests first, and only after it's there, _then_ figure out what's broken and how to fix?

ferki:
As [commented] on #1075, the underlying issue is something else, and I don't consider removing `caller()` as the proper fix for those.

Tests are also a different topic.

PR #1371, opened 2020-07-01
- Allow sync() function work for local rsync operations

### LINKED ISSUE #1370: running sync() locally results in errors
## Describe the bug

Doing a local rsync with the sync() command results in errors.

## How to reproduce it

From comm

```
sync('/Users/me/workshop/rsync1', '/Users/me/workshop/rsync2' );
```

1) Create a task called 'sync_task' with sync command that syncs between two local directories.
2) Run `rex sync_task`

Errors:

`[2020-07-01 18:08:07] ERROR - Can't use string ("") as a HASH ref while "strict refs" in use at /Users/stevedondley/git_repos/Rex_fork/Rex/lib/Rex/Commands/Rsync.pm line 166, <> line 5.`

```
use strict;
use warnings;
use Rex -feature => [ qw( 1.4 ) ];
use Rex::Commands::Rsync;
no_ssh;

task test_sync => sub {
  sync('/Users/me/some_dir',
    '/Users/me/some_other_dir', { download => 1, parameters =>
    '--rsync-path="nice -n 19 rsync" --archive' });
};

```

## Expected behavior

No errors.

## Circumstances

 - Rex version: 1.11 
 - Perl version:  5.24
 - OS running rex: Darwin
 - OS managed by rex: Darwin

sdondley:
This PR fixes #1370 by checking to ensure $auth is set throughout the file.

## Checklist

- [x] tests pass on Travis CI 
- [x] git history is clean    

### DIFF
--- a/lib/Rex/Commands/Rsync.pm
+++ b/lib/Rex/Commands/Rsync.pm
@@ -161,7 +161,7 @@ sub sync {
     Rex::Logger::debug("Downloading $source -> $dest");
     push @rsync_cmd, "rsync -rl --verbose --stats $params ";
 
-    $source = $auth->{user} . "\@$servername:$source";
+    $source = $auth->{user} . "\@$servername:$source" if $auth;
 
     if ( !$local_connection ) {
       push @rsync_cmd, "-e '\%s'";
@@ -175,7 +175,7 @@ sub sync {
 
     if ( !$local_connection ) {
       push @rsync_cmd, "-e '\%s'";
-      $dest = $auth->{user} . "\@$servername:$dest";
+      $dest = $auth->{user} . "\@$servername:$dest" if $auth;
     }
   }
 
@@ -192,11 +192,13 @@ sub sync {
   $cmd = join( " ", @rsync_cmd );
 
   if ( !$local_connection ) {
-    my $pass           = $auth->{password};
+    my $pass;
+    $pass = $auth->{password} if $auth;
     my @expect_options = ();
 
-    my $auth_type = $auth->{auth_type};
-    if ( $auth_type eq "try" ) {
+    my $auth_type;
+    $auth_type = $auth->{auth_type} if $auth;
+    if ( $auth && $auth_type eq "try" ) {
       if ( $server->get_private_key && -f $server->get_private_key ) {
         $auth_type = "key";
       }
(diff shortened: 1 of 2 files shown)


lib/Rex/Commands/Rsync.pm
```diff
@@ -161,7 +161,7 @@ sub sync {
     Rex::Logger::debug("Downloading $source -> $dest");
     push @rsync_cmd, "rsync -rl --verbose --stats $params ";
 
-    $source = $auth->{user} . "\@$servername:$source";
+    $source = $auth->{user} . "\@$servername:$source" if $auth;
```
ferki, line 164: Regarding this and other checks of `$auth`:

- is it enough to check for truth?
- would it be better to check for "does `$auth` look like a HASH ref"?
- can the check be made earlier in the code once, instead of checking it in many places?

lib/Rex/Commands/Rsync.pm
```diff
@@ -192,11 +192,13 @@ sub sync {
...
-    my $pass           = $auth->{password};
+    my $pass;
+    $pass = $auth->{password} if $auth;
     my @expect_options = ();
 
-    my $auth_type = $auth->{auth_type};
-    if ( $auth_type eq "try" ) {
+    my $auth_type;
+    $auth_type = $auth->{auth_type} if $auth;
+    if ( $auth && $auth_type eq "try" ) {
```
ferki, line 201: Since we are touching the code anyway, and the logic is modified a bit, would it be a good time to make it simpler and/or more readable by grouping the similar statements together?

E.g.:

```
my ( $pass, $auth_type );
my @expect_options = ();

if ($auth) {
  $pass      = $auth->{password};
  $auth_type = $auth->{auth_type};
};
```

which might even opens up the possibility to use a single hash slice instead:

```
my ( $pass, $auth_type ) = @auth{qw(password auth_type)} if $auth;
my @expect_options       = ();
```

lib/Rex/Commands/Rsync.pm
ferki, line 164: I wonder if the issue could be solved by moving this line into the `if (!local_connection)` branch, instead of adding the `$auth` checks.

Authentication only seems to be important for non-local conections, and we also only need to prefix `$source` with the `user@server:` string for non-local connections.

ferki:
Closing this in favor of #1372.

PR #1375, opened 2020-07-03
Fix broken server argument for CMDB db calls, #1374

### LINKED ISSUE #1374: cmdb documenation does not match behavior ($server argument is ignored)
## Describe the bug

See the documentation for retrieving a value from the cmdb database: https://metacpan.org/pod/Rex::CMDB#cmdb

With cmdb defaults and according to the documentation, I expect the following code to return the value associated with 'some_key' located in cmdb/some_server.yml (or cmdb/environment/some_server.yml or cmdb/my_os/some_server.yml):

`my $value = get cmdb('some_key', 'some_server');`

This is not the case, however. When the above code is execute in a task, the server argument is ignored and Rex tries to pull the key from the yml file for the server the task is being run on. When the above code is executed outside of task, the value is pulled from the local host's cmdb file

## How to reproduce it

1. create a cmdb/some_server.yml and add key/value pairs to it
2. attempt to retrieve values from yml file with `get cmdb` when not running a task on `some_server`

In file cmdb/some_server.yml, add this key/value pair:

`some_key: some_value`

```perl
use strict;
use warnings;
use Rex -feature => [ qw( 1.4 ) ];

task test, 'some_other_server', sub {
  my $value = get cmdb('some_key', 'some_server'));
  print $value;
};

```

## Expected behavior

`$value` should be value found in `some_server.yml` file. Instead, value from `some_other_server.yml` (if it exists) is returned.

## Circumstances

 - Rex version: 1.11
 - Perl version: 5.24.4
 - OS running rex: Darwin
 - OS managed by rex: Darwin

*Additional notes:*
The bug seems to be introduced at https://github.com/RexOps/Rex/blob/master/lib/Rex/CMDB/YAML.pm#L96

This line eventually calls the `parse_path` function: https://github.com/RexOps/Rex/blob/master/lib/Rex/Helper/Path.pm#L216

This function only uses the current connection to determine the server. So any $server argument passed to `get cmdb()` just gets ignored.

sdondley:
This PR fixes #1374 by allowing `$server` argument to work.

## Checklist

- [x] tests pass on Travis CI 
- [x] git history is clean    

### DIFF
--- a/lib/Rex/CMDB/Base.pm
+++ b/lib/Rex/CMDB/Base.pm
@@ -24,9 +24,9 @@ sub new {
 }
 
 sub _parse_path {
-  my ( $self, $path ) = @_;
+  my ( $self, $path, $server ) = @_;
 
-  return parse_path($path);
+  return parse_path( $path, $server );
 }
 
 1;
--- a/lib/Rex/CMDB/YAML.pm
+++ b/lib/Rex/CMDB/YAML.pm
@@ -93,7 +93,7 @@ sub get {
     @files = @{ $self->{path} };
   }
 
-  @files = map { $self->_parse_path($_) } @files;
+  @files = map { $self->_parse_path( $_, $server ) } @files;
 
   my $all = {};
   Rex::Logger::debug( Dumper( \@files ) );
--- a/lib/Rex/Helper/Path.pm
+++ b/lib/Rex/Helper/Path.pm
@@ -214,12 +214,12 @@ sub resolv_path {
 }
 
 sub parse_path {
-  my ($path) = @_;
+  my ( $path, $server ) = @_;
   my %hw;
 
   require Rex::Commands::Gather;
 
-  $hw{server}      = Rex::Commands::connection()->server;
+  $hw{server}      = $server || Rex::Commands::connection()->server;
   $hw{environment} = Rex::Commands::environment();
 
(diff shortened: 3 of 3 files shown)


ferki:
I can review/test this more as soon as I'm back online again properly in a week or so.

Could you add a test case to this PR please to demonstrate the failure, and then that this proposed change fixes that case? It seems it's keeping the current tests happy too already :+1:

sdondley:
After looking at this issue a little more closely, I think bigger problem is that the current tests at `t/cmdb.t` does not run the tests under a typical default condition. The typical default condition is that cmdb path is set to:

```
[
  'cmdb/{operatingsystem}/{hostname}.yml',
  'cmdb/{operatingsystem}/default.yml',
  'cmdb/{environment}/{hostname}.yml',
  'cmdb/{environment}/default.yml',
  'cmdb/{hostname}.yml',
  'cmdb/default.yml'
]
```

And all the tests fail when the path is set to:

```
[
  't/cmdb/{operatingsystem}/{hostname}.yml',
  't/cmdb/{operatingsystem}/default.yml',
  't/cmdb/{environment}/{hostname}.yml',
  't/cmdb/{environment}/default.yml',
  't/cmdb/{hostname}.yml',
  't/cmdb/default.yml'
]
```

The above setting would more properly mimic typical usage because these are the defaults.

So it looks to that there is a major bug in the algorithm for finding the proper yml file when there are variables in the path.

PR #1383, opened 2020-07-22
Misc improvements

tomhukins:
This PR fixes # by .

## Checklist

- [ ] tests pass on Travis CI 
- [ ] git history is clean    


ferki:
Thanks @tomhukins for the PR!

I would prefer to avoid mixing different issues in the same context, and to discuss these changes first in their own issues (or to have a bit more context and explanation here instead of just keeping the PR template blank).

I'm fairly sure I won't be able to review and merge it as it is, but I might be able to cherry-pick some of the ideas earlier than the rest. There seem to be interesting points though, so I'll keep it open for a bit more until I have time to figure them out on my own.

On one hand, I'll work on making the above more obvious in the contributing guide, and on the other, feel free to initiate discussion for your topics in issues first (or even in a chat on IRC) if you'd like to catalyze things early.

ferki (changes requested):
Alright, so one changeset from here has already been adopted, but unfortunately I don't think I can efficiently work with the rest. Overall, I'd like to kindly point to our [Contributing guide](https://github.com/RexOps/Rex/blob/master/CONTRIBUTING.md). More specifically, I'd prefer discussing a change first in an issue, filling out the provided templates, and avoid mixing topics on the same PR.

The second changeset proposes to replace `List::MoreUtils` with `List::Util`. I find the idea of dependency reduction quite nice, however there's no reasoning provided why one should be preferred over the other, and not the other way around. Also, `List::MoreUtils::uniq` and `List::Util::uniqstr` does not seem to be doing the exact same thing (but probably acceptable though).

Perhaps it's about "PP vs XS" and "core vs non-core" ? Are there any other factors? What's the actual problem we aim to solve, or the actual goal to reach? It feels like it would be nice to open a bug report issue where the details and the desired solution could be discussed first (or at least all the context and reasoning could be provided).

The final changeset proposes dropping `Rex::Helper::Array`. It is a good idea, and the best diff is a red diff :) However even without known open source use cases, it has the potential to break existing code, so it should be marked as deprecated first (and then dropped in the next major version later). It also looks like a good candidate to propose this step in a bug report.

All in all, I don't seem to be able to move it forward easily in this form. I'm leaving it open for now as "changes requested", at least until issues are opened and/or further changes are cherry-picked.

tomhukins:
> Alright, so one changeset from here has already been adopted

Thanks for that.  I find the documentation easier to read now.

> The second changeset proposes to replace `List::MoreUtils` with `List::Util`. I find the idea of dependency reduction quite nice, however there's no reasoning provided why one should be preferred over the other, and not the other way around. Also, `List::MoreUtils::uniq` and `List::Util::uniqstr` does not seem to be doing the exact same thing (but probably acceptable though).
> 
> Perhaps it's about "PP vs XS" and "core vs non-core" ? Are there any other factors? What's the actual problem we aim to solve, or the actual goal to reach?

These are good questions and I appreciate I didn't mention my reasoning.  List::Util has shipped as part of core for many years, so using it instead of List::MoreUtils means one less dependency that users need to install.  As I mention in my commit message, Rex already uses `first` and `max` from List::Util.

> It feels like it would be nice to open a bug report issue where the details and the desired solution could be discussed first (or at least all the context and reasoning could be provided).

That's a fair point, but I'm unsure whether I'll get round to it soon.

> The final changeset proposes dropping `Rex::Helper::Array`. It is a good idea, and the best diff is a red diff :) However even without known open source use cases, it has the potential to break existing code, so it should be marked as deprecated first (and then dropped in the next major version later). It also looks like a good candidate to propose this step in a bug report.

That sounds reasonable to me.

> All in all, I don't seem to be able to move it forward easily in this form. I'm leaving it open for now as "changes requested", at least until issues are opened and/or further changes are cherry-picked.

Thank you for your helpful replies and all your work on Rex.

ferki:
Thanks for coming back to this, and making my part easier by providing more context! :heart: 

> Thanks for that. I find the documentation easier to read now.

:tada: :+1: 

> These are good questions and I appreciate I didn't mention my reasoning. List::Util has shipped as part of core for many years, so using it instead of List::MoreUtils means one less dependency that users need to install. As I mention in my commit message, Rex already uses `first` and `max` from List::Util.

Having slept on it once more, I don't see huge risks about preferring List::Util, while it means one less non-core dependency, and faster execution due to XS. Since we specify a recent version of it, I believe it won't break on older Perls even. Did some software archeology, and it seems List::Util was brought in when it didn't support `uniq` yet. And since `List::MoreUtils` was already in use back then, this duality might be a result of an oversight during the original review process.

To be on the safe side, let's make sure the related code paths are properly covered by the test suite, then I think we can cherry-pick that commit too! \o/

> > It feels like it would be nice to open a bug report issue where the details and the desired solution could be discussed first (or at least all the context and reasoning could be provided).
> 
> That's a fair point, but I'm unsure whether I'll get round to it soon.

Alright, let's do it like that next time for a streamlined process, and I might take care of this one later (especially if it also needs tests to cover all replacements).

> > The final changeset proposes dropping `Rex::Helper::Array`. It is a good idea, and the best diff is a red diff :) However even without known open source use cases, it has the potential to break existing code, so it should be marked as deprecated first (and then dropped in the next major version later). It also looks like a good candidate to propose this step in a bug report.
> 
> That sounds reasonable to me.

So that commit won't be merged in the current form then, therefore it definitely helps to open an issue to be picked by anyone willing to implement it.

> Thank you for your helpful replies and all your work on Rex.

Thanks for the kind words and for contributing! :+1:

ferki:
I've rebased this on top of current default branch in order to get fresh coverage info about all the List::Util replacement locations.

ferki:
I think I now why List::MoreUtils was used in the first place: older Linux distributions with a very long support/maintenance schedule does not package recent enough versions of List::Util (e.g. Centos 6), so they also can't package Rex itself if we depend on that (and similar packages), while we would like to make sure Rex can be run on those systems.

I'm going to close this PR unmerged because of this really bring us into a territory about restructuring our support and/or deprecation policy, ant that is completely out of scope here. That being said, I'm sure this kind of modernization will land in Rex sooner or later (both dropping deps, using more modern alternatives, and deprecating old functions/modules).

Thanks again for the contribution, it really gives a lot of food for thought!

PR #1373, opened 2020-07-03
convert rexify.org/docs/guides/feature_flags.html to POD

randyl:
This PR fixes #1346  by creating a new file, lib/Rex/FeatureFlags.pod, that contains the feature-flag documentation from the website converted to POD.

Note that I moved the "How to enable Feature Flags" section above the list of feature flags, so that users don't have to scroll through a long list of flags to see example code.

Also, this PR does not address what to do with the feature-flag page on the website, but there is a separate ticket for that at https://github.com/RexOps/rexify-website/issues/119.

## Checklist

- [x] based on top of latest source code 
- [x] changelog entry included 
- [x] tests pass on Travis CI 
- [x] git history is clean    
- [x] git commit messages are [well-written](https://chris.beams.io/posts/git-commit/#seven-rules)

PR #1389, opened 2020-08-01
Document feature flags

### LINKED ISSUE #1346: Document feature flags
## User story

**As** a maintainer,
**I would like to** include the documentation of feature flags in the main codebase,
**so I can** make them more discoverable, and the distribution more self-contained.

## Describe the solution you'd like

Currently the [feature flags](https://www.rexify.org/docs/guides/feature_flags.html) are documented only on the website, which makes them harder to discover by relying on external documentation. Including this important bit of information in the main codebase (perhaps directly in the Rex module?) would make the distribution more self-contained.

The web page could either link to the main docs, or be generated directly from the POD docs.

## Describe alternatives you've considered

The authoritative source of this information is either included in the codebase or external to it. Internal feels mandatory, and external feels more optional, so at this point I'm not sure there are any real alternatives.

## Additional context

It might be a good addition to formalize and link to some kind of deprecation policy as well, since feature flags are a related concept.

ferki:
This PR fixes #1346 by moving the feature flags documentation into Rex core from the website.

This is a follow-up PR based on the original work on #1373 by @randyl.

@randyl: turns out I didn't need to modify on your commit too much to reach the original text. Please see the `fixup!` commit for my changes. If you don't mind, I'll squash these commits together before merging (and since that will rewrite git history a bit, I might rewrite the commit message to simply read "Document feature flags").

I'll add a few modifications on top of this, but this way we can keep crediting you for your original work in the git tree :)

## Checklist

- [x] tests pass on Travis CI 
- [x] git history is clean    

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -6,6 +6,7 @@ Revision history for Rex
  [BUG FIXES]
 
  [DOCUMENTATION]
+ - Document feature flags
 
  [ENHANCEMENTS]
 
--- a/lib/Rex.pm
+++ b/lib/Rex.pm
@@ -952,6 +952,211 @@ sub import {
   strict->import;
 }
 
+=head1 FEATURE FLAGS
+
+=head2 A word on backward compatibility
+
+Everyone knows the pain if something gets deprecated and one has to
+port his old (and stable) code to a new version of a library or a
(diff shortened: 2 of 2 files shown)


randyl:
Sounds good to me. Thanks.

PR #1399, opened 2020-08-15
When using source for a file, we don't create old_md5 for comparison

schobes:
This results in on_change always being triggered.

This PR fixes # by .

## Checklist

- [ ] tests pass on Travis CI 
- [x] git history is clean    

### DIFF
--- a/lib/Rex/Commands/File.pm
+++ b/lib/Rex/Commands/File.pm
@@ -543,6 +543,10 @@ sub file {
 
   }
   elsif ( exists $option->{"source"} && !$is_directory ) {
+    if ($need_md5) {
+      eval { $old_md5 = md5($file); };
+    }
+
     $option->{source} =
       Rex::Helper::Path::get_file_path( $option->{source}, caller() );
 


lib/Rex/Commands/File.pm
```diff
@@ -543,6 +543,10 @@ sub file {
 
   }
   elsif ( exists $option->{"source"} && !$is_directory ) {
+    if ($need_md5) {
+      eval { $old_md5 = md5($file); };
+    }
```
ferki, line 548: As you can see a few line below, this case is completely offloaded to the `upload` function. That does its own file identity checks via MD5 checksumming, and then includes it in the returned value whether it had to change the file or not.

I think checking the result of the upload function would be the correct way to handle this case, instead of having to calculate one more checksum.

ferki (changes requested):
Thanks for your contribution, @schobes!

It seems like you opened a pull request without opening an issue first to discuss and design a solution. Please make sure you read and understood our [Contributing guide](https://github.com/RexOps/Rex/blob/master/CONTRIBUTING.md) (note the [GitHub issues and pull requests](https://github.com/RexOps/Rex/blob/master/CONTRIBUTING.md#github-issues-and-pull-requests) section).

It also seems like this PR is incomplete, and without any followup activity. There's no changelog entry about the modification, and there's no test included to prove both the presence of the problematic behavior and the correctness of the proposed fix.

Since I'm in the process of preparing the next release, and such a file management bug would be important to fix, I'm going to triage the bug, and open an issue about it on my own, plus I'll attempt to include a test and fix for it.

I'll leave this PR open for now, but I might close it if I can come up with a complete solution on my own.

PR #1421, opened 2020-08-31
Skip procfs inventory test if /proc is not readable

### LINKED ISSUE #1420: procfs test fails on FreeBSDs
## Describe the bug

The recently added procfs test is failing on FreeBSDs because while `/proc` exists on those systems, it's not necessarily readable by the user running the tests.

edit: It looks like the problem is more like that `/proc/cpuinfo` doesn't exist on FreeBSD.

## How to reproduce it

1. run `prove -l -r t/proc.t` on a FreeBSD system with current master

~~Shortest code example that demonstrates the bug:~~

## Expected behavior

Test suite should gracefully skip in case of known problematic/unsupported scenarios.

## Circumstances

 - Rex version: 1.12.1.1-TRIAL / current master
 - Perl version: any
 - OS running rex: FreeBSD
 - OS managed by rex: local/FreeBSD
 - How rex was installed: trial release from CPAN

ferki:
This PR is an attempt to fix #1420 by checking if `/proc` is readable before testing inventory functionality related to procfs.

## Checklist

- [x] tests pass on Travis CI 
- [x] git history is clean 

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -10,6 +10,7 @@ Revision history for Rex
  - Fix registering custom template functions
  - Fix handling of configuration file settings
  - Skip MD5 binary test if there's no binary available
+ - Skip procfs inventory test if /proc is not readable
 
  [DOCUMENTATION]
  - Clarify contributing guide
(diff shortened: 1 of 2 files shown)


ferki:
The problem is more like that `/proc/cpuinfo` doesn't exist on FreeBSD. That would require a different fix.

PR #1430, opened 2020-09-14
Fix _binary_md5 function behavior with "strangely named files", #1429

### LINKED ISSUE #1429: Wrong _binary_md5 behavior on *BSD when filename looks like a md5 command output
## Describe the bug

Wrong `_binary_md5` behavior on *BSD when filename looks like an `md5` command output

## How to reproduce it

1. Create file like `md5test = 11111111111111111111111111111111`
2. Get MD5 from this file by `Rex::Commands::MD5::_binary_md5()`
3. The MD5 is wrong

Run this test on *BSD

```perl
use strict;
use warnings;

use Test::More tests => 1;
use Rex::Commands::MD5;

my $test_file_md5 = Rex::Helper::File::Spec->catfile( '/', 'tmp', 'md5test = 11111111111111111111111111111111' );

my $fh;
open($fh, '>', $test_file_md5) or
    BAIL_OUT("Can't create file $test_file_md5 $!");
print $fh "\x00";
close($fh);

SKIP: {
  skip 'No MD5 binary seems to be available', 1
    if !defined Rex::Commands::MD5::_binary_md5($test_file_md5);

  is(
    Rex::Commands::MD5::_binary_md5($test_file_md5),
    '93b885adfe0da089cdf634904fd59f71',
    'MD5 checksum OK via binary'
  );
}
```

## Expected behavior

We got right MD5 from file

## Circumstances

 - Rex version: 1.12.2
 - Perl version: any
 - OS running rex: any
 - OS managed by rex: OpenBSD, FreeBSD, OSX maybe
 - How rex was installed: any

## Debug log

none

Ilya33:
This PR is an attempt to fix #1429 by changing regexp in `Rex::Commands::MD5::_binary_md5` function.

## Checklist

- [x] tests pass on Travis CI 
- [x] git history is clean 


t/md5.t
```diff
@@ -1,11 +1,20 @@
...
 use warnings;
 
-use Test::More tests => 3;
+use Test::More tests => 4;
+use File::Temp qw(tempfile);
 use Rex::Commands::MD5;
 
 my $test_file = Rex::Helper::File::Spec->catfile( 't', 'md5test.bin' );
 
+my ( $fh, $strangely_named_file ) = tempfile(
```
ferki, line 10: First I thought it would be nice to use the same way for both test files (either commit both to the repo, or generate both during tests). But then I wondered if it would be enough to just use the new generated file, so we can get rid of the other one. Hmm, I don't see any obvious downsides to that yet :thinking: 

What do you think? Should we drop the old one, or keep both but handle them the same way (either generated or committed)?

ferki, line 10: Hmm, on second pass, I think it might be a good idea to add a short comment about why we name the file like this.

t/md5.t
```diff
@@ -1,11 +1,20 @@
...
+  "XXXXX",
+  SUFFIX => " = 11111111111111111111111111111111",
```
ferki, line 12: Nitpick: technically the double quotes are unnecessary since we don't interpolate anything between them. I try to use single quotes in this case for newly written code, but there's no rule to enforce it yet, so it's fine :+1:

t/md5.t
```diff
@@ -1,11 +1,20 @@
...
+  TMPDIR => 1
```
ferki, line 13: Looks like the temp files are not cleaned up after the testing is done, so I think we should be explicit about that by adding `UNLINK => 1` to the options.

ferki (changes requested):
Thanks for this PR, @Ilya33!

The commit messages seem to be a bit long to my taste, but the only major thing from the checklist is the missing changelog entry. Normally I would include a "Fix XY" item under the bugfix section for this, and would commit that together with the commit that contains the fix.

Apart form that, I've left some review notes to be addressed before merging, but overall I think it's in a quite good shape, and I'd like to include it soon in the next release.

I'd appreciate if you could update the PR based on the feedback items, but if you don't have time for that, I can also try to add follow-up commits to this branch before merging.

t/md5.t
ferki: Alright, I chose to have a single generated temporary file for MD5 tests.

t/md5.t
ferki: Since I have to touched the `tempfile()` arguments anyway, I also fixed this along the way.

t/md5.t
ferki: I added `UNLINK => 1,` to make sure the temporary file is removed upon exit. I also removed `TMPDIR => 1`, and `"XXXXX"` as the defaults seem to be enough (on Linux; awaiting test results from more platforms).

t/md5.t
ferki: I also included a comment about it.

ferki (approved):

PR #1441, opened 2020-10-11
Proposed path and test for issue #1401

### LINKED ISSUE #1401: Error executing commands with space in their full path
## Describe the bug

If there's a space in the full path of a command, execution via `run()` fails with:

```
[2020-08-19 13:12:48][<local>]  ERROR Error executing: /tmp/path with space/hello.
STDOUT:

STDERR:
sh: /tmp/path: No such file or directory at /home/ferki/perl5/perlbrew/perls/ferki/lib/site_perl/5.32.0/Rex/Commands/Run.pm line 326.
```

This kind of error may be more apparent on Windows, where calling `can_run()` returns the full path of the executable, which is often under `C:\Program Files`, which has a space in it.

Without looking into it more closely, it might be a shell quoting issue.

## How to reproduce it

1. Create a path with space(s) in it

    ```
    mkdir -p /tmp/path\ with\ space
    ```

2. Create an executable under that path, e.g. named `hello`:

    ```bash
    #!/bin/bash
    echo hello
    ```

3. Run that command with rex: `rex -e 'my $command = "/tmp/path with space/hello"; say run $command'`

Shortest code example that demonstrates the bug: see above

## Expected behavior

Rex should be able to execute a command even if its full path contains spaces (or possibly other special characters).

## Circumstances

 - Rex version: 1.12.1
 - Perl version: 5.32.0
 - OS running rex: Gentoo
 - OS managed by rex: local/Gentoo

## Debug log

Relevant excerpt is the following:

```
[2020-08-19 13:29:14][<local>]  DEBUG Executing eval-line
[2020-08-19 13:29:14][<local>]  DEBUG Executing: export PATH=/bin:/sbin:/usr/bin:/usr/sbin:/usr/local/bin:/usr/local/sbin:/usr/pkg/bin:/usr/pkg/sbin ; LC_ALL=C /tmp/path with space/hello
[2020-08-19 13:29:14][<local>]  DEBUG ========= ERR ============
[2020-08-19 13:29:14][<local>]  DEBUG sh: /tmp/path: No such file or directory

[2020-08-19 13:29:14][<local>]  DEBUG ========= ERR ============
```

jluis:
This PR solves the issue #1401 on commands with explicit paths *nix stile.

The added code only acts on  commands that start with "/", "./", "../" or "~/"
it quotes with '"' from the start of the command to the first space after the last slash.
Command strings like '/bin/ls /etc/host /etc/hostname' will fail unless the command is  properly quoted or preceded with an space or an envirment variable assinament but I think it will resolve most of the cases

---

EDIT by @ferki:

Let's add this section to follow the original PR template which helps to track the state of the PR:

This PR is an attempt to fix #1401 by finding the longest path prefix of commands passed to `run`, and then quoting them.

## Checklist

- [ ] tests pass on Travis CI 
- [ ] git history is clean 

### DIFF
--- a/lib/Rex/Commands/Run.pm
+++ b/lib/Rex/Commands/Run.pm
@@ -184,7 +184,11 @@ sub run {
     if !exists $option->{auto_die};
 
   # check if $cmd is an eplicit path
-  if ( $cmd =~ /(^(?:[.]{1,2}|[~])?[/].+[/])(.+)/xsm ) {
+  if ( $cmd =~ m{(^
+                 (?:[.]{1,2}|[~])?
+                 [/].+[/])
+                (.+)
+               }xsm ) {
     my ( $head, $tail ) = ( $1, $2 );
     if ( $tail =~ /(.+?)([ ].*)/sxm ) {
       $cmd = qq("$head$1"$2);
--- a/lib/Rex/Commands/Run.pm
+++ b/lib/Rex/Commands/Run.pm
@@ -184,9 +184,9 @@ sub run {
     if !exists $option->{auto_die};
 
   # check if $cmd is an eplicit path
-  if ( $cmd =~ /(^(?:\.{1,2}|\~)?\/.+\/)(.+)/xsm ) {
+  if ( $cmd =~ /(^(?:[.]{1,2}|[~])?[/].+[/])(.+)/xsm ) {
     my ( $head, $tail ) = ( $1, $2 );
-    if ( $tail =~ /(.+?)(\ .*)/sxm ) {
+    if ( $tail =~ /(.+?)([ ].*)/sxm ) {
       $cmd = qq("$head$1"$2);
     }
     else {
--- a/lib/Rex/Commands/Run.pm
+++ b/lib/Rex/Commands/Run.pm
@@ -190,8 +190,13 @@ sub run {
   if (  ( $cmd =~ m{^(?:[.]{1,2}|[~]|[A-Z]:|[\\/])}xsm )
     and ( $cmd !~ m{[\\^]\s|["']}xsm ) )
   {
-    my $split = qr{(?=\s(?:[.]{1,2}|[~]|[A-Z]:|[\\/]))}xsm;
-    my @cmd   = split $split, $cmd;
+
+    #https://stackoverflow.com/questions/4094699\
+    #/how-does-the-windows-command-interpreter-cmd-exe-parse-scripts\
+    #/4095133#4095133
+    my $split = qr{(?=\s(?:[.]{1,2}|[~]|[A-Z]:|[\\/])|\b[\d]?[()?*[\]&|<>])}xsm;
+
+    my @cmd = split $split, $cmd;
     $cmd[0] =~ s{(.*[\\/][^\s]*)}{"$1"}xsm;
     $cmd = join q{}, @cmd;
   }
--- a/lib/Rex/Commands/Run.pm
+++ b/lib/Rex/Commands/Run.pm
@@ -188,9 +188,9 @@ sub run {
 
   if ( $cmd =~ m{^(?:[.]{1,2}|[~]|[A-Z]:|[\\/])}xsm ) {
     my $split = qr{(?=\s(?:[.]{1,2}|[~]|[A-Z]:|[\\/]))}mxs;
-    my @cmd = split $split, $cmd;
-    $cmd[0] = qq{"$cmd[0]"} ;
-    $cmd = join q{}, @cmd;
+    my @cmd   = split $split, $cmd;
+    $cmd[0] = qq{"$cmd[0]"};
+    $cmd    = join q{}, @cmd;
   }
 
   my $res_cmd = $cmd;
--- a/lib/Rex/Commands/Run.pm
+++ b/lib/Rex/Commands/Run.pm
@@ -182,12 +182,12 @@ sub run {
   $option->{auto_die} = Rex::Config->get_exec_autodie()
     if !exists $option->{auto_die};
 
-#  check if $cmd is an eplicit path 
-   if ( $cmd =~ m"(^(?:\.{1,2}|\~)?/.+/)(.+)" ) {
-     my ($head,$tail) = ($1,$2);
-     my $end = ($tail =~ s/(.+?)( .+)/$1/ )?$2:'';
-     $cmd = '"'.$head.$tail.'"'.$end #quote the path of comand in a compatible way
-   }
+  # check if $cmd is an eplicit path
+  if ( $cmd =~ m"(^(?:\.{1,2}|\~)?/.+/)(.+)" ) {
+    my ( $head, $tail ) = ( $1, $2 );
+    my $end = ( $tail =~ s/(.+?)( .+)/$1/ ) ? $2 : '';
+    $cmd = '"' . $head . $tail . '"' . $end # quote the path of comand in a compatible way
+  }
 
   my $res_cmd = $cmd;
 
--- a/lib/Rex/Commands/Run.pm
+++ b/lib/Rex/Commands/Run.pm
@@ -186,19 +186,21 @@ sub run {
   # We only process cmd is an explicit path that means
   # that starts with . .. <letter>: and / or \
   # and aren't scaped spaces or quotes
-
-  if (  ( $cmd =~ m{^(?:[.]{1,2}|[~]|[A-Z]:|[\\/])}xsm )
-    and ( $cmd !~ m{[\\^]\s|["']}xsm ) )
   {
+    my $path_start = qr{(?:[.]{1,2}|[~]|\p{IsAlphabetic}:|[\\/])}msx;
+    if ( ( $cmd =~ m{^$path_start}xsm )
+      && ( $cmd !~ m{[\\^]\s|["']}xsm ) )
+    {
 
-    #https://stackoverflow.com/questions/4094699\
-    #/how-does-the-windows-command-interpreter-cmd-exe-parse-scripts\
-    #/4095133#4095133
-    my $split = qr{(?=\s(?:[.]{1,2}|[~]|[A-Z]:|[\\/])|\b[\d]?[()?*[\]&|<>])}xsm;
+      #https://stackoverflow.com/questions/4094699\
+      #/how-does-the-windows-command-interpreter-cmd-exe-parse-scripts\
+      #/4095133#4095133
+      my $split = qr{(?=\s$path_start|\b\d?[()?*[\]&|<>])}xsm;
 
-    my @cmd = split $split, $cmd;
-    $cmd[0] =~ s{(.*[\\/][^\s]*)}{"$1"}xsm;
-    $cmd = join q{}, @cmd;
+      my @cmd = split $split, $cmd;
+      $cmd[0] =~ s{(.*[\\/]\S*)}{"$1"}xsm;
+      $cmd = join q{}, @cmd;
+    }
   }
 
   my $res_cmd = $cmd;
--- a/lib/Rex/Commands/Run.pm
+++ b/lib/Rex/Commands/Run.pm
@@ -183,10 +183,14 @@ sub run {
     if !exists $option->{auto_die};
 
   # check if $cmd is an eplicit path
-  if ( $cmd =~ m"(^(?:\.{1,2}|\~)?/.+/)(.+)" ) {
+  if ( $cmd =~ /(^(?:\.{1,2}|\~)?\/.+\/)(.+)/xsm ) {
     my ( $head, $tail ) = ( $1, $2 );
-    my $end = ( $tail =~ s/(.+?)( .+)/$1/ ) ? $2 : '';
-    $cmd = '"' . $head . $tail . '"' . $end # quote the path of comand in a compatible way
+    if ( $tail =~ /(.+?)(\ .*)/sxm ) {
+      $cmd = qq("$head$1"$2);
+    }
+    else {
+      $cmd = qq("$head$tail");
+    }
   }
 
   my $res_cmd = $cmd;
--- a/lib/Rex/Commands/Run.pm
+++ b/lib/Rex/Commands/Run.pm
@@ -182,6 +182,13 @@ sub run {
   $option->{auto_die} = Rex::Config->get_exec_autodie()
     if !exists $option->{auto_die};
 
+#  check if $cmd is an eplicit path 
+   if ( $cmd =~ m"(^(?:\.{1,2}|\~)?/.+/)(.+)" ) {
+     my ($head,$tail) = ($1,$2);
+     my $end = ($tail =~ s/(.+?)( .+)/$1/ )?$2:'';
+     $cmd = '"'.$head.$tail.'"'.$end #quote the path of comand in a compatible way
+   }
+
   my $res_cmd = $cmd;
 
   if ( exists $option->{only_notified} && $option->{only_notified} ) {
--- a/lib/Rex/Commands/Run.pm
+++ b/lib/Rex/Commands/Run.pm
@@ -183,22 +183,14 @@ sub run {
   $option->{auto_die} = Rex::Config->get_exec_autodie()
     if !exists $option->{auto_die};
 
-  # check if $cmd is an eplicit path
-  if (
-    $cmd =~ m{(^
-                 (?:[.]{1,2}|[~])?
-                 [/].+[/])
-                (.+)
-               }xsm
-    )
-  {
-    my ( $head, $tail ) = ( $1, $2 );
-    if ( $tail =~ /(.+?)([ ].*)/sxm ) {
-      $cmd = qq("$head$1"$2);
-    }
-    else {
-      $cmd = qq("$head$tail");
-    }
+  # We only process cmd is an explicit path that means
+  # that starts with . .. <letter>: and / or \
+
+  if ( $cmd =~ m{^(?:[.]{1,2}|[~]|[A-Z]:|[\\/])}xsm ) {
+    my $split = qr{(?=\s(?:[.]{1,2}|[~]|[A-Z]:|[\\/]))}mxs;
+    my @cmd = split $split, $cmd;
+    $cmd[0] = qq{"$cmd[0]"} ;
+    $cmd = join q{}, @cmd;
   }
 
   my $res_cmd = $cmd;
--- a/lib/Rex/Commands/Run.pm
+++ b/lib/Rex/Commands/Run.pm
@@ -184,11 +184,14 @@ sub run {
     if !exists $option->{auto_die};
 
   # check if $cmd is an eplicit path
-  if ( $cmd =~ m{(^
+  if (
+    $cmd =~ m{(^
                  (?:[.]{1,2}|[~])?
                  [/].+[/])
                 (.+)
-               }xsm ) {
+               }xsm
+    )
+  {
     my ( $head, $tail ) = ( $1, $2 );
     if ( $tail =~ /(.+?)([ ].*)/sxm ) {
       $cmd = qq("$head$1"$2);
--- a/lib/Rex/Commands/Run.pm
+++ b/lib/Rex/Commands/Run.pm
@@ -185,12 +185,15 @@ sub run {
 
   # We only process cmd is an explicit path that means
   # that starts with . .. <letter>: and / or \
+  # and aren't scaped spaces or quotes
 
-  if ( $cmd =~ m{^(?:[.]{1,2}|[~]|[A-Z]:|[\\/])}xsm ) {
-    my $split = qr{(?=\s(?:[.]{1,2}|[~]|[A-Z]:|[\\/]))}mxs;
+  if (  ( $cmd =~ m{^(?:[.]{1,2}|[~]|[A-Z]:|[\\/])}xsm )
+    and ( $cmd !~ m{[\\^]\s|["']}xsm ) )
+  {
+    my $split = qr{(?=\s(?:[.]{1,2}|[~]|[A-Z]:|[\\/]))}xsm;
     my @cmd   = split $split, $cmd;
-    $cmd[0] = qq{"$cmd[0]"};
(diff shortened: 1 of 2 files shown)


ferki:
Thanks @jluis for working on it! I think this draft PR is a good way to explore all the details of the question.

The test suite should be able to check the code on Linux, Mac and Windows as well. I'd expect Rex should be able to quote paths correctly on all three platforms. (especially Windows, because of `C:\Program Files`). Custom quoting logic is one way to do it, but I believe `Net::OpenSSH::ShellQuoter` might be able to help with most of that (not sure about Windows, though).

There are also checks to make sure we don't introduce more perlcritic issues than what is already there. If the checks are green, mark it as ready to review. If you need help to make some of the tests pass, let me know!

jluis:
I'm looking on the now the perlcritic tests as they pass on my local machine.
I'm planing to test on windows later with strawberry perl .
Thanks for pointing to the  Net::OpenSSH quoter T think the solution I propose is compatible with it,

ferki:
> I'm looking on the now the perlcritic tests as they pass on my local machine.

They should indicate that this branch has more perlcritic violations than the current master branch. The goal is to have fewer or equal violations both per policy _and_ overall (that is to avoid the codebase getting any worse in this regard) with each pull request.

To test locally:

- make sure there's no `xt/author/.perlcritic-history` file already saved
- generate a new history file as a reference point on the master branch by running the progressive perlcritic tests with `AUTHOR_TESTING=1 prove xt/author/critic-progressive.t`
- switch branch and run the same test - it should fail only when there are more perlcritic violations on the branch than on master

> I'm planing to test on windows later with strawberry perl .

Sounds great, thank you! Sometimes the Windows on Travis CI has different results than a real Windows system, so it is very useful to test it directly.

> Thanks for pointing to the Net::OpenSSH quoter T think the solution I propose is compatible with it,

:+1:

ferki:
Hey @jluis, I see it's shaping up nicely, and now it passes tests! :+1: 

I noticed you skip all tests on Windows now, while I believe this fix would be the most interesting fix there, since many commands are located under `C:\Program Files` and now they can't be executed with their full path.

What do you think, how hard would it be to include Windows support as well? Is there anything particular missing for that? I believe Travis CI could help with testing at least.

jluis:
@ferki  I'm on my  way to allow it and make some changes on path detection  that is the only part that needs to be quoted but I wanted to be sure that I have the coding stile OK.
I cannot build the distribution it on my windows machine so I first build my system  and then install and test on windows.

ferki:
@jluis sounds great, thanks! Feel free to experiment here, and I can help with the review of the final version and merging.

ferki:
Hmm, I'm afraid this might be getting more complicated than expected or necessary.

What I would have expected is to also use `Net::OpenSSH::ShellQuote` on `$cmd`, not just on `@{$args}`, [here](https://github.com/RexOps/Rex/blob/1.13.0.2/lib/Rex/Commands/Run.pm#L259):

```perl
my $quoter = Net::OpenSSH::ShellQuoter->quoter( $exec->shell->name );
$cmd = "$cmd " . join( " ", map { $quoter->quote($_) } @{$args} );
```

That might break cases when the arguments and the command are separated with spaces too, like `$cmd = 'C:\Program Files\uname.exe -s'`. For that, I think I would try splitting `$cmd` on spaces (similarly how you do), and do a prefix check with `can_run` (answering "is it in $PATH?"):

- command is `C:\Program Files\uname.exe -s`
- take first split: `C:\Program` => `can_run('C:\Program')? => no, get the next one
- take next split: `C:\Program Files\uname.exe` => `can_run('C:\Program Files\uname.exe')` => yes, this is my real command
- quote it with `Net::OpenSSH::ShellQuoter`

What do you think? Could that work? I can try writing tests for that too.

ferki:
FTR: I started to play with the above idea in another branch, so I see why it's tricky now :)

jluis:
Yes its more complicated that it looks so I changed the algorithm to detect the path:
1 if is a quote on the cmd or an escape (\ and also ^ on windows ) do nothing
2 split cmd on 0 length assertions that for sure not form part of the command 
3 quote the first element from the start of the path to the first space after the last path separator or the end of element 
4 join the parts

I think that the can run aproach is factible  but would it leak information of the remote?

ferki:
> I think that the can run aproach is factible but would it leak information of the remote?

What is more of a problem with the `can_run` approach, is that it only checks if the queried executable is in $PATH (or %PATH%), by asking `which` (or `where`). So a custom command outside that path might be executable, but it would not find it.

I like your approach with zero-length assertions and finding the "first space after the last path separator" better for this purpose :+1:

ferki (changes requested):
I see you marked it as ready for review, but it is still failing on Linux and Mac (works on Windows though :tada: ).

Let's ignore the perlcritic warnings for now, and I'll help cleaning it up once it is working on all platforms.

jluis:
I did it before the  tests ended as it was working on my linux & windows systems.
When I saw it was failing I could not undo the action.
I'm using now testing how the commands are processed on win and linux using a simple:
perl -e"say qx(..)" and just writing then on the command line but I will do it on an alternate branch to no pollute the Travis tests here

ferki:
Thanks @jluis for your continued work on this important fix! I see that now it passes tests except the perlcritic ones, so I'll continue with a full review in the coming release cycle, and see what's needed to get it merged :+1: I expect to be able to do any leftover cleanup steps on my own too.

ferki:
Thanks @jluis, for the followups! I've started to clean the commit history up in a separate branch, and I'll ask for your feedback before going forward with it. In these days I'm busy with `$work`, but in Rex context this is at the top if the priorities for me, so I will address this among firsts when I have capacity for open source again.

jluis:
Thanls for the feedbak .
I think that the test is now clean on perlcritic bases  I run it on both files and  only saw two in the test saying that it don't start with a package declaration and that it does not end with 1;
and on Run.pm none of the advises are in the line range of the path.
So I will Keep working on other branch of my fork to reduce noise here an documenting some differences between the different shells.

t/issue/1401.t
```diff
@@ -0,0 +1,86 @@
...
+}
+
+sub relative {
+  return qq{./$_[0]};
+}
+
+my $path    = File::Temp->newdir( 'path with spaces XXXX', DIR => q{./} );
+my $tail    = 'hello';
+my $parm    = 'Rex is wonderfull';
+my $testcmd = '/bin/echo';
```
ferki, line 32: Would this exist on all platform where Rex might be run or what Rex might be asked to manage? Could it be that this is located elsewhere on special/ancient/etc. systems (like `/usr/bin/echo`)?

Can we avoid hardcoding the path here, e.g. by discovering the path to `echo` with `can_run('echo')`?

Also, perhaps `$path_to_echo` is a more readable variable name here.

t/issue/1401.t
```diff
@@ -0,0 +1,86 @@
...
+my $s       = run command relative($path), $tail, $parm;
```
ferki, line 33: What does `$s` mean here?

t/issue/1401.t
```diff
@@ -0,0 +1,86 @@
...
+  my ( $path, $exe, @parm ) = @_;
+  return qq{$path/$exe } . join $empty, @parm;
+}
+
+sub relative {
+  return qq{./$_[0]};
+}
+
+my $path    = File::Temp->newdir( 'path with spaces XXXX', DIR => q{./} );
+my $tail    = 'hello';
```
ferki, line 30: So `$tail` is the actual command name we would like to create later. Perhaps `$test_command_name` is a better variable name then?

t/issue/1401.t
```diff
@@ -0,0 +1,86 @@
...
+my $parm    = 'Rex is wonderfull';
```
ferki, line 31: I find `$parm` harder to read than it should be; I associate it with parmesan and parma ham, not with param(eter), or argument :D

Since this is used as an argument for our testing commands, perhaps `$argument` would be a better name for readability.

t/issue/1401.t
```diff
@@ -0,0 +1,86 @@
...
+my $space = q{ };
+my $empty = q{};
+
+sub command {
+  my ( $path, $exe, @parm ) = @_;
+  return qq{$path/$exe } . join $empty, @parm;
+}
+
+sub relative {
+  return qq{./$_[0]};
```
ferki, line 26: `File::Spec->join()`

t/issue/1401.t
```diff
@@ -0,0 +1,86 @@
...
+}
+
+my $path    = File::Temp->newdir( 'path with spaces XXXX', DIR => q{./} );
+my $tail    = 'hello';
+my $parm    = 'Rex is wonderfull';
+my $testcmd = '/bin/echo';
+my $s       = run command relative($path), $tail, $parm;
+
+#on windows if the path not exists not apears on the output
+my $not_exists = ( $win && $s !~ /(?-x:$path)/smx ) ? $s : $path;
```
ferki, line 36: I don't think I fully understand the logic here.

On Windows, trying to execute a non-existing command seems to return the string `The  system cannot find the path specified.`. Would it be better to test for that on Windows?

t/issue/1401.t
```diff
@@ -0,0 +1,86 @@
...
+
+like( $s, qr{(?-x:$not_exists)}smx, q{inexistent command gives an error} );
+
+my $cmd = relative qq{$path/$tail};
```
ferki, line 40: `File::Spec->join()`

t/issue/1401.t
```diff
@@ -0,0 +1,86 @@
...
+  $cmd .= q{.bat};
+  open my $hello, q{>}, $cmd;
+  print {$hello} "echo %*\n";
+  close $hello;
+}
+else {
+  #hello is a sybolic link to echo
+  if ( !-X $testcmd ) {
+    substr $testcmd, 0, 0, q{/usr};
+  }
```
ferki, line 54: If I understand correctly this is exactly to handle the case I noted above (`echo` is not under `/bin`, but under `/usr/bin`). Let's get rid of this logic here by discovering the real path above, and relying on that.

t/issue/1401.t
```diff
@@ -0,0 +1,86 @@
...
+  symlink $testcmd, $cmd;
+}
+
+my $result = qr{^(?-x:$parm)}smx;
```
ferki, line 58: If we could use `is` instead of `like` to test the output, we could drop this variable, and just use `$parm` directly below (or perhaps it needs a trailing newline, but still).

t/issue/1401.t
```diff
@@ -0,0 +1,86 @@
...
+
+$s = run command( relative($path), $tail, $parm );
```
ferki, line 60: This is reusing a previous variable which is potentially dangerous if `run` fails internally somehow. Let's put these into their separate scopes, or subtests.

t/issue/1401.t
```diff
@@ -0,0 +1,86 @@
...
+like( $s, $result, q{path with spaces parse ok} );
```
ferki, line 61: Could we use `is` instead of `like`? Would we gain more by testing for exact match?

t/issue/1401.t
```diff
@@ -0,0 +1,86 @@
...
+like( $s, qr{(?-x:$not_exists)}smx, q{inexistent command gives an error} );
+
+my $cmd = relative qq{$path/$tail};
+
+if ($win) {
+
+  #On windows we create hello.bat contaning echo %1
+  $cmd .= q{.bat};
+  open my $hello, q{>}, $cmd;
+  print {$hello} "echo %*\n";
```
ferki, line 47: Let's check for successful file write with `print .. or croak ...` here. It could catch cases when the disk is full, or other troubles when we can't actually write to the disk. I'm sure CPAN Testers would have a weird edge case sooner or later, and it might be hard to debug.

t/issue/1401.t
```diff
@@ -0,0 +1,86 @@
...
+
+$s = run command( relative($path), $tail, $parm, q{ /extra} );
+like( $s, $result, 'an " /"  on the parmeters parse ok' );
+
+$s = run command( $quote . relative($path), $tail . $quote, $parm );
+like( $s, $result, 'Quoted commands ok' );
+
+#Partialy quoted paths don't work on windows
+#$s = run command( relative(qq{"$path"}), $tail, $parm );
+#like( $s, $result, 'Partially quoted' );
```
ferki, line 71: I think we might be able to put it into a `TODO` block, so it doesn't mean test failure, but we would notice it when it gets fixed.

t/issue/1401.t
ferki: We might not need the `relative` subroutine, if we `cwd` into the parent of the temp directory first, like this:

```perl
run $command, cwd => dirname($path);
```

t/issue/1401.t
```diff
@@ -0,0 +1,86 @@
...
+
+my $mpath = join $empty, relative($path), q{/}, $tail;
+
+if ($win) {
+  $mpath =~ s{/}{\\}gsmx;
```
ferki, line 76: I think if we would use `File::Spec->join()`, we would not need this kind of magic replacement of path separators.

t/issue/1401.t
```diff
@@ -0,0 +1,86 @@
...
+like( $s, $result, 'an " /"  on the parmeters parse ok' );
+
+$s = run command( $quote . relative($path), $tail . $quote, $parm );
+like( $s, $result, 'Quoted commands ok' );
+
+#Partialy quoted paths don't work on windows
+#$s = run command( relative(qq{"$path"}), $tail, $parm );
+#like( $s, $result, 'Partially quoted' );
+
+my $mpath = join $empty, relative($path), q{/}, $tail;
```
ferki, line 73: What does `$mpath` mean here?

ferki (changes requested):
Alright, I finally could sit down and force myself through most of it :)

TL;DR: this is good, but I need to update many little details before I can merge it for long-term maintenance.

Longer version:

Overall I think the logic is good (and works! :tada: ), and this is quite close to be merged; I definitely want to include it in the next release (and have a development release ASAP).

I've left quite a few comments, mostly to remind myself to not forget to look into that again during my postprocessing. A bit more details on these:

- I can squash away all the unnecessary experimental/development/follow-up commits from the history before merging
- I can fix any remaining perlcritic violations
- I can fix most of the refactoring that I wish to be there before merging (especially annotating the complex regular expressions)
- I can fix most of the potential bugs (e.g. Windows drive letter matching)
- I find the test especially hard to read and understand, so I'll try to overhaul it to be simpler and to avoid potential bugs (especially because not using `File::Spec->join()` to concatenate path elements)

I've pushed my updates to the `path_with_space` branch. We can review it there together before updating this PR. Until then, please don't change this PR anymore as that would complicate the remaining steps considerably.

ferki:
Two things came out during testing:

- it quotes `~/bin/cmd` into `"~/bin/cmd"`, and thus the commands starting with a tilde are now failing because the tilde expansion in the shell doesn't happen anymore
- all the logic happens before processing the `command` option, so that use case is still broken (have to move the logic after `$option->{command}` is retrieved)

lib/Rex/Commands/Run.pm
```diff
@@ -183,6 +183,26 @@ sub run {
   $option->{auto_die} = Rex::Config->get_exec_autodie()
     if !exists $option->{auto_die};
 
+  # We only process cmd is an explicit path that means
+  # that starts with . .. <letter>: and / or \
+  # and aren't scaped spaces or quotes
+  {
```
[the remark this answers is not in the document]
ferki: Also, do we need to always run this logic, or only when there's a space in the command?

ferki:
I'm considering to put the logic into its own subroutine, and test that subroutine's output with various input. Emulating a fully featured command environment with a matrix of absolute/relative/tilde/etc. path cases on Windows and non-Windows plus all of that with different styles of `run()` invocations doesn't feel efficient, if feasible. Moreover the command name itself may have spaces, not just the path leading to it.

I'm thinking about mappings like this:

- non-Windows:
  - path with space:
    - `/tmp/path with space/cmd` => `'/tmp/path with space/cmd`'
    - `./path with space/cmd` => `'./path with space/cmd'`
    - `~/path with space/cmd` => `~/'path with space/cmd'`
  - optionally also the command has spaces:
    - `/tmp/path with space/cmd with space` => `'/tmp/path with space/cmd with space`'
    - `./path with space/cmd with space` => `'./path with space/cmd with space'`
    - `~/path with space/cmd with space` => `~/'path with space/cmd with space'`
  
- Windows:
  - `C:\Program Files\cmd` => `"C:\Program Files\cmd"`
  - `.\Program Files\cmd` => `".\Program Files\cmd"`
  - `~\cmd` => `~\cmd` (noop)

It might also be the case that we can't support all the situations, and we just have to let the user quote their input properly.

ferki:
Speculation:

Based on the complexity to correctly achieve this goal for all the possible corner cases, I wonder if it would be a wiser option to state that the user is expected to quote their commands properly based on the requirements of their specific use case.

Or to reduce the risks of having to maintain it long term, we could also consider to convert this quoting logic into an experimental function to be called by the user when they think it's needed. This approach would enable this function to evolve on this own over time, instead of making us work hard to address all corner cases without having an opt-out from it.

This opt-in approach could be something like:

```perl
my $command = 'C:\Program Files\command.exe';

run quote_command($command);
run quote_command($command, target => 'windows');
```

`target` could be optional, and may be auto-detected based on the OS of the endpoint where the command is supposed to be run.

lib/Rex/Commands/Run.pm
```diff
@@ -183,6 +183,26 @@ sub run {
...
+    my $path_start = qr{(?:[.]{1,2}|[~]|\p{IsAlphabetic}:|[\\/])}msx;
```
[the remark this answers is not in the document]
jluis: I wrote it this way to evade a perlcritic policy i think my first regex was something similar to  qr{(?:[.]{1,2}|[~]|[a-z]:|[\\/])}imsx

lib/Rex/Commands/Run.pm
```diff
@@ -183,6 +183,26 @@ sub run {
...
+    if ( ( $cmd =~ m{^$path_start}xsm )
+      && ( $cmd !~ m{[\\^]\s|["']}xsm ) )
+    {
+
+      #https://stackoverflow.com/questions/4094699\
+      #/how-does-the-windows-command-interpreter-cmd-exe-parse-scripts\
+      #/4095133#4095133
+      my $split = qr{(?=\s$path_start|\b\d?[()?*[\]&|<>])}xsm;
```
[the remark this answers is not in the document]
jluis: I think I get your view it could be better named $marks_path_end_regex is to discriminate thinks like "ls |grep regex" or "/path with spaces/command 2>/dev/null"

lib/Rex/Commands/Run.pm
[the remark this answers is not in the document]
jluis: The code block is here oniy to encapsulate the variables but as we need only to run it if there are spaces it could be better write as if ($cmd =~ /\ /msx ) {

lib/Rex/Commands/Run.pm
```diff
@@ -183,6 +183,26 @@ sub run {
...
+
+      my @cmd = split $split, $cmd;
```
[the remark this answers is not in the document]
jluis: I think you are right it is probable that the first command could contain an space but the probability that  the right side of a pipe or a redirection could contain it

lib/Rex/Commands/Run.pm
```diff
@@ -183,6 +183,26 @@ sub run {
...
+      $cmd[0] =~ s{(.*[\\/]\S*)}{"$1"}xsm;
```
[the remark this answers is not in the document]
jluis: I didn't test for it but if we wish it works the same way on both systems we had to quote it with single quotes to prevent variable expansion

lib/Rex/Commands/Run.pm
```diff
@@ -183,6 +183,26 @@ sub run {
...
+      $cmd = join q{}, @cmd;
```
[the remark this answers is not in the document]
jluis: Ok But if I let it as "" it trigered another perlcritic isue thanks for the pointer a

t/issue/1401.t
jluis: You are right its that most of the times I cannot assume or install things on machine and I involuntarily code the most things from the base instead of search the available functions or objects in the system

t/issue/1401.t
jluis: I tend to use generic names but yours is more descriptive

t/issue/1401.t
jluis: The problem is that i don't know how force windows use a specific locale or canonical form for the messages.It also reports different messages on if the path exist but no the command or if both don't exist (in one of then it reports the command you send but I don't remember what).

t/issue/1401.t
jluis: Ok I'll work on it

t/issue/1401.t
jluis: This is to test that it works with windows path separators but I will work alternatives

t/issue/1401.t
jluis: is just the full path with the command to  change the path separator to windows on the windows test and to escape the spaces on the other systems

ferki:
I believe this problem might not be possible to solve reliably in a generic way for all possible combinations of shells and operating systems.

Therefore I think it's currently best to say that proper quoting is the duty of the user. Generally speaking, they are in the best position to be aware of the actual details of their own use cases.

That being said, it might still worth to work on exporting the logic into a `quote_command()` or similar function. That could work best as an external module, like `Rex::Commands::ShellQuoter`, or perhaps even contributed into `Net::OpenSSH::ShellQuoter`. After all, this kind of quoting logic might be useful outside Rex as well, and I would also prefer to keep maintenance of such logic outside Rex.

Because of the focus, complexity, and stalled activity, I'm closing this PR for now. Nevertheless, thank you for your pathfinding work on this. Thanks to you, I learned a lot about this problem space!

PR #1447, opened 2020-10-30
Remove documentation for non-existant autodie feature

### LINKED ISSUE #1439: Feature flag "autodie" mentioned in documentation does not exist
## Describe the bug

The page https://metacpan.org/pod/Rex mentions "autodie" under feature flags, but this flag seems to no longer exist.

## How to reproduce it

1. try activating autodie flag
2. rex exits with error

```bash

perl -E 'use Rex -feature => [qw[autodie]];'

```

## Expected behavior

If the documentation states that autodie feature is available, then it should be available.

## Circumstances

 - Rex version: 1.10.0
 - Perl version: 5.26.2
 - OS running rex: Debian Sid
 - OS managed by rex: not applicable
 - How rex was installed: cpanm Rex

## Debug log

```
% cat Rexfile
use Rex -feature => ['autodie'];
% rex -T -d  
[2020-10-06 12:34:57] DEBUG - This is Rex version: 1.10.0
[2020-10-06 12:34:57] DEBUG - Command Line Parameters
[2020-10-06 12:34:57] DEBUG - 	T = 1
[2020-10-06 12:34:57] DEBUG - 	d = 1
[2020-10-06 12:34:57] DEBUG - Found Rexfile.lock
[2020-10-06 12:34:57] DEBUG - Found stale lock file. Removing it.
[2020-10-06 12:34:57] DEBUG - Creating lock-file (Rexfile.lock)
[2020-10-06 12:34:57] DEBUG - Loading Rexfile
[2020-10-06 12:34:57] WARN - You tried to load a feature (autodie) that doesn't exists in your Rex version. Please update.
%
```

nfg:
This PR is an attempt to fix #1439 by updating the feature documentation.

More generally, I removed the explicit "autodie feature" item. I merged the description of what it does into the "0.56" feature text. As for "no_autodie", I changed the text "autodie feature" to "autodie" so it doesn't imply that there's a feature called "autodie".

## Checklist

- [ ] tests pass on Travis CI 
- [ ] git history is clean 


nfg:
I think this is good! There aren't any new tests to write, since the change was entirely in the POD. The tests pass (I ran perltidy on the source file. Perlcritic is angry, but mostly about using quotation marks for non-interpolated strings vs apostrophes. xt/author/critic-progressive.t passes.

Let me know if there's anything else I should do, eh!

nfg:
(I've marked most of the checkboxes, but I'll leave the judgment whether the commit message is well-written to whoever reviews.)

ferki:
> I think this is good! There aren't any new tests to write, since the change was entirely in the POD. The tests pass.

Yup, I agree that no further explicit tests needed for POD-only updates in general. That being said, POD are still tested e.g. for spelling (and POD coverage tests planned for later too), so passing tests are always nice to see ;)

> I ran perltidy on the source file.

Great! AFAIK, that should not touch POD though, so we should be fine (and tests would have complained anyway). Thanks for erring on the side of safety!

> Perlcritic is angry, but mostly about using quotation marks for non-interpolated strings vs apostrophes. xt/author/critic-progressive.t passes.

Yes there are sadly quite a few violations accumulated over the last decade, so that's expected. The progressive perlcritic tests are in place to make sure that it doesn't get any worse (test passes if PR has less or equal violations both in total _and_ per-policy compared to the last saved state). Should be replaced with classic perlcritic tests once all of them are cleaned up.

> (I've marked most of the checkboxes, but I'll leave the judgment whether the commit message is well-written to whoever reviews.)

Thanks, the checkboxes are serving as a guide for both contributors and reviewers. I can assure you that the commit message is nice! :heart:

ferki (changes requested):
I see this PR is not marked ready for review yet, but as [commented on the related issue](https://github.com/RexOps/Rex/issues/1439#issuecomment-719354766), I believe the proper fix would be to add the missing feature flag, instead of removing the documentation bits referring to it.

If you would like to work on that too, feel free to update this one by either force-pushing here (and updating title/OP), or close this and open a new one.

PR #1470, opened 2021-02-12
Add ability to pass options to openssh via REX_OPENSSH_OPTS

alip:
As an example, per RFC, IPv6 takes precedent over IPv4, but is sometimes broken
in various exciting ways. In that case it would be useful to pass -4 to ssh.

Since all breakage points are hard to predict, it'd make sense to do it like
git, e.g. set REX_OPENSSH_OPTS and avoid further opt parsing.

## Checklist

- [ ] tests pass on Travis CI 
- [ ] git history is clean 

### DIFF
--- a/lib/Rex/Config.pm
+++ b/lib/Rex/Config.pm
@@ -572,6 +572,14 @@ sub get_openssh_opt {
   return %openssh_opt;
 }
 
+sub get_rex_openssh_opt {
+  if ( exists $ENV{REX_OPENSSH_OPTS} ) {
+    my @opts = split / /, $ENV{REX_OPENSSH_OPTS};
+    return @opts;
+  }
+  return 0;
+}
+
 =head2 set_sudo_without_locales
 
 =head2 get_sudo_without_locales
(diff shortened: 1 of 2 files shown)


ferki (changes requested):
Hmm, the overarching pattern in Rex::Config is to have `{set,get}_$option` pairs, so if we decide to include a first class config option for this, I'd prefer to follow that pattern with it as well.

However, I would note that the primary configuration place for ssh is `~/.ssh/config`, since Net::OpenSSH should pick that up. If that's not the case, arbitrary ssh options (or even Net::OpenSSH constructor options) should still be possible to set with [`Rex::Config->set_openssh_opt`](https://metacpan.org/pod/Rex::Config#set_openssh_opt).

Please test if you can implement your use case by using some of the above, and if none of those work, please open a discussion thread or a new issue.

ferki:
I believe forcing IPv4 in `~/.ssh/config` it should be something like this:

```
Host *.myhosts
    AddressFamily inet
```

And with `Rex::Config->set_openssh_opt`, something like this: `Rex::Config->set_openssh_opt( AddressFamily => 'inet' );`

Please make sure to double-check the `-o` option names and values of ssh in `ssh_config(5)`.

alip:
Setting AddressFamily in `~/.ssh/config` works so this is no longer necessary.
Thank you for the suggestion!

PR #1471, opened 2021-02-12
Allow uploading files even when permissions cannot be set.

alip:
By default, the "file" function uses sftp to upload the target file to
the server. During the upload, the Net::SFTP::Foreign runs chmod
unconditionally, failing if the target file is not owned by the ssh
user, unless that user is a root. This commit changes that behavior by
passing an optional "best_effort" flag that instructs sftp to continue
even if setting permissions has failed.

As a result, an unprivileged user will be able to use Rex scenarios that
call the "file" function without getting permission denied errors.

## Checklist

- [ ] tests pass on Travis CI 
- [ ] git history is clean 

### DIFF
--- a/lib/Rex/Interface/Fs/OpenSSH.pm
+++ b/lib/Rex/Interface/Fs/OpenSSH.pm
@@ -142,7 +142,7 @@ sub upload {
   Rex::Commands::profiler()->start("upload: $source -> $target");
 
   my $sftp = Rex::get_sftp();
-  unless ( $sftp->put( $source, $target ) ) {
+  unless ( $sftp->put( $source, $target, best_effort => 1) ) {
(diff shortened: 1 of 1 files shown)


ferki (changes requested):
I believe this would be best discussed in a feature request issue first.

If we are going to make it a feature, I think the `best_effort` functionality should be controlled via a config option in Rex::Config (whether the user explicitly want it or not), instead of basing it on the fact whether sudo is used or not (which could break for other privilege escalation methods).

And if it's going to be a first class config option, I believe it would be best to make it a generic one to pass arbitrary (Net::SFTP::Foreign?) options similarly to `set_openssh_opt`. That would avoid having to do further pull requests for each use case that might arise.

PR #1492, opened 2021-06-03
Recognize AlmaLinux as RedHat clone

kopytov:


Htbaa:
Would like to see this one merged as well. Currently if you want to use Rex with AlmaLinux you just have to make sure that `redhat-lsb` isn't installed.

ferki (changes requested):
Hey @kopytov, thanks for your contribution!

This PR does not fully follow the project expectations, so I'd like to take this chance to kindly link our [Contributing guide](https://github.com/RexOps/Rex/blob/master/CONTRIBUTING.md), specifically the sections about [GitHub issues and pull requests](https://github.com/RexOps/Rex/blob/master/CONTRIBUTING.md#github-issues-and-pull-requests) and [Tests should pass](https://github.com/RexOps/Rex/blob/master/CONTRIBUTING.md#tests-should-pass).

That being said, I believe the general goal is something we can try to onboard if it's really just about listing another Red Hat clone. However I have some questions at this point, and I'd also like to request some changes on this PR before we can merge:

1. As indicated by the red cross on the commit/PR and the "Some checks were not successful" text, currently the test suite fails. It seems to be caused by [adding a new perlcritic violation](https://github.com/RexOps/Rex/runs/2739104017#step:16:21) about a double quoted string that does not require interpolation.

    I appreciate that the existing style was followed here, but could you fix the quoting, please? As a minimum approach, I think _replacing_ the commit with a single quoted version would do. Alternatively, if you'd like to help out with the code base cleanup effort, a first commit could fix all the existing double quoted strings in this array, and then a second commit could add the actual change itself. I'd be fine with both.

2. The provided and expected PR template seems to be completely deleted, and that makes it harder to review and follow the status. At least we miss a changelog entry here on top of the normal "issue first" approach.

    It feels like the commit message itself could be used as a changelog entry under the enhancement category. Could you add it on the same commit too, please? For bonus points, please also restore the original PR template format.

3. Other than that, I think the change itself is good to be included, but the Wikipedia page of AlmaLinux says that the package manager is DNF, which is not yet supported by Rex. Would simply recognizing AlmaLinux as a Red Hat clone be enough for promising Rex support (which would try to use `yum` for package management)? Wouldn't we need to add DNF support as a new feature first?

    This is exactly a discussion I would rather have on an issue before a PR is being worked on :) Do you think it would make sense to open an issue where we could design and discuss the implementation details first?

Should you need any further assistance, or have questions, come join us on Matrix or IRC (see links on our [Support](https://www.rexify.org/support/index.html) page)!

ehuelsmann:
@ferki or you could defer abstracting out the package management API to a project that already has: PackageKit. It only works on Linux, but it works on most distributions, as far as I know. So it seems like a nice fallback without needing to hard-code support for any of the distributions with their expected package manager.

PR #1497, opened 2021-06-22
Redirect dmidecode STDERR output

nfetisov:
When running dmidecode as non-privileged user, it complains on unsufficient privileges to STDERR.
Redirect all output to STDOUT.

### DIFF
--- a/lib/Rex/Inventory/DMIDecode.pm
+++ b/lib/Rex/Inventory/DMIDecode.pm
@@ -133,7 +133,7 @@ sub _read_dmidecode {
       return;
     }
 
-    eval { @lines = i_run "dmidecode"; };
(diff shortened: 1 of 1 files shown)


ferki:
Thanks again for your contribution, @nfetisov!

I still can't reproduce the issue on my own, and there's no bug report containing the necessary information about the circumstances of the failing use case. Without understanding the root cause, it's also impossible for me to judge what would be the best course of action to fix it.

There were also possibly related changes since this PR was opened. It might worth checking the behavior again, perhaps this is not a problem anymore. If it's still causing a trouble for you, please open a full bug report, or reach out on one of the [support channels](https://www.rexify.org/support/index.html). Either way, I hope to learn more about it through your valuable feedback.

Based on all the above, I decided to close this PR without merging this time.

PR #1499, opened 2021-06-29
Fix file hooks when source option is used

### LINKED ISSUE #1498: Inconsistent file hook handling between "content" and "source" options
## Describe the bug

When using `file $target, content => $content;`, Rex creates a temporary file on the managed host first, then renames that into the actual target file. It also runs `before_change` and `after_change` file hooks around the rename operation.

But when using `file $target, source => $source;`, Rex reuses `Rex::Commands::Upload::upload();` to simply do a file upload without the need for the temporary file, and relies on all the _upload_ hooks, so no `before_change` and `after_change` _file_ hooks are being run.

This leads to inconsistent behavior from the perspective of the `file` resource, and makes it harder than necessary to develop various hooks.

## How to reproduce it

1. Specify a `before_change` file hook.
2. Manage a file with the `content` option, and the above hook will be triggered.
3. Manage a file with the `source` option, and the above hook will not be triggered.

```perl
use Rex -feature => [qw(1.4 exec_autodie)];

use Rex::Hook;
use DDP;

register_function_hooks {
    before        => { file => \&my_before_hook },
    before_change => { file => \&my_before_change_hook },
    after_change  => { file => \&my_after_change_hook },
    after         => { file => \&my_after_hook },
};

sub my_before_hook {
    say '-' x 20;
    say 'before_hook';
    p @_;
}

sub my_before_change_hook {
    say 'before_change hook';
}

sub my_after_change_hook {
    say 'after_change hook';
}

sub my_after_hook {
    say 'after_hook';
    say '-' x 20;
}

task 'inconsistent_hook', sub {
    file '/tmp/1', content => '1';
    file '/tmp/2', source  => 'Rexfile';
    unlink '/tmp/1', '/tmp/2';
};
```

## Expected behavior

I would expect `before_change` and `after_change` file hooks to be executed even for `file $target, source => $source;` cases. For even greater consistency, the same "upload temporary file first, then rename it into target" approach could be used too.

## Circumstances

 - Rex version: 1.13.3
 - Perl version: 5.34.0
 - OS running rex: Gentoo
 - OS managed by rex: Gentoo
 - How rex was installed: `dzil install`

ferki:
This PR is an attempt to fix #1498 by making sure that the file command's `source` option takes the same code path as `content`. This ensures having a temporary file created first, which is then moved to the final location. It also enables `before_change` and `after_change` hooks for the `source` option.

Both of these changes increase the consistency of the file commands behavior across the variety of available options.

## Checklist

- [x] tests pass in CI 
- [x] git history is clean 

### DIFF
--- a/lib/Rex/Commands/File.pm
+++ b/lib/Rex/Commands/File.pm
@@ -387,13 +387,9 @@ The return value of this hook overwrites the original parameters of the function
 
 This gets executed right before the new file is written. All original parameters are passed to it, including the applied defaults (C<ensure =E<gt> 'present'>, resolved path for C<source>).
 
-Only called when the C<content> parameter is used. For the C<source> parameter, the L<upload|Rex::Commands::Upload#upload> hooks are used.
-
 =item after_change
 
 This gets executed right after the file is written. All original parameters, including the applied defaults (C<ensure =E<gt> 'present'>, resolved path for C<source>), and any returned results are passed to it.
-
-Only called when the C<content> parameter is used. For the C<source> parameter, the L<upload|Rex::Commands::Upload#upload> hooks are used.
 
 =item after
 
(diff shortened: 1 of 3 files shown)


lib/Rex/Commands/File.pm
```diff
@@ -552,15 +558,6 @@ sub file {
...
 
   }
-  elsif ( exists $option->{"source"} && !$is_directory ) {
-    $option->{source} =
-      Rex::Helper::Path::get_file_path( $option->{source}, caller() );
-
-    # HOOKS: for this case you have to use the upload hooks!
-    $__ret = upload $option->{"source"}, "$file", force => 1;
-
-    $need_md5 = 0 if $__ret->{changed} == 0;
```
ehuelsmann, line 562: At first I wondered why `$need_md5` doesn't need to be set in the merged codepath. However, the merged code path turns out to have its own changed-ness detection (through md5s, but not through "need_md5". (thanks for answering my questions on IRC).

ehuelsmann (approved):
All in all a very nice cleanup both of the test suite as well as for the `file` command.

PR #1511, opened 2021-10-13
lib/Rex/Config.pm: slurp file content using IO::File.

### LINKED ISSUE #1466: Don't use local @ARGV=...; <> for slurping file contents
Migrated from [rt.cpan.org #98871](https://rt.cpan.org/Ticket/Display.html?id=98871) (status was 'new')

Requestors:
* SREZIC@cpan.org (@eserte)

From srezic@cpan.org (@eserte) on 2014-09-14 13:47:34
:

Rex::Config is using something like

    @ARGV = ...; <>

to slurp file contents, and this may happen during compile time (e.g. read_ssh_config is called in import(), so a "use Rex::Config" would trigger this code). Unfortunately this means that a currently active <> loop would break. See https://github.com/cpan-testers/CPAN-Reporter/issues/20 for an issue where this actually happens.

Probably it's safer to use instead

    do { open my $fh, $file or die $!; undef $/; <$fh> }

even if it's slightly longer. Or alternatively

    use IO::File ();
    join '', IO::File->new($file)->getlines;

monsieurp:
Fixes #1466.

This PR is an attempt to fix #1466. The ticket is self-explanatory as well as those changes.

## Checklist

- [ ] changelog entry included: not needed. 
- [ ] tests pass in CI: not sure there are tests for such a change. 
- [x] git history is clean 

### DIFF
--- a/lib/Rex/Config.pm
+++ b/lib/Rex/Config.pm
@@ -1739,7 +1739,11 @@ sub read_config_file {
   $config_file ||= _home_dir() . "/.rex/config.yml";
 
   if ( -f $config_file ) {
-    my $yaml = eval { local ( @ARGV, $/ ) = ($config_file); <>; };
+    my $yaml = do {
+      use IO::File;
+      my $fh = IO::File->new( $config_file, 'r' );
+      join '', $fh->getlines();
+    };
     eval { $HOME_CONFIG_YAML = Load($yaml); };
 
     if ($@) {
@@ -1762,7 +1766,11 @@ sub read_ssh_config_file {
   $config_file ||= _home_dir() . '/.ssh/config';
 
   if ( -f $config_file ) {
-    my @lines = eval { local (@ARGV) = ($config_file); <>; };
+    my @lines = do {
+      use IO::File;
+      my $fh = IO::File->new( $config_file, 'r' );
+      join '', $fh->getlines();
+    };
     %SSH_CONFIG_FOR = _parse_ssh_config(@lines);
   }
 }


ferki:
Thanks @monsieurp for your contribution!

Tis PR is still marked as draft only, but reading through the linked issue and the proposed changes, I can provide some early feedback below and/or as code comments.

---

> * [ ]  changelog entry included: not needed.

Hmm, yeah, I agree it's probably fine to skip changelog entries for changes such as automated code formatting, refactoring, or CI workflow changes.

---

> * [ ]  tests pass in CI: not sure there are tests for such a change.

As a minimum quality requirement the current test suite should still pass. Ideally, if a changed code path is not covered yet, tests could be added first (to demonstrate a code is solid either by fixing a failing test to proving that it still passes the tests).

Currently `t/config-ssh.t` fails in CI and on my system too.

---

There are currently 35 instances of `local @ARGV ... <>`-style slurps. What do you think about fixing them all? Would we need a good module for that, or perhaps some helper function in Rex core?

---

I've fixed some CI-related issues recently, so you might need to rebase this PR on top if current default branch, and (force) push your branch again.

lib/Rex/Config.pm
```diff
@@ -1739,7 +1739,11 @@ sub read_config_file {
   $config_file ||= _home_dir() . "/.rex/config.yml";
 
   if ( -f $config_file ) {
-    my $yaml = eval { local ( @ARGV, $/ ) = ($config_file); <>; };
+    my $yaml = do {
+      use IO::File;
+      my $fh = IO::File->new( $config_file, 'r' );
+      join '', $fh->getlines();
+    };
```
ferki, line 1746: What do you think about using `LoadFile` from the YAML module to read this YAML file below instead of reading the whole file and `eval`ing it?

lib/Rex/Config.pm
```diff
@@ -1762,7 +1766,11 @@ sub read_ssh_config_file {
   $config_file ||= _home_dir() . '/.ssh/config';
 
   if ( -f $config_file ) {
-    my @lines = eval { local (@ARGV) = ($config_file); <>; };
+    my @lines = do {
+      use IO::File;
+      my $fh = IO::File->new( $config_file, 'r' );
+      join '', $fh->getlines();
```
ferki, line 1772: Can it be that `t/config-ssh.t` fails because now all lines of the file is concatenated with `join` while it was treated as an array of lines before?

ferki (changes requested):
Let's make sure all tests pass. Please [rebase](https://www.w3docs.com/snippets/git/how-to-rebase-git-branch.html) these changes on top of the default branch in order to test them against the current state.

ferki:
ps.: a git commit is explicit about which files are being changed, so there's no need to duplicate that info in the commit message. and to stop it with the period. Also [do not end the subject line with a period](https://chris.beams.io/posts/git-commit/#end).

E.g. instead of:

`lib/Rex/Config.pm: slurp file content using IO::File.`

it's fine to simply write:

`Slurp file content using IO::File`

It might also work better to [explain what and why vs. how](https://chris.beams.io/posts/git-commit/#why-not-how). E.g.

`Avoid slurping with <> during compile time`

lib/Rex/Config.pm
monsieurp: done

PR #1512, opened 2021-10-14
lib/Rex/CLI.pm: display a warning when/if Rex::Template is loaded.

### LINKED ISSUE #1475: Deprecate built-in template engines
## User story

**As** maintainer,
**I would like to** deprecate the built-in template engines,
**so I can** spend more of my efforts on the core capabilities of Rex.

## Describe the solution you'd like

Rex::Template was the original templating engine, but its approach had problems, like almost useless diagnostics. So Rex::Template::NG was born to address some of the shortcomings, but its results are still not fully compatible yet, so we never truly made the switch (default engine has changed, but the old one is still around "just in case").

This situation provides some false promises to users, and adds considerable overhead for both users and maintainers. Therefore I'd like to help avoiding the whole rabbit hole in the first place.

I believe templating is complex topic on its own, and it's best to offload the responsibility to external modules, of which there are dozens on CPAN already. That way we could focus more on the core capabilities of Rex, while everyone may pick their preferred templating engine for their needs.

I think we could put up a one-time warning on the first use of any of the built-in engines in a given run, so later when the time comes, get rid of the modules themselves. It would also help raising awareness about the potential problems when using the built-in ones.

## Describe alternatives you've considered

Creating our own "best" templating engine, at least in the limited scope of configuration management. It doesn't sound realistic to me.

## ~~Additional context~~

monsieurp:
Fixes #1475.

This PR is an attempt to fix #1475.

## Checklist

- [ ] changelog entry included: we might want to add one.
- [ ] tests pass in CI: let's see.
- [x] git history is clean 

### DIFF
--- a/lib/Rex/CLI.pm
+++ b/lib/Rex/CLI.pm
@@ -30,8 +30,6 @@ use Rex::TaskList;
 use Rex::Logger;
 use YAML;
 
-use Data::Dumper;
-
 my $no_color = 0;
 eval "use Term::ANSIColor";
 if ($@) { $no_color = 1; }
@@ -197,6 +195,7 @@ FORCE_SERVER: {
 
   load_server_ini_file($::rexfile);
   load_rexfile($::rexfile);
+  check_template_modules();
 
   #### check if some parameters should be overwritten from the command line
 CHECK_OVERWRITE: {
@@ -453,6 +452,30 @@ sub __help__ {
 
   CORE::exit 0;
 
+}
+
+# Check whether Rex/Template.pm
+# or Rex/Template/NG.pm modules are loaded. If there are,
+# they should be present in the %INC hash.
+sub check_template_modules {
+  my @tpl_modules_loaded =
+    grep { /(Rex\/Template|Rex\/Template\/NG)/ } keys %INC;
+  if ( scalar @tpl_modules_loaded > 0 ) {
+    Rex::Logger::info(
+      "WARNING! You are using Rex built-in template engine aka Rex::Template.",
+      "warn"
+    );
+    Rex::Logger::info(
+      "This engine is obsolete, unmaintained and will be retired in future releases.",
+      "warn"
+    );
+    Rex::Logger::info(
+      "Please consider migrating to a more modern template engine.", "warn" );
+    Rex::Logger::info(
+      "Text::Template::Simple, Template::Toolkit, etc. As usual, CPAN is you friend.",
+      "warn"
+    );
+  }
 }
 
 sub add_help {


lib/Rex/CLI.pm
```diff
@@ -30,8 +30,6 @@ use Rex::TaskList;
 use Rex::Logger;
 use YAML;
 
-use Data::Dumper;
-
```
ferki, line 34: Is there any reason why Data::Dumper is also removed in this context about Rex::Template::* deprecation?

lib/Rex/CLI.pm
```diff
@@ -455,6 +454,30 @@ sub __help__ {
 
 }
 
+# Check whether Rex/Template.pm
+# or Rex/Template/NG.pm modules are loaded. If there are,
+# they should be present in the %INC hash.
+sub check_template_modules {
```
ferki, line 460: We have `Rex::deprecated` to mark things deprecated in Rex core in a uniform way. That might come in handy here too. Or perhaps during `import` time of Rex::Template and Rex::Template::NG.

lib/Rex/CLI.pm
```diff
@@ -455,6 +454,30 @@ sub __help__ {
...
+    Rex::Logger::info(
+      "This engine is obsolete, unmaintained and will be retired in future releases.",
+      "warn"
+    );
+    Rex::Logger::info(
+      "Please consider migrating to a more modern template engine.", "warn" );
+    Rex::Logger::info(
+      "Text::Template::Simple, Template::Toolkit, etc. As usual, CPAN is you friend.",
+      "warn"
+    );
```
ferki, line 477: What do you think about pointing to [Rex::Template::TT](https://metacpan.org/pod/Rex::Template::TT) as a readily available and recommended external module, _and_ point to the docs about how to set one's own custom templating functions:

- [set_template_function](https://metacpan.org/pod/Rex::Config#set_template_function) config option
- [template_function](https://metacpan.org/pod/Rex::Commands#template_function) command

lib/Rex/CLI.pm
```diff
@@ -455,6 +454,30 @@ sub __help__ {
...
+    grep { /(Rex\/Template|Rex\/Template\/NG)/ } keys %INC;
+  if ( scalar @tpl_modules_loaded > 0 ) {
+    Rex::Logger::info(
+      "WARNING! You are using Rex built-in template engine aka Rex::Template.",
+      "warn"
+    );
+    Rex::Logger::info(
+      "This engine is obsolete, unmaintained and will be retired in future releases.",
+      "warn"
+    );
```
ferki, line 471: That would also show up when internal core modules import Rex::Template and the user doesn't use it directly. Is there a good way to make an exception based on who imports it?

ferki (changes requested):
Thanks for this contribution as well, @monsieurp!

Hmm, quite a few core modules depend on Rex::Template so I'm afraid we can't simply deprecate it like that. I'd say we have to make sure the internal template-processing keeps working as before (and without deprecation warnings if not used directly).

We might need to move the discussion and design part back to the original #1475 issue. As briefly mentioned there, my preferred way would be to see if we could migrate the internal custom logic of Rex::Template to something like [Text::Template::Simple](https://metacpan.org/pod/Text::Template::Simple) (which seems close enough syntax-wise), and then perhaps making Rex::Template::NG to be an alias of Rex::Template (with a deprecation notice upon import).

What do you think about that?

lib/Rex/CLI.pm
monsieurp: This module is unused in `CLI.pm`.

lib/Rex/CLI.pm
monsieurp: Good point but I don't really know how to solve this problem though.

lib/Rex/CLI.pm
monsieurp: I can do that indeed.

PR #1526, opened 2022-06-03
POD: Make @regexp optional in syntax

rwp0:
Docs: made `@regexp` argument optional in the syntax since:

> If no regexp is supplied, the line is appended unless there is already an identical line in $file.

It confused me to pass a third argument reading the function POD in MetaCPAN.

cf. `extract($file [, %options])` ibid.

https://metacpan.org/pod/Rex::Commands::File#append_if_no_such_line($file,-$new_line,-@regexp)

This PR is an attempt to fix `append_if_no_such_line` function documentation.

## Checklist

- [ ] changelog entry included (minor change)
- [ ] tests pass in CI (documentation change)
- [ ] git history is clean

### DIFF
--- a/lib/Rex/Commands/File.pm
+++ b/lib/Rex/Commands/File.pm
@@ -1016,7 +1016,7 @@ sub delete_lines_according_to {
 
 }
 
-=head2 append_if_no_such_line($file, $new_line, @regexp)
+=head2 append_if_no_such_line($file, $new_line[, @regexp])
 
 Append $new_line to $file if none in @regexp is found. If no regexp is
 supplied, the line is appended unless there is already an identical line


ferki:
Thanks for your contribution, @rwp0!

This PR is still marked as draft, so I can't tell if there will be more commits coming here or not :) Documentation tweaks are highly appreciated, the pending change itself is well-spotted and looks good :+1: 

When you are ready with drafting this pull request, please mark it as ready for review. In case you have questions before that, please ask.

----

## General feedback at first glance

A [GitHub issue is expected before a pull request](https://github.com/RexOps/Rex/blob/master/CONTRIBUTING.md#github-issues-and-pull-requests), and the [Git workflow](https://github.com/RexOps/Rex/blob/master/CONTRIBUTING.md#git-workflow) section of the Contributing guide lists specific ideas about how it's best to approach git details. I'll list a few improvement opportunities for now or next time to streamline the contributing process:

- branching: it is generally recommended to use a different branch for contributions in your fork than `master`. It's not a problem in itself, but it's often more convenient to manage changesets separately.
- commit message formatting: it's currently a very long single line, which makes git history harder to read. Please look at the git history for examples of well-written commit messages, and see [The seven rules of a great Git commit message](https://cbea.ms/git-commit/#seven-rules) for general ideas to aim for. There are now checks for most of those rules to give fast automated feedback, so if you [rebase](https://timonweb.com/misc/how-to-update-a-forked-repo-from-an-upstream-with-git-rebase-or-merge/) your changes on top of the current default branch, the failing GitHub Actions workflow would explain why.
- changelog: Rex follows (dotted) semantic versioning for releases, and the [ChangeLog](https://github.com/RexOps/Rex/blob/8a0f1288a34ab686622423a2ac3167e83ef49bed/ChangeLog) entries are used to automatically determine the next version number upon release. Therefore it's necessary to include proper changelog entry for each logical change to the user-facing codebase. It's probably best to include it in the same commit that introduces the changelog entry, otherwise the maintainer has to add one manually before merging.
- passing tests for doc changes: tests must pass before merging a pull request, even if the change is documentation-only. It must be still demonstrated that the given change doesn't break any tested behavior, and docs can be tested as well.

rwp0:
Done

lib/Rex/Commands/File.pm
```diff
@@ -1016,7 +1016,7 @@ sub delete_lines_according_to {
 
 }
 
-=head2 append_if_no_such_line($file, $new_line, @regexp)
+=head2 append_if_no_such_line($file, $new_line[, @regexp])
```
ferki, line 1019: As a general rule, I aim for going towards unified formatting style (including docs) in the largest scope possible (codebase, or at least within the files, or functions).

The formatting of optional and/or trailing parameters seems to be inconsistent throughout the codebase, but there's an example for a similar an optional trailing parameter in Rex::Commands::File: [`extract($file [, %options])`](https://metacpan.org/pod/Rex::Commands::File#extract($file-%5B,-%25options%5D)).

To have consistent formatting for the same concepts within the same file, what do you think about adding one more space and write it as `($file, $new_line [, @regexp])`?

ferki: Apparently, for `append_if_no_such_line` there is an old format with positional parameters (like here), and a new (preferred) format with named parameters (like `regexp => ...`).

The old format supports optionally passing a single regular expression or multiple ones as an arrayref. It also supports optional `on_change` and `on_no_change` hooks. Some of that is also apparently missing from the detailed docs.

It is not clear how it would be best to address all those variations in a unified format, or whether thinking about these would be in the scope of this story or not.

Removing one point of confusion is certainly a step forward, so let's push this towards merging.

I just wanted to mention the above "accidental complexity" because I think it's a good example of why discussing a planned change first is almost always a good idea in this repo's context. "Open an issue before investing work in a PR" is the translation of that idea to GitHub workflow ;)

ferki (changes requested):
As mentioned in the early feedback, please:

- [rebase](https://timonweb.com/misc/how-to-update-a-forked-repo-from-an-upstream-with-git-rebase-or-merge/) this change on top of the default branch in order to test it against the current state (should test the git commit message formatting now)
- fix the git commit message
- include a changelog entry

----

I would probably write a `Clarify append_if_no_such_file docs` entry under the `[DOCUMENTATION]` category in the ChangeLog file.

The same could also be used as the subject of the commit message, and the commit message body might not required at all as it is mostly a copy-paste of the docs anyway.

I don't think a separate `Update changelog` commit would add any value in this case, so the changelog entry and the change could both go into the same single commit.

----

I left some notes directly on the changeset about consistent formatting aspects as well. Since there's no single consistent style today about optional trailing parameters, it's not necessarily required to fix that now, but since the commit has to be replaced anyway, it might be a good opportunity to address that too.

ferki:
Alright, let's take a look at the wider scope on the other PR :)

rwp0:
> Alright, let's take a look at the wider scope on the other PR :)

it's here https://github.com/RexOps/Rex/pull/1530

PR #1569, opened 2023-02-09
Avoid nested subtests while forking

### LINKED ISSUE #1565: Attempt to free unreferenced scalar error on Windows before perl-5.20.x
### Describe the bug

Forking from either Rex::Fork::Task or via Parallel::ForkManager results in "Attempt to free unreferenced scalar" errors upon `exit()`-ing from the child process on Windows with Perl versions before 5.20.2 or 5.20.3.

Since the failure is in the last step of the execution, it only affects the termination of the child process. The task running in the child process can still properly do most of its work, including reporting back its own results via internal data structures (exit code, error messages, etc.), which happens before.

The return value of the task may be lost, though. This might need further tests on top of what is in `t/summary.t`.

Since this seems to be related to a reference counting bug in the affected Perl versions, there might be other side-effects too (memory leaks, crashing, etc.).

See details in automated test results on GitHub Actions and/or CPAN Testers.

### Expected behavior

Ideally, there should be no errors reported from forks on Windows.

At least it should be made sure it doesn't cause problems with Rex functionality, for example by adding more tests (especially for the task's return values).

Speculatively, it may be possible to add a non-forking execution model for Windows. The amount of work required for this is unclear, though.

If all else fails, consider dropping Windows support for Perl versions before 5.20.3.

### How to reproduce it

Run Rex tasks, and/or `t/summary.t` on Windows with a Perl version before 5.20.2 or 5.20.3.

### Code example

No specific code example besides `t/summary.t` at this point.

### Additional context

5.20.1 is affected, and it seems to be first fixed on 5.20.3, but I couldn't test 5.20.2 yet.

[Selected Bug Fixes](https://perldoc.perl.org/perl5202delta#Selected-Bug-Fixes) from perl5202delta mentions:

> On Win32, restoring in a child pseudo-process a variable that was `local()ed` in a parent pseudo-process before the `fork` happened caused memory corruption and a crash in the child pseudo-process (and therefore OS process). [GH #8641]

which might be related. It was fixed in 5.22.0 and backported into 5.20.3.

### Rex version

1.13.4

### Perl version

5.20.1

### Operating system running rex

Windows

### Operating system managed by rex

local

### How rex was installed?

dzil install

ferki:
This PR is a proposal to fix #1565 by avoiding nested subtests while also forking.

I open it as a draft since there are other improvement opportunities in the same file, but the exact scope of them is still to be decided. This is probably a minimal set of changes to fix the immediate issue, though.

## Checklist

- [x] automated tests pass 
- [x] git history is clean 

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -61,7 +61,7 @@ As a general rule, managing an endpoint with Rex is only supported for platforms
 
 ### Supported Perl versions
 
-Rex aims to run even on older Perl versions up to 10 years old. Currently this means 5.12.5. The recommended minimum version of Perl on Windows is 5.20.3.
+Rex aims to run even on older Perl versions up to 10 years old. Currently this means 5.12.5.
 
 On top of the supported minimum version of Perl, the goal is to support the latest versions of all minor Perl 5 releases. That makes the full list the following:
 
--- a/ChangeLog
+++ b/ChangeLog
@@ -4,6 +4,7 @@ Revision history for Rex
  [API CHANGES]
 
  [BUG FIXES]
+ - Fix attempt to free unreferenced scalar on Windows
 
  [DOCUMENTATION]
 
--- a/t/summary.t
+++ b/t/summary.t
@@ -1,7 +1,7 @@
 use strict;
 use warnings;
 
-use Test::More tests => 2;
+use Test::More;
 use Test::Deep;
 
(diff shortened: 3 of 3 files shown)


ferki:
@uralm1: your suggested approach to remove a layer of subtests indeed seems to help with the immediate problem :+1:

In case you could test it on your machines as well, please leave a note about the results.

PR #1593, opened 2023-06-02
Sync command improvements

bbrtj:
This PR introduces a couple of changes to `sync_up` and `sync_down` commands:
- speeds them up significantly when there are a lot of excluded files/directories by skipping them before calculating their md5
- improves handling of excluded files - previously only file basename was taken into account, which made it very confusing. Directory handling was also suboptimal. It should now work more like things like `.gitignore` work, however **it breaks backcompat**.

Because of this I'm unsure whether this PR will be considered. If it's not, I'll probably upload a standalone CPAN module instead, as current implementation of sync is not very usable in my case (more excluded than included files, takes ages to sync anything and can't exclude properly).

Not very tested yet, but `sync_up` works in my case.

## Checklist

- [ ] automated tests pass 
- [ ] git history is clean 

### DIFF
--- a/lib/Rex/Commands/Sync.pm
+++ b/lib/Rex/Commands/Sync.pm
@@ -94,30 +94,7 @@ sub sync_up {
   $source = get_file_path( $source, caller );
 
   #
-  # first, get all files on source side
-  #
-  my @local_files = _get_local_files($source);
-
-  #print Dumper(\@local_files);
-
-  #
-  # second, get all files from destination side
-  #
-
-  my @remote_files = _get_remote_files($dest);
-
-  #print Dumper(\@remote_files);
-
-  #
-  # third, get the difference
-  #
-
-  my @diff = _diff_files( \@local_files, \@remote_files );
-
-  #print Dumper(\@diff);
-
-  #
-  # fourth, build excludes list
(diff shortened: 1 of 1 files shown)


bbrtj:
Thanks @ferki, as you can imagine it would be wasteful to follow contribution guide to its fullest if the change would get rejected in the end due to backcompat breakage. Now that you confirmed it's not as big of an issue I can close this one and open a new one the proper way.

ferki:
Yep, we collected the bits in the contribution guide over the last decade or so, and that's exactly the idea behind most things in it: foster communication and discovery before investing too much into some changeset.

I'd probably go with something like the following flow:

- treat the two ideas here separately so they don't hold each other up accidentally
- look into how we can add some initial tests for Rex::Commands::Sync before changing things around (apparently there aren't any yet :disappointed:)
- this might prove to be a good opportunity to clean up or refactor things in order to simplify our work on the actual changes
- the "filter before diff" idea sounds less complex to me, so that might be good to tackle first
- I'm curious to see how we can best accommodate for the "advanced filtering" bits

I'm looking forward to learn more about your ideas, and to hack together on bringing them to more Rex users!

PR #1598, opened 2023-06-12
Improve excluded files handling in Rex::Commands::Sync

### LINKED ISSUE #1597: Rex::Commands::Sync excludes are poorly handled
### Describe the bug

There are two issues with sync_up and sync_down commands regarding excluded files and directories:
1. Excludes are looked at too late, after calculating md5 of all files in the directory (recursively). This means a lot of time is spent doing meaningless work, for example like calculating md5 of all `node_modules` which are often excluded anyway.
2. Excludes work well only when a flat file is excluded. Excluding nested files or whole directories does not work as one would expect from using tools like gitignore.

### Expected behavior

1. There's absolutely no need to calculate md5 sums of excluded files, as those will be thrown away unused.
2. I can't grasp how the excluded directories currently work, but I would expect I have to pass a full path to a file which I want excluded, minus any wildcards. 

### How to reproduce it

1. Have a large directory alongside your files
2. Try syncing up with the directory excluded

### Code example

```perl
sync_up 'my_dir', 'my_remote_dir', {
  exclude => ['directory_name'],
};
```

### Additional context

_No response_

### Rex version

(R)?ex 1.14.2

### Perl version

This is perl 5, version 36, subversion 0 (v5.36.0) built for amd64-freebsd

### Operating system running rex

FreeBSD 13.2 (amd64)

### Operating system managed by rex

FreeBSD 13.1

### How rex was installed?

cpan client

bbrtj:
This PR is a proposal to fix #1597  by refactoring Rex::Commands::Sync .

## Checklist

- [ ] automated tests pass 
- [ ] git history is clean 


bbrtj:
Windows tests are failing (chmod error) - maybe that's the reason sync was not in automated tests up until now?

ferki:
> Windows tests are failing (chmod error)

My first thought was that Windows doesn't have `chmod`, but apparently Rex:Interface::Base::chmod() [should fake success]((https://github.com/RexOps/Rex/blob/9899caafcf8fb847f2cefe586ea7b80d9ff70cb0/lib/Rex/Interface/Fs/Base.pm#L154)) in those cases (and `t/file.t` should also [cover that situation](https://github.com/RexOps/Rex/blob/9899caafcf8fb847f2cefe586ea7b80d9ff70cb0/t/file.t#L54)).

> maybe that's the reason sync was not in automated tests up until now?

Hmm, the reason is more like Rex::Commands::Sync was a quite early feature, and at the time tests were (sadly) often neglected. It's certainly time to add proper tests now before we are going to change related code

In similar situations I tend to first push only the tests as a first commit to a draft PR (or even only to my own fork) and make sure my tests are correct. Then start on the actual change(s) in separate commit(s). Perhaps it's best to follow something like that here too (or even just focus on an "Add initial sync tests" PR first).

It might also be relevant to keep "[When do I use SKIP vs. TODO?](https://metacpan.org/pod/Test::More#When-do-I-use-SKIP-vs.-TODO?)" in mind. If something is unsupported or impossible to test on a platform that may be OK to SKIP in the test suite. If something is known to fail, but possible/have to be fixed later, that may be OK to mark as TODO.

> Edit: seems to only apply to dir with spaces, so maybe easily fixable nonetheless.

Good catch :+1: Paths with special characters, like spaces, could be requiring special quoting and/or escaping on Windows (and/or other platforms).

ps.: given the test failures, I haven't done a full code review yet, but please consider using [explicit test plans](https://metacpan.org/pod/Test::More#I-love-it-when-a-plan-comes-together) overall and in subtests too.

PR #1613, opened 2024-02-02
Remove unused prerequisite for IO::String

robrwo:
Also note that IO::String is only necessaryfor Perls before 5.8. Since Rex requires 5.12, this is unnecessary.

### DIFF
--- a/lib/Rex/Commands/Augeas.pm
+++ b/lib/Rex/Commands/Augeas.pm
@@ -51,7 +51,6 @@ use Rex::Commands::Fs;
 use Rex::Commands::File;
 use Rex::Helper::Path;
 use Rex::Helper::Run;
-use IO::String;
 use Module::Load::Conditional qw(can_load);
 
 my $has_config_augeas =


ferki:
Thanks for contributing, and for your patience, @robrwo, it looks like a good find!

It apparently dates back to the [initial commit](https://github.com/RexOps/rex-augeas/commit/c08df4c7290659fba63d412e2c9f901306f26d7e) of the originally out-of-core [Rex::Augeas](https://github.com/RexOps/rex-augeas] module – and seems already unused even there.

At first glance, I agree we should remove it if it’s unused.

## About our contributing guide

This PR does not follow our [Contributing guide](https://github.com/RexOps/rex/blob/master/CONTRIBUTING.md) without explaining why, forcing me to fall back to guessing. Some things I kept wondering about:

- work status: it’s still marked as draft suggesting “still work in progress”
- intended goal/scope: I haven’t found a related open issue this would resolve
- progress: the PR template seems ignored/deleted, giving some extra work to track the checklist
- safety: no tests covering the related functionality (though I have recently covered most of this module with tests separately, see #1623)
- nitpick: the commit message has a typo (missing space)

## Towards merging

Nevertheless, let’s see how to move it forward. My ideas:

- Now that I merged initial Augeas tests, I suggest rebasing this draft PR on top of the current default branch, which also gives an opportunity to fix the typo.
- Updating the PR should trigger running the test suite again – when tests pass, please remove its draft status to signal “ready to review”.

## Do we know a generic solution for unused dependencies?

As a related topic, I wonder how to test for unused modules as a generic approach :thinking:

Is there a good tool, or perhaps a Perl::Critic policy which could have caught this unused dependency earlier?

If we can identify a good approach we can turn it into a test, I’d love to discover and discuss that separately.

